PR #43, opened 2022-05-19
Add `.gitignore`, `.editorconfig`, `CONTRIBUTING.md`

AlexWaygood:


srittau (approved):
Contents look good, but I think we should just squash the commits and reference python/typing for the full credits.

AlexWaygood:
> Contents look good, but I think we should just squash the commits and reference python/typing for the full credits.

Duly squashed!

Issue #6, opened 2022-05-17
Adding protocols and aliases from _typeshed to typing-extensions


srittau:
This was already partly discussed in python/typing_extensions#9. I think we should start adding some commonly used protocols, type aliases, and other features from `_typeshed` to typing\_extensions so they are available at runtime. This would prevent awkward guarded imports at runtime.

I would suggest to put this into a submodule to separate those types from the "backports and experimental features" that are added to the main namespace. Suggestion: `typing_extensions.ext`. For a start I would suggest to add the following:

* Some simple dunder protocols: `SupportsNext`, `SupportsAnext`, `SupportsTrunc`, `SupportsDivMod`, `SupportsRDivMod`. (I would hold off on protocols like `SupportsAdd` for now as they are using `Any`.)
* `SupportsGetItem`, `SupportsItems` and `SupportsKeysAndGetItem`
* The path aliases `StrPath`, `BytesPath`. `StrOrBytesPath` probably doesn't have much value over `StrPath | BytesPath`.
* Some simple I/O protocols: `SupportsRead`, `SupportsReadline`, `SupportsWrite`

hauntsaninja:
Thanks for pushing this forward! It's great to have a place where we can put generally useful definitions.

That said, there are a couple things it might be nice to have clarity on before we hit merge on #45

First, I'm curious what the relationship we expect between utils and typing.py is. Do we expect to ever merge definitions into typing.py? Note that if we do, the currently clear line between backport and utils gets murky. (If the answer is no, then I'm definitely in favour of subpackage. If the answer is yes, then I have some more questions about when we'd do so)

Second, it might be worth discussing and documenting criteria for inclusion and for stability. As it stands, I think we'll need to be conservative here because `typing_extensions` is so totally insanely widely used. For instance, if we were to make a breaking change that causes many libraries to pin `typing_extensions`, we'd limit the whole ecosystem's ability to access new type system features (related, https://github.com/python/typing_extensions/issues/40 ).

Note that the chances of us making breaking changes are very real. An example of this is the issue linked above,  https://github.com/python/typing_extensions/issues/9, which sparked a lot of interest in doing this in the first place — but `SupportsLessThan` has been removed from typeshed.

[...]

PR #45, opened 2022-05-21
Add typing_extensions.utils

srittau:
Closes: #6


src/typing_extensions/ext.py
```diff
@@ -0,0 +1,99 @@
...
+    def __anext__(self) -> Awaitable[_T_co]:
+        ...
+
+
+class SupportsDivMod(Protocol[_T_contra, _T_co]):
+    def __divmod__(self, __other: _T_contra) -> _T_co:
+        ...
+
+
+class SupportsGetItem(Container[_KT_contra], Protocol[_KT_contra, _VT_co]):
```
JelleZijlstra, line 31: I don't know if inheriting from Container is right. A type that supports `__getitem__` also automatically supports `in`, but type checkers may not know that, so a type providing only `__getitem__` may get rejected from this protocol.

src/typing_extensions/ext.py
```diff
@@ -0,0 +1,99 @@
...
+    def __getitem__(self, __k: _KT_contra) -> _VT_co:
+        ...
+
+
+class SupportsNext(Protocol[_T_co]):
+    def __next__(self) -> _T_co:
+        ...
+
+
+class SupportsRDivMod(Protocol[_T_contra, _T_co]):
```
JelleZijlstra, line 41: This seems very obscure, and Python only uses it in two places. I think we can skip it.

In fact even `SupportsDivMod` (without the R) seems marginal; it's also only used twice in typeshed.

JelleZijlstra (commented):
Thanks! Some general feedback:

- I don't like the `ext` name. It's short for "extension", so it's extensions to the extensions?
- I'm worried making typing-extensions into a package will break some downstream use of the library. I suppose we could just put the new names into typing_extensions.py, and then we don't have to think about the `ext` name either.
- With these additions, we should also provide better documentation. Maybe a page on typing.readthedocs.io.

AlexWaygood (commented):
This looks great! I agree with @JelleZijlstra's points, however. On the naming, maybe `typing_extensions.utils`, on the grounds that these are small utilities that can be reused a bunch of times?

I'm also slightly nervous about including SupportsANext, given that sometimes you want a guarantee that `__anext__` will be a coroutine method, and sometimes you don't (https://github.com/python/typeshed/pull/7491). It might not be as easy to use as users expect.

srittau:
I like `utils` as a sub-package name. I'd still want to use a sub-package, to clearly distinguish between back-/forwards ports and real "extensions". I will test a package version of typing_extensions later with mypy, but apart from that what could realistically break?

src/typing_extensions/ext.py
srittau: But it *is* used. I think it doesn't hurt to have them here instead of `typing`. In fact, for completeness and consistency's sake I think it will make sense in the long run to have protocols for the other arithmetic operators here.

srittau:
> I'm also slightly nervous about including `SupportsANext`, given that sometimes you want a guarantee that `__anext__` will be a coroutine method, and sometimes you don't ([python/typeshed#7491](https://github.com/python/typeshed/pull/7491)). It might not be as easy to use as users expect.

In cases, where `__anext__()` needs to return a coroutine, a type error would be reported, wouldn't it, so it's easy to spot. Maybe this is something we can point out in the documentation?

src/typing_extensions/ext.py
AlexWaygood: I think `SupportsRDivMod` is _okay_, though I agree with @JelleZijlstra that it isn't really used much, so I'd also be fine with omitting it.

A lot of the other arithmetic dunders are quite tricky, though (BinOp dunders in general are quite tricky), due to the presence of the reflected dunders. It can be hard to predict when you _actually_ need an object with an `__add__` method, and when "anything with either an `__add__` or an `__radd__` method" will do. `SupportsRichComparison` in `_typeshed` is a really good example: it has to be a union of protocols to work and, although it works really well for loads of functions in typeshed, we still have to use a _different_ union of protocols for the functions in `_operator.pyi`.

AlexWaygood:
> > I'm also slightly nervous about including `SupportsANext`, given that sometimes you want a guarantee that `__anext__` will be a coroutine method, and sometimes you don't ([python/typeshed#7491](https://github.com/python/typeshed/pull/7491)). It might not be as easy to use as users expect.
> 
> In cases, where `__anext__()` needs to return a coroutine, a type error would be reported, wouldn't it, so it's easy to spot. Maybe this is something we can point out in the documentation?

I still think this would be somewhat surprising behaviour for users, but I'm okay with just adding a note to the documentation :)

AlexWaygood:
> I'd still want to use a sub-package, to clearly distinguish between back-/forwards ports and real "extensions".

I also really like the idea of a subpackage, FWIW!

srittau:
So, any updates? Is the PR fine as it is or should I change things? In case it's considered a blocker, I can merge the subpackage, but I think that will cause more problems down the line, concerning maintainability/extensibility and distinction between typing backports and extensions.

src/typing_extensions/utils.py
```diff
@@ -0,0 +1,99 @@
+"""
+Protocols and type aliases that are not scheduled for inclusion in typing.
+"""
+
+from os import PathLike
+from typing import AbstractSet, Awaitable, Iterable, Tuple, TypeVar, Union
+from typing_extensions import Protocol, TypeAlias
```
JelleZijlstra, line 7: Maybe add `__all__`?

src/typing_extensions/utils.py
```diff
@@ -0,0 +1,99 @@
...
+# Protocols for dunder methods
+#
+
+
+class SupportsAnext(Protocol[_T_co]):
+    def __anext__(self) -> Awaitable[_T_co]:
+        ...
+
+
+class SupportsDivMod(Protocol[_T_contra, _T_co]):
```
JelleZijlstra, line 26: I still think we should drop the divmod protocols:

- They're not widely used: in typeshed we only use them for `divmod` itself.
- They're unintuitive in that divmod() normally returns a pair, but the protocols don't reflect that.

README.md
```diff
@@ -144,4 +144,4 @@ These types are only guaranteed to work for static type checking.
 ## Running tests
 
 To run tests, navigate into the appropriate source directory and run
-`test_typing_extensions.py`.
+`test_typing_extensions/__init__.py`.
```
JelleZijlstra, line 147: I can't comment there, but we should also mention the `utils` module in the README around line 34.

harahu:
See the comments of @gvanrossum in #47 regarding the naming of the subpackage. `typing_utils` was suggested as a better alternative to `utils`.

harahu:
What's the status on this PR? It seems somewhat stale.

hauntsaninja:
At least from my perspective:
There's some stuff we should think through re: upstream typing.py, stability, versioning and type checker special casing.
I think there's a good chance that once we do that we find we may want a separate package, so to that extent ideally we'd resolve some of this before merging.
See my comments on the issue for more details: https://github.com/python/typing_extensions/issues/6#issuecomment-1146498383

harahu:
> `typing_extensions` is a top 10 package on PyPI so it behooves us to be thoughtful, even if that constrains how fast we can move.

Not implying you should rush. Just couldn't find the reasoning for it being left in limbo in the comments here. Hoped for something explicit, like you just gave me. Appreciated.

JelleZijlstra:
Closing this following the policy to only include things that are in CPython or in a PEP. Let's instead create a separate package for reusable types.

Issue #5, opened 2022-05-19
Add a covariant `Mapping`-like type


harahu:
## What

I want a new protocol added to `typing` or `typing_extensions`, expressing the aspects of `Mapping` that are covariant with regards to the key type. That is, the `Mapping` type, but without the `__getitem__` and `get` methods.

Something along the lines of:

```py
class CovariantMappingLike(Protocol[KT_co, VT_co]):
    # A better name for this protocol might be needed
    def __eq__(self, other: object) -> bool: ...
    def __ne__(self, other: object) -> bool: ...
    def keys(self) -> KeysView[KT_co]: ...
    def items(self) -> ItemsView[KT_co, VT_co]: ...
    def values(self) -> ValuesView[VT_co]: ...
```
## Why

`Mapping`s are often, in practice, treated in ways that make them covariant with regards to key. A specific example is laid out in a comment of mine found here: https://github.com/python/typing/issues/445#issuecomment-1131413084

The `Mapping` type, however, cannot be covariant w. regards to key, as it also implements methods that don't allow for it (`__getitem__` and `get`) to be covariant.

This leaves me unable to express, in a simple and idiomatic way, that I intend to use a `Mapping` in a way that respects covariance.

Being able to write:
```py
def foo(mapping: CovariantMappingLike[Hashable, Hashable]) -> int:
    ...
```

_would_ allow me to express what I'm after. It would allow me to express: "Give me a mapping, but I promise to not be looking up keys on it."

This, again, makes it possible to have narrower key types for my `Mapping` objects, as I can pass a narrowly typed `Mapping` to `foo`, without casting:

```py
my_map: Dict[Literal["foo", "bar", "baz"], str] = {}

# I want this to be fine
foo(my_mapping)
```

_I was prompted by @srittau to create this issue in https://github.com/python/typing/issues/445#issuecomment-1131458824_

gvanrossum:
I feel this is too rarely needed to add here. You can just define a protocol in your own code.

harahu:
> I feel this is too rarely needed to add here. You can just define a protocol in your own code.

The need to interact co-variantly with mapping keys seems pretty common to me. See:
- https://github.com/python/typing/issues/445
- https://github.com/python/mypy/issues/1114

This type can be used for every function that interacts with a mapping without looking up values, i.e. covering all "safe" use-cases of what's suggested in the issues mentioned above. That's a pretty good compromise solution IMO.

[...]

PR #47, opened 2022-06-03
Add a partial, covariant Mapping type

harahu:
Addresses #5


harahu:
@srittau Decided to make an attempt at implementing the protocol we discussed a while ago.

Is there a checklist for what needs covering in a PR to typing_extensions?
I'm thinking in terms of docs, testing and the like.

srittau:
Side note: Depending on where #45 goes, this might be better suited for the new submodule.

JelleZijlstra:
The main typing-extensions namespace should not contain names that don't have a prospect of going into `typing`. Potentially this could go into `typing_extensions.util` if and when we add that.

Aside, I don't like the name much.

gvanrossum:
FWIW 'util' or 'utils' is a terrible name for a submodule. (And as I wrote in #5, I am not in favor of this PR in general, but I'll defer to Jelle+Sebastian.)

harahu:
> The main typing-extensions namespace should not contain names that don't have a prospect of going into `typing`. Potentially this could go into `typing_extensions.util` if and when we add that.

See comments above. This is the intention. Was just eager to get started. Hope you don't mind.

> Aside, I don't like the name much.

I agree. The name is the part I struggle with the most. I have several other ideas, all of which I have trouble with:
- PartialMapping
- PartialMappingCo
- MappingLike
- MappingLikeCo

Suggestions welcome, in other words.

harahu:
> FWIW 'util' or 'utils' is a terrible name for a submodule. (And as I wrote in #5, I am not in favor of this PR in general, but I'll defer to Jelle+Sebastian.)

It would be constructive to provide an alternative to the claimed terrible name.

gvanrossum:
> > FWIW 'util' or 'utils' is a terrible name for a submodule. (And as I wrote in #5, I am not in favor of this PR in general, but I'll defer to Jelle+Sebastian.)
> 
> It would be constructive to provide an alternative to the claimed terrible name.

Sorry, you're right. In my experience in any large code base there are dozens of leaf modules named util[s], and the name just doesn't convey any meaning. Worse, if you are tempted to write `from foobar import utils` now you have a `utils` global that could have been imported from many different places.

Even if we called it `typing_utils` that would be better. Depending on the intended use of this package we could name it more specifically but I haven't followed what it's for -- @JelleZijlstra or @srittau could probably comment. (Not clear that it belongs in this PR's discussion, except you alluded that putting it in such a submodule was your intention.)

AlexWaygood:
> Even if we called it `typing_utils` that would be better.

I'm the person who suggested `utils`, and I see your point — `typing_utils` sounds good to me!

But yeah, this is off-topic for this thread.

harahu:
> (Not clear that it belongs in this PR's discussion, except you alluded that putting it in such a submodule was your intention.)

Yeah, sorry for encouraging continuing the conversation here. I made a comment mentioning this conversation in #45, in order to make the relevant conclusions visible.

thomkeh:
Isn't this basically already covered by the `SupportsItems` protocol in #45 ?

harahu:
> Isn't this basically already covered by the `SupportsItems` protocol in #45 ?

This protocol would be a subtype of `SupportsItems`, but requires more from the implementer than just `items`, so no.

JelleZijlstra:
Closing this following the policy to only include things that are in CPython or in a PEP.

PR #49, opened 2022-06-19
Small speed-ups.

denballakh:


src/typing_extensions.py
```diff
@@ -135,7 +135,7 @@ def _collect_type_vars(types, typevar_types=None):
         ):
             tvars.append(t)
         if _should_collect_from_parameters(t):
-            tvars.extend([t for t in t.__parameters__ if t not in tvars])
+            tvars.extend(t for t in t.__parameters__ if t not in tvars)
```
hauntsaninja, line 138: It's actually usually faster to extend from a list comprehension than a generator comprehension. So I'd undo this one.

src/typing_extensions.py
```diff
@@ -664,7 +664,7 @@ def _typeddict_new(*args, total=True, **kwargs):
                             "argument: '_typename'")
         if args:
             try:
-                fields, = args  # allow the "_fields" keyword be passed
+                fields = args[0]  # allow the "_fields" keyword be passed
```
hauntsaninja, line 667: This change does not preserve semantics in exception cases. Please undo.

src/typing_extensions.py
```diff
@@ -937,7 +937,7 @@ def __repr__(self):
 
         def __reduce__(self):
             return operator.getitem, (
-                Annotated, (self.__origin__,) + self.__metadata__
+                Annotated, (self.__origin__, *self.__metadata__)
```
hauntsaninja, line 940: This seems to be slower based on a microbenchmark

src/typing_extensions.py
```diff
@@ -1321,7 +1321,7 @@ def __parameters__(self):
 # 3.7-3.9
 @typing._tp_cache
 def _concatenate_getitem(self, parameters):
-    if parameters == ():
+    if parameters is (): # empty tuple is singleton
```
hauntsaninja, line 1324: This creates a syntax warning and is less clearly correct, please undo

src/typing_extensions.py
```diff
@@ -1993,7 +1993,7 @@ def _caller():
             return None
 
     def _make_nmtuple(name, types, module, defaults=()):
-        fields = [n for n, t in types]
+        fields = (n for n, t in types)
```
hauntsaninja, line 1996: This will be faster as a list comprehension, please undo

src/typing_extensions.py
```diff
@@ -2028,7 +2028,7 @@ def __new__(cls, typename, bases, ns):
                                     f"{', '.join(default_names)}")
             nm_tpl = _make_nmtuple(
                 typename, types.items(),
-                defaults=[ns[n] for n in default_names],
+                defaults=(ns[n] for n in default_names),
```
hauntsaninja, line 2031: Same thing here

hauntsaninja (commented):
While most of these are fine, some regress both speed and readability, and at least one of these regresses functionality.

denballakh:
I agree with you, I reverted those changes in a new commit.

src/typing_extensions.py
```diff
@@ -97,7 +97,7 @@ def _check_generic(cls, parameters, elen=_marker):
     alen = len(parameters)
     if alen != elen:
         if hasattr(cls, "__parameters__"):
-            parameters = [p for p in cls.__parameters__ if not _is_unpack(p)]
+            parameters = (p for p in cls.__parameters__ if not _is_unpack(p))
```
AlexWaygood, line 100: Are you sure this is faster? As @hauntsaninja has pointed out elsewhere, list comprehensions are often faster than generator expressions if the list is small.

src/typing_extensions.py
```diff
@@ -352,27 +352,27 @@ def clear_overloads():
 TYPE_CHECKING = typing.TYPE_CHECKING
 
 
-_PROTO_WHITELIST = ['Callable', 'Awaitable',
+_PROTO_WHITELIST = {'Callable', 'Awaitable',
```
AlexWaygood, line 355: If we're changing this, why not make it a frozenset? It's never mutated.

src/typing_extensions.py
```diff
@@ -352,27 +352,27 @@ def clear_overloads():
...
 def _get_protocol_attrs(cls):
     attrs = set()
     for base in cls.__mro__[:-1]:  # without object
-        if base.__name__ in ('Protocol', 'Generic'):
+        if base.__name__ in {'Protocol', 'Generic'}:
             continue
         annotations = getattr(base, '__annotations__', {})
-        for attr in list(base.__dict__.keys()) + list(annotations.keys()):
-            if (not attr.startswith('_abc_') and attr not in (
+        for attr in (*base.__dict__.keys(), *annotations.keys()):
```
AlexWaygood, line 367: How much faster is `*`-unpacking here compared to list addition? Can you show some benchmarks?

AlexWaygood: Given that dictionary keys are set-like objects, the following is probably more idiomatic, and might be faster yet:

```suggestion
        for attr in base.__dict__.keys() | annotations.keys():
```

denballakh: There are benchmarks for `list(d1.keys()) + list(d2.keys())` (1), `(*d1.keys(), *d2.keys())` (2) and `d1.keys() | d2.keys()` (3).
I am using CPython 3.10.4.

For small dicts: (2) is 20-30% faster than (1), (3) is 12% faster than (1) if we dont iterate over result and is 11% slower than (1) if we do iterate over result.
For huge dicts: (2) is 5% than (1), (3) is much slower than (1).

So I prefer unpacking into tuple over list addition.

```py
Small dicts:
$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list(d1.keys()) + list(d2.keys())"
1000000 loops, best of 5: 315 nsec per loop

$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "(*d1.keys(), *d2.keys())"
1000000 loops, best of 5: 218 nsec per loop

$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "d1.keys() | d2.keys()"
1000000 loops, best of 5: 276 nsec per loop

Also measure time for iterating over result:
$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list(list(d1.keys()) + list(d2.keys()))"
1000000 loops, best of 5: 372 nsec per loop

$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list((*d1.keys(), *d2.keys()))"
1000000 loops, best of 5: 296 nsec per loop

$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list(d1.keys() | d2.keys())"
500000 loops, best of 5: 413 nsec per loop



Large dicts:
$>py -m timeit -s "d1 = dict.fromkeys(range(10000)); d2 = dict.fromkeys(range(10000,20000))" "list(d1.keys()) + list(d2.keys())"
1000 loops, best of 5: 165 usec per loop

$>py -m timeit -s "d1 = dict.fromkeys(range(10000)); d2 = dict.fromkeys(range(10000,20000))" "(*d1.keys(), *d2.keys())"
1000 loops, best of 5: 159 usec per loop

$>py -m timeit -s "d1 = dict.fromkeys(range(10000)); d2 = dict.fromkeys(range(10000,20000))" "d1.keys() | d2.keys()"
500 loops, best of 5: 720 usec per loop

Also measure time for iterating over result:
$>py -m timeit -s "d1 = dict.fromkeys(range(10000)); d2 = dict.fromkeys(range(10000,20000))" "list(list(d1.keys()) + list(d2.keys()))"
1000 loops, best of 5: 211 usec per loop

$>py -m timeit -s "d1 = dict.fromkeys(range(10000)); d2 = dict.fromkeys(range(10000,20000))" "list((*d1.keys(), *d2.keys()))"
1000 loops, best of 5: 202 usec per loop

$>py -m timeit -s "d1 = dict.fromkeys(range(10000)); d2 = dict.fromkeys(range(10000,20000))" "list(d1.keys() | d2.keys())"
500 loops, best of 5: 821 usec per loop
```

AlexWaygood: Thanks, that's a convincing benchmark! My suggestion definitely loses the speed race, in that case.

Looks like the fastest way yet, however, might be to drop the call to `.keys()`, which isn't actually necessary here at all:

```
$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list(list(d1.keys()) + list(d2.keys()))"
1000000 loops, best of 5: 364 nsec per loop

$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list((*d1.keys(), *d2.keys()))"
1000000 loops, best of 5: 296 nsec per loop

$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list(d1.keys() | d2.keys())"
500000 loops, best of 5: 415 nsec per loop

$>py -m timeit -s "d1 = {'a': 1, 'b': 2, 'c': 3}; d2 = {'x': 4, 'y': 5, 'z': 6}" "list((*d1, *d2))"
1000000 loops, best of 5: 208 nsec per loop
```

```suggestion
        for attr in (*base.__dict__, *annotations):
```

src/typing_extensions.py
denballakh: You're right. 
I will suggest another change: replace two iterables with one by inlining one into another:
```py
parameters = [p for p in cls.__parameters__ if not _is_unpack(p)]
num_tv_tuples = sum(isinstance(p, TypeVarTuple) for p in parameters)
# replace with:
num_tv_tuples = sum(isinstance(parameter, TypeVarTuple) for parameter in cls.__parameters__ if not _is_unpack(parameter))
```
Inlining gives us 10-20% speed-up.

Here are the benchmarks:
```py
# iterating over list
$>py -m timeit -s "pars=['a','b','cc','d','ee','ff'];cond=lambda x:len(x)%2==0" "it=[p for p in pars if cond(p)];sum(isinstance(p,str) for p in it)"
200000 loops, best of 5: 1.18 usec per loop

# iterating over generator
$>py -m timeit -s "pars=['a','b','cc','d','ee','ff'];cond=lambda x:len(x)%2==0" "it=(p for p in pars if cond(p));sum(isinstance(p,str) for p in it)"
200000 loops, best of 5: 1.26 usec per loop

# inline first generator into second
$>py -m timeit -s "pars=['a','b','cc','d','ee','ff'];cond=lambda x:len(x)%2==0" "sum(isinstance(p,str) for p in pars if cond(p))"
500000 loops, best of 5: 984 nsec per loop

$>py -m timeit -s "pars=[*map(str,range(10000))];cond=lambda x:len(x)%2==0" "it=[p for p in pars if cond(p)];sum(isinstance(p,str) for p in it)"
200 loops, best of 5: 1.52 msec per loop

$>py -m timeit -s "pars=[*map(str,range(10000))];cond=lambda x:len(x)%2==0" "it=(p for p in pars if cond(p));sum(isinstance(p,str) for p in it)"
200 loops, best of 5: 1.71 msec per loop

$>py -m timeit -s "pars=[*map(str,range(10000))];cond=lambda x:len(x)%2==0" "sum(isinstance(p,str) for p in pars if cond(p))"
200 loops, best of 5: 1.38 msec per loop
```

src/typing_extensions.py
denballakh: Didn't notice it :)

AlexWaygood (approved):

hauntsaninja:
@denballakh If you're willing to sign the CLA and fix CI, I will merge :-)

AlexWaygood:
There hasn't been a response from the OP in 3 months, the CLA is still unsigned, the CI is failing and there are merge conflicts. Closing for now, but if you're still interested in this PR, please feel free to open a new one @denballakh! You will need to sign the CLA, though :)

PR #58, opened 2022-07-15
Use inline table syntax for license key

srittau:
This fixes a problem with old pip version that use an outdated TOML
parser.

Closes: #57

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ version = "4.3.0"
 description = "Backported and Experimental Type Hints for Python 3.7+"
 readme = "README.md"
 requires-python = ">=3.7"
-license.file = "LICENSE"
+license = { file = "LICENSE" }
 keywords = [
     "annotations",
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
Do you think this is worth a hotfix release?

srittau:
I think this only relevant when pip installing the .tar.gz or from source, not when installing the wheel or from pypi, so I'm not sure how relevant this is. Also, I'm not sure if I'm right. :)

hauntsaninja (approved):
Thanks for figuring this one out! Pretty sure you're right, pip used to use pytoml, which only supports TOML 0.4. Dotted keys were a feature of 0.5

mweinelt:
Linux distros are regularly doing source builds, so this issue bit us. I will backport this patch manually, but other distros will probably also be affected.

hgoldfish:
the default python pip of debian 10 is took down by this minor issue.

PR #60, opened 2022-07-27
TypedDict: Resolve string-based annotations

### LINKED ISSUE #55: `Required`/`NotRequired` don't work with `__future__.annotations`
`Required`/`NotRequired` aren't properly detected when `from __future__ import annotations` is used, leading to incorrect `__required_keys__`/`__optional_keys__` sets.

Reproduced with 4.3.0.

```python
from __future__ import annotations  # comment this line out and things work correctly
from typing_extensions import NotRequired, TypedDict

class Test(TypedDict):
    a: int
    b: NotRequired[int]

assert Test.__required_keys__ == {"a"}
assert Test.__optional_keys__ == {"b"}
```

```shell
$ python test.py
  File "/home/jcristharif/Code/test.py", line 8, in <module>
    assert Test.__required_keys__ == {"a"}
AssertionError
```

d-k-bo:
Fixes #55 

I'm calling `typing.get_type_hints` because `typing_extensions.get_type_hints` is not yet defined in the relevant section. This shouldn't be a problem as long the behaviour of `get_type_hints(..., include_extras=True)` doesn't change. When porting this to the `typing` module, it should work fine because the definition order is fine there.

The test on Python 3.11 fails because `TypedDict` is rexported from `typing` (it took some time until I realized that this was the problem :D).

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -638,6 +638,10 @@ class AnnotatedMovie(TypedDict):
     title: Annotated[Required[str], "foobar"]
     year: NotRequired[Annotated[int, 2000]]
 
+class StringBasedMovie(TypedDict):
+    title: "str"
+    year: "NotRequired[int]"
+
 
 gth = get_type_hints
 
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
I don't think we should do this, since get_type_hints() can and will throw errors for some TypedDicts.

d-k-bo:
It seems like I underestimated the current limitations of `get_type_hints`.

Possible cases where it fails:

- imported type aliases that include forward references (described in https://github.com/python/cpython/issues/89089)
- using the union operator `|` with Python <3.10
- type imports hidden behind `if TYPE_CHECKING`

These are pretty common patterns and I can see that this would be unacceptable.

It might be better to evaluate the annotations on a per-item basis with a fallback in case of failure.

PR #67, opened 2022-08-10
Remove deprecated py36 from README

MoshiBin:
Python 3.6 is not supported since version 4.2.0, but it's still listed in the readme (and therefore also on PyPI)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The `typing_extensions` module serves two related purposes:
 
 - Enable use of new type system features on older Python versions. For example,
   `typing.TypeGuard` is new in Python 3.10, but `typing_extensions` allows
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,7 @@ The `typing_extensions` module serves two related purposes:
 
 - Enable use of new type system features on older Python versions. For example,
   `typing.TypeGuard` is new in Python 3.10, but `typing_extensions` allows
-  users on Python 3.6 through 3.9 to use it too.
+  users on Python 3.7 through 3.9 to use it too.
```
hauntsaninja, line 11:
```suggestion
  users on previous Python versions to use it too.
```
Maybe less busy work for whenever 3.7 dies? ;-)

PR #68, opened 2022-08-14
Implement typing_extensions.Any

Gobot1234:
Backport from 3.11+ version (also with fix from https://github.com/python/cpython/issues/95987)

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -171,7 +171,7 @@ class AnyTests(BaseTestCase):
             mod_name = 'typing_extensions'
         self.assertEqual(repr(Any), f"{mod_name}.Any")
         if sys.version_info < (3, 11):  # skip for now on 3.11+ see python/cpython#95987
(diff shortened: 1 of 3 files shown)


srittau (approved):
LGTM, except for the minor flake8 error, but I'd like @hauntsaninja's feedback.

src/test_typing_extensions.py
```diff
@@ -160,6 +160,32 @@ def test_exception(self):
             assert_never(None)
 
 
+class AnyTests(BaseTestCase):
+    class SubclassesAny(Any):
```
hauntsaninja, line 164: Let's put this into an actual test case?
Wouldn't hurt to copy over the tests from `test_can_subclass` in the original PR; I seem to recall there was some trickiness wrt inheritance, so might as well test it.

hauntsaninja (approved):
Thanks, this looks good to me, just had one nit about testing

PR #77, opened 2022-10-02
Initial support for TypeVarLike default parameter (PEP 696)

### LINKED ISSUE #75: Initial runtime support for PEP 696 (TypeVarLike defaults)
I would like to start adding runtime support for [PEP 696 ](https://peps.python.org/pep-0696/). If the `default` parameter is added to `TypeVar`, `ParamSpec`, and `TypeVarTuple`, type checkers could start adding provisional support.

As I understand it, there exists a reference implementation for mypy already. It would be awesome, if we could start using that even before the PEP is formally accepted. With the 0.981 release for example, mypy added support for [PEP 692](https://peps.python.org/pep-0692/) although it hasn't been approved yet.

Before I start working on it, would a PR be accepted?

/CC: @Gobot1234

cdce8p:
Closes #75
Ref: https://peps.python.org/pep-0696/

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -2463,18 +2463,20 @@ class ParamSpecTests(BaseTestCase):
             pass
 
     def test_pickle(self):
-        global P, P_co, P_contra
+        global P, P_co, P_contra, P_default
         P = ParamSpec('P')
         P_co = ParamSpec('P_co', covariant=True)
         P_contra = ParamSpec('P_contra', contravariant=True)
+        P_default = ParamSpec('P_default', default=int)
         for proto in range(pickle.HIGHEST_PROTOCOL):
             with self.subTest(f'Pickle protocol {proto}'):
-                for paramspec in (P, P_co, P_contra):
+                for paramspec in (P, P_co, P_contra, P_default):
                     z = pickle.loads(pickle.dumps(paramspec, proto))
                     self.assertEqual(z.__name__, paramspec.__name__)
                     self.assertEqual(z.__covariant__, paramspec.__covariant__)
                     self.assertEqual(z.__contravariant__, paramspec.__contravariant__)
                     self.assertEqual(z.__bound__, paramspec.__bound__)
+                    self.assertEqual(z.__default__, paramspec.__default__)
 
     def test_eq(self):
         P = ParamSpec('P')
@@ -2839,6 +2841,17 @@ class TypeVarTupleTests(BaseTestCase):
         t = Tuple[tuple(Ts)]
         self.assertEqual(t.__args__, (Unpack[Ts],))
         self.assertEqual(t.__parameters__, (Ts,))
+
+    def test_pickle(self):
+        global Ts, Ts_default  # pickle wants to reference the class by name
+        Ts = TypeVarTuple('Ts')
+        Ts_default = TypeVarTuple('Ts_default', default=Unpack[Tuple[int, str]])
+
+        for proto in range(pickle.HIGHEST_PROTOCOL):
+            for typevartuple in (Ts, Ts_default):
+                z = pickle.loads(pickle.dumps(typevartuple, proto))
+                self.assertEqual(z.__name__, typevartuple.__name__)
+                self.assertEqual(z.__default__, typevartuple.__default__)
 
 
 class FinalDecoratorTests(BaseTestCase):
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -1147,6 +1148,42 @@ def __repr__(self):
                                above.""")
 
 
+class _DefaultMixin:
+    """Mixin for TypeVarLike defaults."""
+
+    def __init__(self, default):
+        if isinstance(default, tuple):
```
Gobot1234, line 1155:
```suggestion
        if isinstance(default, (tuple, list)):
```
Whilst tuples are the internal ways these are stored mypy shows ParamSpec args as lists so I think that would be good to support

Gobot1234 (changes requested):
Not entirely sure how I feel about these subclassing the typing versions if possible but apart from the one comment LGTM

cdce8p:
> Not entirely sure how I feel about these subclassing the typing versions if possible

I agree. When I first experimented with it, it felt strange. However, with subclassing we don't need to worry about the base implementation and avoid reimplementing that again.

Gobot1234:
> > Not entirely sure how I feel about these subclassing the typing versions if possible
> 
> I agree. When I first experimented with it, it felt strange. However, with subclassing we don't need to worry about the base implementation and avoid reimplementing that again.

I hadn't considered that, although it would still probably be a good idea to sys.version_info checks to make sure that you aren't redefining the behaviour if TypeVarLikes already have the default argument

src/typing_extensions.py
```diff
@@ -1147,6 +1148,42 @@ def __repr__(self):
                                above.""")
 
 
+class _DefaultMixin:
```
Gobot1234, line 1151: This class should define empty slots to make TypeVar not have __dict__

src/test_typing_extensions.py
```diff
@@ -3395,5 +3398,28 @@ def test_same_as_typing_NamedTuple_38_minus(self):
         )
 
 
+class TypeVarLikeDefaultsTests(BaseTestCase):
+    def test_typevar(self):
+        T = typing_extensions.TypeVar("T", default=int)
```
JelleZijlstra, line 3403: Add tests for pickling a defaulted TypeVar.

src/typing_extensions.py
```diff
@@ -1246,14 +1252,18 @@ def __eq__(self, other):
                 return NotImplemented
             return self.__origin__ == other.__origin__
 
+
+if sys.version_info >= (3, 12):
+    ParamSpec = typing.ParamSpec
+
 # 3.10+
 if hasattr(typing, 'ParamSpec'):
```
AlexWaygood, line 1260:
```suggestion
elif hasattr(typing, 'ParamSpec'):
```

src/typing_extensions.py
cdce8p: Done. Although I'm not sure it will work. For 3.10 `_TypeVarLike` and for 3.11 `_BoundVarianceMixin` don't add `__slots__`, so the `__dict__` for `TypeVar` is created anyway.

src/typing_extensions.py
```diff
@@ -1147,6 +1148,48 @@ def __repr__(self):
...
+        if isinstance(default, (tuple, list)):
+            self.__default__ = tuple((typing._type_check(d, "Default must be a type")
+                                      for d in default))
+        elif default:
+            self.__default__ = typing._type_check(default, "Default must be a type")
+        else:
+            self.__default__ = None
+
+
+if sys.version_info >= (3, 12):
```
JelleZijlstra, line 1166: We don't know yet that defaults will be in 3.12, and it definitely won't be in the first alpha (which I'm told is getting released this week). Let's unconditionally export our version of TypeVar/ParamSpec/TypeVarTuple even on 3.12.

cdce8p: I agree, we can always push a new commit later once the PEP has been accepted. Pushed a new commit to revert these changes.

src/typing_extensions.py
JelleZijlstra: If you're interested, we can fix that for 3.12. (I'd be hesitant for older versions since it may break some users.)

JelleZijlstra (approved):

src/typing_extensions.py
cdce8p: Seems like `__slots__` was removed on purpose since it didn't have any use. I've updated the PR to remove it from `TypeVar` as well. Ref: https://github.com/python/cpython/pull/30444

Gobot1234:
Just realised that the default for the default argument shouldn't be None and this needs a new object that should be the default as you currently can't make a distinction between default=None and default=empty

JelleZijlstra:
I expect this just wasn't considered, and it's never come up as a problem because a bound of `None` has no practical use.

(Then again, pyright just got a bug report asking for `NewType(None)` to work, so maybe someone has a use for subtypes of None.)

PR #78, opened 2022-10-03
Add @typing.override

JelleZijlstra:


src/typing_extensions.py
```diff
@@ -2000,6 +2001,34 @@ def decorator(cls_or_fn):
         return decorator
 
 
+if hasattr(typing, "override"):
+    override = typing.override
+else:
+    _T = typing.TypeVar("_T", bound=typing.Callable[..., typing.Any])
```
hauntsaninja, line 2007: nit: _F is a common name for this typevar and _T exists elsewhere in this file
```suggestion
    _F = typing.TypeVar("_F", bound=typing.Callable[..., typing.Any])
```

hauntsaninja (approved):
Looks good, assuming you don't want any runtime recording of override

JelleZijlstra:
> Looks good, assuming you don't want any runtime recording of override

Thanks! The PEP specifies that there's no runtime effect, so that's what we're doing.

JelleZijlstra:
@stroxler could you take a look? The docstring is probably the only interesting part.

src/typing_extensions.py
```diff
@@ -2000,6 +2001,36 @@ def decorator(cls_or_fn):
...
+
+            class Child(Base):
+                @override
+                def method(self) -> None:
+                    super().method()
+
+        When this decorator is applied to a method, the type checker will
+        validate that it overrides a method with the same name on a base class.
+        This helps prevent bugs that may occur when the base class is changed
+        without an equivalent change to a child class.
```
stroxler, line 2026: Nit: I think "the base" and "a child" sounds funny to me. I'd probably us "a" in both places:
```suggestion
        This helps prevent bugs that may occur when a base class is changed
        without an equivalent change to a child class.
```

stroxler (approved):
Looks good to me, thanks @JelleZijlstra!

src/typing_extensions.py
JelleZijlstra: Thanks, good point

src/typing_extensions.py
JelleZijlstra: Done, thanks

src/typing_extensions.py
```diff
@@ -2000,6 +2001,36 @@ def decorator(cls_or_fn):
         return decorator
 
 
+if hasattr(typing, "override"):
+    override = typing.override
+else:
+    _F = typing.TypeVar("_F", bound=typing.Callable[..., typing.Any])
+
+    def override(__arg: _F) -> _F:
```
cdce8p, line 2009: Just asking, my understanding was that types should not be added here and only provided via typeshed. I can't find a reference for it unfortunately. Did I miss something?

hauntsaninja: It is true that type checkers will not look at this code. But types have value as documentation and perhaps some niche value for dynamic introspection.

AlexWaygood (approved):
Looks good! Some minor suggestions on docs/tests.

src/typing_extensions.py
```diff
@@ -2078,6 +2079,36 @@ def decorator(cls_or_fn):
...
+
+            class Child(Base):
+                @override
+                def method(self) -> None:
+                    super().method()
+
+        When this decorator is applied to a method, the type checker will
+        validate that it overrides a method with the same name on a base class.
+        This helps prevent bugs that may occur when a base class is changed
+        without an equivalent change to a child class.
```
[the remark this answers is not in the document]
JelleZijlstra: I prefer the current more general wording. For example, the base class method could be removed instead of renamed.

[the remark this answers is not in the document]
AlexWaygood: Oh, good point.

PR #80, opened 2022-10-05
Add TypeVar(infer_variance=True)

### LINKED ISSUE #79: Add support for TypeVar(infer_variance=True) (PEP 695)
Most of PEP 695 isn't relevant to typing-extensions, but the new infer_variance parameter can be backported. We should add it.

cc @erictraut

JelleZijlstra:
From PEP 695. Fixes #79.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Release 4.4.0 ()
+
 - Add `typing_extensions.Any` a backport of python 3.11's Any class which is
   subclassable at runtime. (backport from python/cpython#31841, by Shantanu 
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
I decided here not to add `infer_variance` support to `ParamSpec`. `ParamSpec` supports covariant=/contravariant= at runtime, but the behavior is unspecified. PEP 695 only mentions the `infer_variance` parameter for `TypeVar`.

AlexWaygood (approved):
LGTM other than maybe needing an addition to the README!

JelleZijlstra:
Thanks! I'll improve the README in a separate PR.

Issue #83, opened 2022-10-07
Test regression with pypy3.9: test_overload_on_compiled_functions


mgorny:
```pytb
$ pypy3 --version
Python 3.9.12 (05fbe3aa5b0845e6c37239768aa455451aa5faba, Sep 05 2022, 05:41:51)
[PyPy 7.3.9 with GCC 12.2.1 20220903]
$ tox -e pypy3
pypy3 inst-nodeps: /tmp/typing_extensions/.tox/.tmp/package/1/typing_extensions-4.4.0.tar.gz
pypy3 installed: cffi==1.15.0,greenlet==0.4.13,hpy==0.0.3,readline==6.2.4.1,typing_extensions @ file:///tmp/typing_extensions/.tox/.tmp/package/1/typing_extensions-4.4.0.tar.gz
pypy3 run-test-pre: PYTHONHASHSEED='150667971'
pypy3 run-test: commands[0] | python -m unittest discover
..................s...........................................................................................s.s..s...............................F.................................................Runtime type is 'object'
......................................................s.........
======================================================================
FAIL: test_overload_on_compiled_functions (test_typing_extensions.OverloadTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/pypy3.9/unittest/mock.py", line 1336, in patched
    return func(*newargs, **newkeywargs)
  File "/tmp/typing_extensions/src/test_typing_extensions.py", line 526, in test_overload_on_compiled_functions
    self.assertEqual(get_overloads(sum), [])
AssertionError: Lists differ: [<built-in function sum>] != []

First list contains 1 additional elements.
First extra element 0:
<built-in function sum>

- [<built-in function sum>]
+ []

----------------------------------------------------------------------
Ran 261 tests in 0.253s

FAILED (failures=1, skipped=5)
ERROR: InvocationError for command /tmp/typing_extensions/.tox/pypy3/bin/python -m unittest discover (exited with code 1)
_______________________________________________________________ summary _______________________________________________________________
ERROR:   pypy3: commands failed
```

srittau:
One preliminary note: As the implementation of `overload` matches the implementation of `overload` in Python 3.11, pypy would also fail with the `typing.py` module from there.

srittau:
The problem is that `sum()` and `print()` are not compiled functions in pypy. The test in question checks whether `@overload` is skipped for those, so it can't work. #84 just skips this tests when running pypy.

mgorny:
[...]

PR #84, opened 2022-10-07
Fix tests when running pypy

srittau:
Closes: #83

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,10 +23,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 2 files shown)


hauntsaninja (approved):

.github/workflows/ci.yml
```diff
@@ -18,15 +18,15 @@ jobs:
         # Python version, because typing sometimes changed between bugfix releases.
         # For available versions, see:
         # https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
-        python-version: ["3.7", "3.7.1", "3.8", "3.8.0", "3.9", "3.9.0", "3.10", "3.10.0", "3.11-dev"]
+        python-version: ["3.7", "3.7.1", "3.8", "3.8.0", "3.9", "3.9.0", "3.10", "3.10.0", "3.11-dev", "pypy3.9"]
```
mgorny, line 21: You may want to also test `pypy3.8` and (perhaps) `pypy3.7`, as all three branches are still being maintained.

srittau: Let's skip the other pypy versions for now to save on those precious CI minutes. (The Python org ran into troubles before IIRC.)

PR #93, opened 2022-11-11
Remove Python 3.6 from tox.ini

carljm:
The 3.6 tox env fails because the minimum supported version is now 3.7.


JelleZijlstra:
Thanks!

PR #94, opened 2022-11-11
Fix get_type_hints() on x-module inherited TypedDict in 3.9 and 3.10

carljm:
The fix in https://github.com/python/cpython/pull/27017 was never backported into typing-extensions.

In 3.11 it doesn't matter because we just use `TypedDict` from `typing.py`. But in 3.9 and 3.10, it means this cross-module inheritance scenario with stringified annotations and `get_type_hints` will fail if you use `TypedDict` from `typing_extensions` but work if you use it from `typing`.

I don't attempt here to backport the fix to pre-3.9 versions where `typing.py` doesn't have it; that would require pulling in lots more code (`ForwardRef` and some utility functions) into `typing_extensions.py`. Just aiming for parity with `typing.py` for 3.9 and 3.10.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -41,6 +41,10 @@ TYPING_3_10_0 = sys.version_info[:3] >= (3, 10, 0)
 # 3.11 makes runtime type checks (_type_check) more lenient.
 TYPING_3_11_0 = sys.version_info[:3] >= (3, 11, 0)
 
+# https://github.com/python/cpython/pull/27017 was backported into some 3.9 and 3.10
+# versions, but not all
+HAS_FORWARD_MODULE = "module" in inspect.signature(typing._type_check).parameters
+
 
 class BaseTestCase(TestCase):
     def assertIsSubclass(self, cls, class_or_tuple, msg=None):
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -1978,6 +1986,13 @@ class PointDict3D(PointDict2D, total=False):
         assert is_typeddict(PointDict2D) is True
         assert is_typeddict(PointDict3D) is True
 
+    @skipUnless(HAS_FORWARD_MODULE, "ForwardRef.__forward_module__ was added in 3.9")
+    def test_get_type_hints_cross_module_subclass(self):
+        self.assertEqual(
```
AlexWaygood, line 1991: Drive-by suggestion: you could maybe add an assertion here to verify that the test is working as intended, and `_DoNotImport` has not been imported 

```suggestion
        # Test the test: _DoNotImport cannot be in globals() or locals(),
        # or the second assertion will pass for the wrong reasons
        self.assertNotIn("_DoNotImport", {**globals(), **locals()})
        self.assertEqual(
```

carljm: Done!

JelleZijlstra (approved):

PR #100, opened 2022-11-25
CI: test on Ubuntu-20.04, not Ubuntu-2004

AlexWaygood:
Fixes a typo in https://github.com/python/typing_extensions/commit/5f9258d1efd5d8e4900ffc23ba2486240f2a2f35. We might have to wait a while before Ubuntu-2004 is released :)


JelleZijlstra:
Thanks!

PR #101, opened 2022-12-03
Add `frozen_default` parameter on `dataclass_transform`

debonte:
Mirroring the `typing.py` change at https://github.com/python/cpython/pull/99958.

@JelleZijlstra


hauntsaninja (approved):
Looks great! Jelle didn't have any typing-extensions relevant feedback on the CPython PR, so I'll go ahead and merge :-)

PR #105, opened 2022-12-31
Add @typing_extensions.deprecated

JelleZijlstra:
See PEP 702, python/peps#2942. This should not be merged until the PEP
is merged into the repo.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -52,6 +52,7 @@ __all__ = [
     'assert_type',
     'clear_overloads',
     'dataclass_transform',
+    'deprecated',
     'get_overloads',
     'final',
     'get_args',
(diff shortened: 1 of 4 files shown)


src/typing_extensions.py
```diff
@@ -2129,6 +2130,47 @@ def method(self) -> None:
...
+            @deprecated("Use B instead")
+            class A:
+                pass
+
+            @deprecated("Use g instead")
+            def f():
+                pass
+
+            @deprecated("int support is deprecated")
+            @overload
```
JelleZijlstra, line 2152:
```suggestion
            @overload
            @deprecated("int support is deprecated")
```

src/typing_extensions.py
```diff
@@ -2129,6 +2130,47 @@ def method(self) -> None:
...
+            def g(x: int) -> int: ...
+            @overload
+            def g(x: str) -> int: ...
+
+        When this decorator is applied to an object, the type checker
+        will generate a diagnostic on usage of the deprecated object.
+
+        No runtime warning is issued. The decorator sets the ``__deprecated__``
+        attribute on the decorated object to the deprecation message
+        passed to the decorator.
```
hauntsaninja, line 2162: Should this more clearly call out the behaviour of `__deprecated__` with overloads?

hauntsaninja (commented):
Looks reasonable, just had one question

src/typing_extensions.py
JelleZijlstra: How is it now?

hauntsaninja (approved):
Thanks! The new words looks good to me.

PR #112, opened 2023-02-05
PEP 702: Runtime warnings

JelleZijlstra:
See python/peps#3006.


hauntsaninja (approved):
This looks great!

I was a little surprised by the choice of `category=None` as a way to spell no warning, since with `warnings.warn`, `category=None` is UserWarning. But I think it's fine, the behaviour will be clearly visible + the documentation will make clear that `category=None` will disable the warning.

hauntsaninja:
If you wanted, could add a test for a method as well

PR #128, opened 2023-03-28
Move builtin protocol whitelist to mapping instead of list

YouJiacheng:
Backport from:
https://github.com/python/cpython/pull/15647
Also rename to allowlist, following these changes:
https://github.com/python/cpython/pull/21823
https://github.com/python/cpython/pull/21825

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1909,7 +1909,6 @@ class ProtocolTests(BaseTestCase):
 
         self.assertIsSubclass(B, Custom)
         self.assertNotIsSubclass(A, Custom)
-    
     def test_builtin_protocol_allowlist(self):
         with self.assertRaises(TypeError):
             class CustomProtocol(TestCase, Protocol):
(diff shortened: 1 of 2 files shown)


src/test_typing_extensions.py
```diff
@@ -1909,6 +1909,14 @@ def close(self):
 
         self.assertIsSubclass(B, Custom)
         self.assertNotIsSubclass(A, Custom)
+
```
AlexWaygood, line 1912: This should fix the linting failure in CI:

```suggestion

```

src/test_typing_extensions.py
```diff
@@ -1909,6 +1909,13 @@ def close(self):
 
         self.assertIsSubclass(B, Custom)
         self.assertNotIsSubclass(A, Custom)
+    def test_builtin_protocol_allowlist(self):
```
AlexWaygood, line 1912: Argh, sorry -- you _should_ have an empty line here (the issue was there used to be trailing whitespace). My mistake; I must have screwed up the GitHub suggestion

```suggestion

    def test_builtin_protocol_allowlist(self):
```

YouJiacheng: Oh, I am not familiar with GitHub diff. I assumed, wrongly, that your previous suggestion delete the whitespaces only(instead of whole line).

AlexWaygood (approved):
LGTM. The added test does indeed fail with the current latest version of `typing_extensions` on Python 3.7, and this is the fix that was applied in CPython.

JelleZijlstra (approved):

PR #132, opened 2023-04-09
Backport CPython PR 26067

AlexWaygood:
This backports the CPython PR https://github.com/python/cpython/pull/26067, which was backported to Python 3.10 but no earlier in the CPython repo.

(It was initially backported to 3.9, but the backport was reverted: https://github.com/python/cpython/pull/26077.)

Since we're now backporting a bugfix to <=3.9, this means that we re-implement `Protocol`/`runtime_checkable` on <=3.9, rather than on just 3.7. Doing this requires a change to `_get_protocol_attrs`; I also tweaked `typing_extensions.runtime_checkable` so that it also works on `typing.Protocol`, if `typing.Protocol` exists.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -289,7 +289,7 @@ else:
                            instead of a type.""")
 
 
-_overload_dummy = typing._overload_dummy  # noqa
+_overload_dummy = typing._overload_dummy
 
 
 if hasattr(typing, "get_overloads"):  # 3.11+
@@ -497,7 +497,7 @@ else:
         if type(self)._is_protocol:
             raise TypeError('Protocols cannot be instantiated')
 
-    class _ProtocolMeta(abc.ABCMeta):  # noqa: B024
+    class _ProtocolMeta(abc.ABCMeta):
         # This metaclass is a bit unfortunate and exists only because of the lack
         # of __instancehook__.
         def __instancecheck__(cls, instance):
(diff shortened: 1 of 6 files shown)


AlexWaygood:
Not sure why that test is failing only on 3.9; it's a bit weird. I'll try to take another look tomorrow, but any suggestions are welcome :)

AlexWaygood:
> Not sure why that test is failing only on 3.9; it's a bit weird.

I figured it out -- `__class_getitem__` was being counted as a protocol member on Python 3.9. All green now!

src/typing_extensions.py
```diff
@@ -581,12 +609,12 @@ def _proto_hook(other):
                 if not cls.__dict__.get('_is_protocol', None):
                     return NotImplemented
                 if not getattr(cls, '_is_runtime_protocol', False):
-                    if _caller(depth=3) in {'abc', 'functools'}:
+                    if _allow_reckless_class_checks():
```
srittau, line 612: Is it correct that the depth is changed from 3 to 4?

src/typing_extensions.py
```diff
@@ -581,12 +609,12 @@ def _proto_hook(other):
...
                         return NotImplemented
                     raise TypeError("Instance and class checks can only be used with"
                                     " @runtime protocols")
                 if not _is_callable_members_only(cls):
-                    if _caller(depth=3) in {'abc', 'functools'}:
+                    if _allow_reckless_class_checks():
```
srittau, line 617: Here as well?

srittau (approved):
My understanding of typing internals is limited, but this LGTM, apart from the comment about flake8 and a question below.

src/typing_extensions.py
AlexWaygood: Yes, because by adding a new function that the code has to pass through before it gets to the `sys._getframe` call, the call stack becomes "another frame deep"

src/typing_extensions.py
AlexWaygood: Same as above

src/typing_extensions.py
srittau: I could just have looked at `_caller` and figured that out myself. 🤦

AlexWaygood:
Thanks!

PR #136, opened 2023-04-12
Backport test coverage improvements for runtime-checkable protocols

AlexWaygood:
This backports the following two CPython PRs (both by me):

- https://github.com/python/cpython/pull/102449
- https://github.com/python/cpython/pull/103347


srittau (approved):
The tests pass, nothing glaring stands out, and it's in CPython already.

PR #143, opened 2023-04-13
State in README that we backport the 3.12 version of `Protocol`, `runtime_checkable` and various runtime-checkable protocols

AlexWaygood:


README.md
```diff
@@ -143,6 +144,13 @@ Certain objects were changed after they were added to `typing`, and
 - `TypeVar` gains two additional parameters, `default=` and `infer_variance=`,
   in the draft PEPs [695](https://peps.python.org/pep-0695/) and [696](https://peps.python.org/pep-0696/), which are being considered for inclusion
   in Python 3.12.
+- `Protocol` was added in Python 3.8, but several bugfixes have been made in
+  subsequent releases, as well as significant performance improvements to
+  runtime-checkable protocols in Python 3.12. `typing_extensions` backports the
+  3.12+ version to Python 3.7+.
+- `SupportsIndex` was added in Python 3.8, but `isinstance()` checks against
```
JelleZijlstra, line 151: What's special about this one relative to all the other Supports* protocols?

AlexWaygood: The special thing is that it was added in 3.8, meaning we already provided a backport for it, meaning it was a 2-line change to provide a 10x speedup :)

We could consider reimplementing faster versions of all the other typing-module protocols as well, but we don't currently provide backports for any of the other ones (since they're available in `typing` on 3.7+)

JelleZijlstra: I see. This is going to be hard to explain to users though, especially after we drop support for 3.7. I'm not sure we should be backporting pure performance improvements at all.

AlexWaygood: I don't feel too strongly about it if you'd like to revert #141. It just felt like a very easy change to make, given that there was already a backport in place :)

JelleZijlstra: I'd rather do that honestly. I wouldn't be too opposed to exposing optimized versions of all the protocols, but doing it for just one is going to be confusing for users.

JelleZijlstra: cc @srittau who merged #141

AlexWaygood: I'm also happy to make a PR implementing optimised versions of all the protocols. No strong opinion either way; happy to defer to your judgement.

PR #146, opened 2023-04-17
Add faster versions of various runtime-checkable protocols

AlexWaygood:
This is one way of resolving the disagreement in #143. The other way would be to just revert #141. No strong opinion either way :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,11 @@
   Patch by Alex Waygood.
 - Speedup `isinstance(3, typing_extensions.SupportsIndex)` by >10x on Python
   <3.12. Patch by Alex Waygood.
+- Add `typing_extensions` versions of `SupportsInt`, `SupportsFloat`,
+  `SupportsComplex`, `SupportsBytes`, `SupportsAbs` and `SupportsRound`. These
+  have the same semantics as the versions from the `typing` module, but
+  `isinstance()` checks against the `typing_extensions versions are >10x faster
+  at runtime on Python <3.12. Patch by Alex Waygood.
 
 # Release 4.5.0 (February 14, 2023)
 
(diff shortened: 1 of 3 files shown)


srittau:
Looking back at #141, I'm not sure how relevant the speedups are and whether it's worth complicating typing_extensions for it. But I don't feel strongly about either. So, -0 and +0 on reverting #141.

AlexWaygood:
> I'm not sure how relevant the speedups are

I'd argue the speedups are quite relevant. The optimisations we've introduced into Python 3.12 were designed to address a longstanding user report where `_ProtocolMeta.__instancecheck__` was revealed to be a significant bottleneck in their code: https://github.com/python/cpython/issues/74690. Elsewhere, the slowness of `_ProtocolMeta.__instancecheck__` has led the beartype library to completely reimplement `_ProtocolMeta` in order to address the performance issues: https://github.com/beartype/beartype/blob/main/beartype/typing/_typingpep544.py.

I think it's fine to decide that we don't want to bloat the API of typing_extensions, or that it's not our job to backport things just to make them faster. But I'll defend the relevance of the optimisations here to our users; I think the speedup addresses a real problem that's been a common pain point for users of typing.

CHANGELOG.md
```diff
@@ -49,6 +49,11 @@
   Patch by Alex Waygood.
 - Speedup `isinstance(3, typing_extensions.SupportsIndex)` by >10x on Python
   <3.12. Patch by Alex Waygood.
+- Add `typing_extensions` versions of `SupportsInt`, `SupportsFloat`,
+  `SupportsComplex`, `SupportsBytes`, `SupportsAbs` and `SupportsRound`. These
+  have the same semantics as the versions from the `typing` module, but
+  `isinstance()` checks against the `typing_extensions versions are >10x faster
```
JelleZijlstra, line 55:
```suggestion
  `isinstance()` checks against the `typing_extensions` versions are >10x faster
```

JelleZijlstra (approved):
Thinking about it more I'm OK with merging this. Performance improvements are new features, and typing-extensions is about backporting new features.

PR #148, opened 2023-04-20
Reimplement `Literal` on Python <=3.10.0

### LINKED ISSUE #147: Backport fixes to Literal from 3.11?
Hello, I see @AlexWaygood recently backported some fixes to `Literal` for Python versions prior to 3.9.1. Might you consider doing the same for some further fixes (last commit [here](https://github.com/python/cpython/commit/634984d7dbdd91e0a51a793eed4d870e139ae1e0)) that only appeared in 3.11? Those issues -- basically problems with the type cache causing `Literal[True]` to return `Literal[1]` or vice versa when both are used -- break my code since I use annotations for runtime typechecking. If you could put the 3.11 implementation of `Literal` in `typing_extensions` then everybody could use it on older Python versions without needing their own monkeypatch to fix this problem. Thanks!

(If you would like I can try putting together a PR for this, although I'm not sure I understand the differences between your version of `Literal` and the 3.11 one well enough to do it properly.)

AlexWaygood:
Fixes #147.

No code changes are required -- the `typing_extensions` reimplementation of `Literal` has never suffered from the same bug that was fixed in https://github.com/python/cpython/pull/29334, because we don't cache `Literal` subscripts in the same way that CPython does. (I think if we wanted to do so, it might require us to reimplement `@typing._tp_cache` as it exists on CPython `main`, since the signature of `_tp_cache` changed between bugfix versions at some point.) As such, all we need to do is reimplement `Literal` on <=3.10.0 instead of <=3.9.0.

Cc. @dfremont :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,9 +143,10 @@ Certain objects were changed after they were added to `typing`, and
 - `TypeVar` gains two additional parameters, `default=` and `infer_variance=`,
   in the draft PEPs [695](https://peps.python.org/pep-0695/) and [696](https://peps.python.org/pep-0696/), which are being considered for inclusion
   in Python 3.12.
-- `Literal` does not flatten or deduplicate parameters on Python <3.9.1. The
-  `typing_extensions` version flattens and deduplicates parameters on all
-  Python versions.
+- `Literal` does not flatten or deduplicate parameters on Python <3.9.1, and a
+  caching bug was fixed in 3.10.1/3.9.8 The `typing_extensions` version
+  flattens and deduplicates parameters on all Python versions, and the caching
+  bug is also fixed on all versions.
 
 There are a few types whose interface was modified between different
 versions of typing. For example, `typing.Sequence` was modified to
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -143,9 +143,10 @@ Certain objects were changed after they were added to `typing`, and
 - `TypeVar` gains two additional parameters, `default=` and `infer_variance=`,
   in the draft PEPs [695](https://peps.python.org/pep-0695/) and [696](https://peps.python.org/pep-0696/), which are being considered for inclusion
   in Python 3.12.
-- `Literal` does not flatten or deduplicate parameters on Python <3.9.1. The
-  `typing_extensions` version flattens and deduplicates parameters on all
-  Python versions.
+- `Literal` does not flatten or deduplicate parameters on Python <3.9.1, and a
+  caching bug was fixed in 3.10.1/3.9.8 The `typing_extensions` version
```
JelleZijlstra, line 147:
```suggestion
  caching bug was fixed in 3.10.1/3.9.8. The `typing_extensions` version
```

JelleZijlstra (approved):

PR #150, opened 2023-04-21
Add __orig_bases__ to TypedDict

adriangb:


adriangb:
This should be up to date with the [CPython change which just got merged](https://github.com/python/cpython/pull/103698)

src/test_typing_extensions.py
```diff
@@ -2260,7 +2253,7 @@ def test_basics_iterable_syntax(self):
         self.assertEqual(Emp.__total__, True)
 
     def test_basics_keywords_syntax(self):
-        with self.assertWarnsIf(sys.version_info >= (3, 11), DeprecationWarning):
+        with self.assertWarns(DeprecationWarning):
```
JelleZijlstra, line 2256: I don't think we want this change and a few others below.

JelleZijlstra (commented):
Thanks! Could you also add a CHANGELOG entry? Look at some of Alex's recent PRs for examples.

AlexWaygood (commented):
We need to change these lines so that we reimplement NamedTuple on <=3.11 rather than <=3.10. Otherwise call-based `typing_extensions.NamedTuple`s will have `__orig_bases__` on 3.7-3.10 and 3.12+, but not 3.11 😄

https://github.com/python/typing_extensions/blob/fb37b2ee0ab56baf2dfd5df27aec0430435ba17f/src/typing_extensions.py#L2316-L2320

`test_typing_extensions_defers_when_possible` will need to be adjusted as a result -- you'll need to move `"NamedTuple"` into the `if sys.version_info < (3, 12)` set here:

https://github.com/python/typing_extensions/blob/fb37b2ee0ab56baf2dfd5df27aec0430435ba17f/src/test_typing_extensions.py#L3610-L3613

adriangb:
> Could you also add a CHANGELOG entry?

Done, I copied Alex's changelog note from CPython.

src/typing_extensions.py
```diff
@@ -813,6 +813,15 @@ def _typeddict_new(*args, total=True, **kwargs):
             raise TypeError("TypedDict takes either a dict or keyword arguments,"
                             " but not both")
 
+        if kwargs:
+            warnings.warn(
```
AlexWaygood, line 817: I think it's important that we copy CPython's behaviour here, and Cpython emits a deprecation warning here on 3.11+. @JelleZijlstra do you think we should only emit the deprecation warning if the user is running typing_extensions on 3.11+?

JelleZijlstra: That's a hard question and gets into what we would do if we ever want to break backwards compatibility for typing-extensions. My first instinct is to say we should always warn in typing-extensions, regardless of the version, but then what would we do when CPython removes support for kwargs-based TypedDicts? I'd be hesitant to remove the runtime behavior in typing-extensions and break backwards compatibility.

AlexWaygood: I agree it's a hard question.

Elsewhere in the `typing_extensions` implementation of `TypedDict`, we actually already have two places where CPython  long ago removed support for a certain feature, but `typing_extensions` has in effect had "eternal deprecation warnings":

https://github.com/python/typing_extensions/blob/fb37b2ee0ab56baf2dfd5df27aec0430435ba17f/src/typing_extensions.py#L789-L791

https://github.com/python/typing_extensions/blob/fb37b2ee0ab56baf2dfd5df27aec0430435ba17f/src/typing_extensions.py#L804-L806

JelleZijlstra: I think eternal deprecation warnings might be the right answer, actually. (Or eternal until we ever make typing-extensions 5 in the distant uncertain future.)

AlexWaygood: So maybe we do want the change in https://github.com/python/typing_extensions/pull/150#discussion_r1174612309? ;)

JelleZijlstra: Yes, let's do that. I commented there because the change looked unrelated to this PR.

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@
   Patch by Alex Waygood.
 - Speedup `isinstance(3, typing_extensions.SupportsIndex)` by >10x on Python
   <3.12. Patch by Alex Waygood.
+- Add `__orig_bases__` to non-generic TypedDicts, call-based TypedDicts, and
+  call-based NamedTuples. Other TypedDicts and NamedTuples already had the attribute.
+  Patch by Adrian Garcia Badaracco.
```
AlexWaygood, line 54: Let's note the new `DeprecationWarning` for keyword arguments for call-based `TypedDict`s in the changelog as well:

```suggestion
  Patch by Adrian Garcia Badaracco.
- Constructing a call-based `TypedDict` using keyword arguments for the fields
  now causes a `DeprecationWarning` to be emitted. This matches the behaviour
  of `typing.TypedDict` on 3.11 and 3.12.
```

AlexWaygood:
I filed https://github.com/adriangb/typing_extensions/pull/2 as a PR to your branch to address my comment above here: https://github.com/python/typing_extensions/pull/150#pullrequestreview-1396937557

AlexWaygood (approved):
Looks great to me!

src/test_typing_extensions.py
```diff
@@ -2829,6 +2822,50 @@ def test_get_type_hints_typeddict(self):
...
+        self.assertEqual(OtherChild.__orig_bases__, (Parent,))
+        self.assertEqual(MixedChild.__orig_bases__, (Child, OtherChild, Parent,))
+        self.assertEqual(GenericParent.__orig_bases__, (TypedDict, Generic[T]))
+        self.assertEqual(GenericChild.__orig_bases__, (GenericParent[int],))
+        self.assertEqual(OtherGenericChild.__orig_bases__, (GenericParent[str],))
+        self.assertEqual(MixedGenericChild.__orig_bases__, (GenericChild, OtherGenericChild, GenericParent[float]))
+        self.assertEqual(MultipleGenericBases.__orig_bases__, (GenericParent[int], GenericParent[float]))
+        self.assertEqual(CallTypedDict.__orig_bases__, (TypedDict,))
+
+
```
JelleZijlstra, line 2868:
```suggestion
```

JelleZijlstra (approved):

PR #152, opened 2023-04-24
Backport tests of `Union` + `Literal` from CPython

sobolevn:
This PR is the backport of two my CPython's PR:
1. https://github.com/python/cpython/pull/103706 (merged)
2. https://github.com/python/cpython/pull/103747 (not merged yet)

Sorry, I often forget to do the backport﻿ when I write new tests.


sobolevn:
Hm, looks like tests for `|` do not make sense here, because `|` is a C-level implementation. Removed.

sobolevn:
It is merged now :)

PR #157, opened 2023-05-08
Backport `NewType` as it exists on py310+

### LINKED ISSUE #156: Backport Python 3.10 NewTypes
As of Python 3.10, the following types are picklable:

```Python
from pathlib import Path
from typing import NewType

File = NewType("File", Path)
Directory = NewType("Directory", Path)
```

Prior to that, this fails. Would it be possible to backport the `NewType` class to Python 3.7+?

AlexWaygood:
Fixes #156. The implementation and tests are basically just copied from CPython, with a few tweaks

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -440,7 +440,6 @@ else:
 Counter = typing.Counter
 ChainMap = typing.ChainMap
 AsyncGenerator = typing.AsyncGenerator
-NewType = typing.NewType
 Text = typing.Text
 TYPE_CHECKING = typing.TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


src/typing_extensions.py
```diff
@@ -2546,3 +2545,68 @@ class Baz(list[str]): ...
...
+        def __mro_entries__(self, bases):
+            # We defined __mro_entries__ to get a better error message
+            # if a user attempts to subclass a NewType instance. bpo-46170
+            supercls_name = self.__name__
+
+            class Dummy:
+                def __init_subclass__(cls):
+                    subcls_name = cls.__name__
+                    raise TypeError(
+                        f"Cannot subclass an instance of NewType. "
```
JelleZijlstra, line 2591: I feel like this would be clearer as
```suggestion
                        f"Cannot subclass a NewType. "
```
as it may not be obvious to users that "an instance of NewType" refers to a type. I suppose this is the same error as CPython though, so up to you whether it's worth changing.

src/typing_extensions.py
```diff
@@ -2546,3 +2545,68 @@ class Baz(list[str]): ...
...
+
+        def __repr__(self):
+            return f'{self.__module__}.{self.__qualname__}'
+
+        def __reduce__(self):
+            return self.__qualname__
+
+        if sys.version_info >= (3, 10):
+            # PEP 604 methods
+            # It doesn't make sense to have these methods on Python <3.10
```
JelleZijlstra, line 2606: We could allow this in theory, but it would probably be more confusing than useful to users to allow `|` on only a small subset of types.

JelleZijlstra: On the other hand:

```
Python 3.7.16 (default, Dec 26 2022, 20:28:00) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> from typing_extensions import Self
>>> Self | int
typing.Union[typing_extensions.Self, int]
```

src/typing_extensions.py
AlexWaygood: The nice thing about the "an instance of NewType" wording is that it explains _why_ you can't subclass the thing you're trying to subclass (it's an instance, not a class). But maybe the current phrasing is too terse to be useful to people who aren't familiar with how NewType works.

src/typing_extensions.py
AlexWaygood: Seems like we don't backport PEP 604 for `TypeVar` or `ParamSpec`:

```pycon
Python 3.7.16 (default, Jan 17 2023, 16:06:28) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> import typing_extensions
>>> typing_extensions.TypeVar("T") | int
TypeError: unsupported operand type(s) for |: 'TypeVar' and 'type'
>>> typing_extensions.ParamSpec("P") | int
TypeError: unsupported operand type(s) for |: 'ParamSpec' and 'type'
```

JelleZijlstra (approved):

PR #160, opened 2023-05-14
Add TypeAliasType

### LINKED ISSUE #159: Backport PEP 695's TypeAliasType
Cross-referencing here for visibility.

https://discuss.python.org/t/please-consider-delaying-or-even-rejecting-pep-695/26408/62?u=adriangb
https://mail.python.org/archives/list/typing-sig@python.org/thread/T5MRLQL7H4Y4HWFA7CBC34MHJCKQEI7L/

cc @JelleZijlstra, I didn't want to bug everyone in the discuss thread 😅

JelleZijlstra:
Fixes #159

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ This module currently contains the following:
 - In the standard library since Python 3.12
 
   - `override` (equivalent to `typing.override`; see [PEP 698](https://peps.python.org/pep-0698/))
-  - `TypeAliasType` (equivalent to `typing.TypeAliasType`; see [PEP 695](https://peps.python.org/pep-0698/))
+  - `TypeAliasType` (equivalent to `typing.TypeAliasType`; see [PEP 695](https://peps.python.org/pep-0695/))
   - `Buffer` (equivalent to `collections.abc.Buffer`; see [PEP 688](https://peps.python.org/pep-0688/))
   - `get_original_bases` (equivalent to
     [`types.get_original_bases`](https://docs.python.org/3.12/library/types.html#types.get_original_bases)
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ This module currently contains the following:
 - In the standard library since Python 3.12
 
   - `override` (equivalent to `typing.override`; see [PEP 698](https://peps.python.org/pep-0698/))
+  - `TypeAliasType` (equivalent to `typing.TypeAliasType`; see [PEP 695](https://peps.python.org/pep-0698/))
   - `Buffer` (equivalent to `collections.abc.Buffer`; see [PEP 688](https://peps.python.org/pep-0688/))
   - `get_original_bases` (equivalent to
     [`types.get_original_bases`](https://docs.python.org/3.12/library/types.html#types.get_original_bases)
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -4577,16 +4577,20 @@ class TypeAliasTypeTests(BaseTestCase):
 
     def test_immutable(self):
         Simple = TypeAliasType("Simple", int)
-        with self.assertRaisesRegex(TypeError, "cannot be modified"):
+        with self.assertRaisesRegex(AttributeError, "Can't set attribute"):
             Simple.__name__ = "NewName"
-        with self.assertRaisesRegex(TypeError, "cannot be modified"):
+        with self.assertRaisesRegex(AttributeError, "Can't set attribute"):
             Simple.__value__ = str
-        with self.assertRaisesRegex(TypeError, "cannot be modified"):
+        with self.assertRaisesRegex(AttributeError, "Can't set attribute"):
             Simple.__type_params__ = (T,)
-        with self.assertRaisesRegex(TypeError, "cannot be modified"):
+        with self.assertRaisesRegex(AttributeError, "Can't set attribute"):
             Simple.__parameters__ = (T,)
-        with self.assertRaisesRegex(TypeError, "cannot be modified"):
+        with self.assertRaisesRegex(AttributeError, "Can't set attribute"):
             Simple.some_attribute = "not allowed"
+        with self.assertRaisesRegex(AttributeError, "Can't delete attribute"):
+            del Simple.__name__
+        with self.assertRaisesRegex(AttributeError, "Can't delete attribute"):
+            del Simple.nonexistent_attribute
 
     def test_or(self):
         Alias = TypeAliasType("Alias", int)
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -21,7 +21,7 @@ from unittest.mock import patch
 import typing
 from typing import TypeVar, Optional, Union, AnyStr
 from typing import T, KT, VT  # Not in __all__.
-from typing import Tuple, List, Dict, Iterable, Iterator, Callable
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -42,6 +42,7 @@ This module currently contains the following:
 - In the standard library since Python 3.12
 
   - `override` (equivalent to `typing.override`; see [PEP 698](https://peps.python.org/pep-0698/))
+  - `TypeAliasType` (equivalent to `typing.TypeAliasType`; see [PEP 695](https://peps.python.org/pep-0698/))
```
srittau, line 45:
```suggestion
  - `TypeAliasType` (equivalent to `typing.TypeAliasType`; see [PEP 695](https://peps.python.org/pep-0695/))
```

srittau (changes requested):
One typo, LG otherwise.

src/typing_extensions.py
```diff
@@ -2610,3 +2611,81 @@ def __or__(self, other):
...
+          of the same name. (For example, 'X = TypeAliasType("Y", int)' is invalid,
+          as is 'X, Y = TypeAliasType("X", int), TypeAliasType("Y", int)').
+
+        """
+        def __init__(self, name: str, value, *, type_params=()):
+            if not isinstance(name, str):
+                raise TypeError("TypeAliasType name must be a string")
+            self.__name__ = name
+            self.__value__ = value
+            self.__type_params__ = type_params
```
AlexWaygood, line 2651: The PEP specifies that these attributes should be read-only, so shouldn't they be properties?

JelleZijlstra: I suppose we could do that, but it feels unnecessarily complicated.

AlexWaygood: I don't feel strongly, but it doesn't feel _that_ complicated to me, and I think it would be nice to match CPython on 3.12+ here 🤷‍♂️

src/test_typing_extensions.py
```diff
@@ -4494,6 +4494,11 @@ def test_pickle(self):
                 unpickled = pickle.loads(pickled)
                 self.assertIs(unpickled, Alias)
 
+    def test_no_instance_subclassing(self):
+        with self.assertRaises(TypeError):
+            class MyAlias(TypeAliasType):
+                pass
```
AlexWaygood, line 4500: (Might be worth adding a test like this to https://github.com/python/cpython/pull/103764 as well; I couldn't spot one)

JelleZijlstra: Good point, https://github.com/python/cpython/pull/103764/commits/d4e72a5669a6a11b17c5591fe1494f29e25f963d

src/typing_extensions.py
```diff
@@ -2610,3 +2611,86 @@ def __or__(self, other):
...
+            parameters = []
+            for type_param in type_params:
+                if isinstance(type_param, TypeVarTuple):
+                    parameters.extend(type_param)
+                else:
+                    parameters.append(type_param)
+            self.__parameters__ = tuple(parameters)
+            def_mod = _caller()
+            if def_mod != 'typing_extensions':
+                self.__module__ = def_mod
```
AlexWaygood, line 2662: In https://github.com/python/typing_extensions/pull/160, it looks like `__module__` is always `typing`, no matter what module the type alias is defined in:

```pycon
>>> type T = int | str
>>> T.__module__
'typing'
```

I like _this_ behaviour more, but it's more complicated to implement, and doesn't match the CPython PR currently :)

srittau (approved):

src/typing_extensions.py
AlexWaygood: (Resolved by https://github.com/python/cpython/pull/104550, for future reference)

src/typing_extensions.py
```diff
@@ -2610,3 +2611,86 @@ def __or__(self, other):
...
+
+        Static type checkers should only support type aliases declared using
+        TypeAliasType that follow these rules:
+
+        - The first argument (the name) must be a string literal.
+        - The TypeAliasType instance must be immediately assigned to a variable
+          of the same name. (For example, 'X = TypeAliasType("Y", int)' is invalid,
+          as is 'X, Y = TypeAliasType("X", int), TypeAliasType("Y", int)').
+
+        """
```
AlexWaygood, line 2645: Could disallow arbitrary setting of attributes (matching the behaviour on 3.12) by adding `__slots__`:

```suggestion
        """
        __slots__ = (
            "__name__", "__value__", "__type_params__", "__parameters__", "__module__"
        )

```

JelleZijlstra: Unfortunately this doesn't work due to the way `__module__` is set up. I'll override `__setattr__` instead.

src/typing_extensions.py
```diff
@@ -2646,3 +2647,93 @@ def __or__(self, other):
...
+            self.__parameters__ = tuple(parameters)
+            def_mod = _caller()
+            if def_mod != 'typing_extensions':
+                self.__module__ = def_mod
+            # Setting this attribute closes the TypeAliasType from further modification
+            self.__name__ = name
+
+        def __setattr__(self, __name: str, __value: object) -> None:
+            if hasattr(self, "__name__"):
+                raise TypeError("TypeAliasType cannot be modified")
```
AlexWaygood, line 2704: Shouldn't it be an `AttributeError`? I feel like having `__setattr__` or `__getattr__` raising anything that's not `AttributeError` or a subclass is kind-of evil. CPython `main` also raises `AttributeError`s:

```pycon
>>> type T = int | str
>>> T.__name__ = 'foo'
AttributeError: readonly attribute
>>> T.bar = 'aba'
AttributeError: 'typing.TypeAliasType' object has no attribute 'bar'
```

```suggestion
                raise AttributeError(f"Can't set attribute {__name!r} on an instance of TypeAliasType")
```

AlexWaygood: We could possibly also add a `__delattr__` method as well, so these attributes can't be deleted

JelleZijlstra: Good points, pushed both

AlexWaygood (approved):
Looks great!

src/typing_extensions.py
```diff
@@ -2646,3 +2647,96 @@ def __or__(self, other):
...
+            # Setting this attribute closes the TypeAliasType from further modification
+            self.__name__ = name
+
+        def __setattr__(self, __name: str, __value: object) -> None:
+            if hasattr(self, "__name__"):
+                raise AttributeError(f"Can't set attribute {__name!r} on an instance of TypeAliasType")
+            super().__setattr__(__name, __value)
+
+        def __delattr__(self, __name: str) -> None:
+            raise AttributeError(f"Can't delete attribute {__name!r} on an instance of TypeAliasType")
```
AlexWaygood, line 2708: To fix flake8:

```suggestion
                raise AttributeError(
                    f"Can't set attribute {__name!r} on an instance of TypeAliasType"
                )
            super().__setattr__(__name, __value)

        def __delattr__(self, __name: str) -> None:
            raise AttributeError(
                f"Can't delete attribute {__name!r} on an instance of TypeAliasType"
            )
```

JelleZijlstra: Thanks!

AlexWaygood (approved):

PR #161, opened 2023-05-19
Backport some recent `Protocol` fixes from 3.12

AlexWaygood:
Backports these PRs:
- https://github.com/python/cpython/pull/104556
- https://github.com/python/cpython/pull/104559
- https://github.com/python/cpython/pull/104622
- https://github.com/python/cpython/pull/104649

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1907,6 +1907,63 @@ class ProtocolTests(BaseTestCase):
         with self.assertRaises(TypeError):
             issubclass(D, PNonCall)
 
+    def test_no_weird_caching_with_issubclass_after_isinstance(self):
+        @runtime_checkable
+        class Spam(Protocol):
+            x: int
+
+        class Eggs:
+            def __init__(self) -> None:
+                self.x = 42
+
+        self.assertIsInstance(Eggs(), Spam)
+
+        # gh-104555: If we didn't override ABCMeta.__subclasscheck__ in _ProtocolMeta,
+        # TypeError wouldn't be raised here,
+        # as the cached result of the isinstance() check immediately above
+        # would mean the issubclass() call would short-circuit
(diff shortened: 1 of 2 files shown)


src/test_typing_extensions.py
```diff
@@ -2325,6 +2397,48 @@ def bar(self, x: str) -> str:
...
+
+                self.assertIsInstance(Eggs(42), Spam)
+
+                # gh-104555: If we didn't override ABCMeta.__subclasscheck__ in _ProtocolMeta,
+                # TypeError wouldn't be raised here,
+                # as the cached result of the isinstance() check immediately above
+                # would mean the issubclass() call would short-circuit
+                # before we got to the "raise TypeError" line
+                with self.assertRaises(TypeError):
+                    issubclass(Eggs, Spam)
```
AlexWaygood, line 2438: I _think_ these pass, but had no way of checking, since `typing_extensions` currently fails to import on 3.12:

```
  File "C:\Users\alexw\coding\typing_extensions\src\test_typing_extensions.py", line 29, in <module>
    import typing_extensions
  File "C:\Users\alexw\coding\typing_extensions\src\typing_extensions.py", line 1361, in <module>
    class TypeVar(typing.TypeVar, _DefaultMixin, _root=True):
TypeError: type 'typing.TypeVar' is not an acceptable base type
```

AlexWaygood:
Oh, the PyPy version of `ABCMeta` results in a bunch of extra attributes being present in a class's `__dict__`? TIL.

AlexWaygood:
> Oh, the PyPy version of `ABCMeta` results in a bunch of extra attributes being present in a class's `__dict__`? TIL.

Ugh, that means we'll probably have to mark the test in CPython as `@cpython_only`.

JelleZijlstra (approved):

PR #162, opened 2023-05-19
Fix tests on Python 3.12

JelleZijlstra:
On current Python main, TypeVar cannot be subclassed. However, we still need to enhance TypeVar to add the `default=` argument. This PR introduces an alternative approach to achieve that: we create a dummy class that wraps `typing.TypeVar` and when instantiated, returns an instance of `typing.TypeVar`, possibly with some additional attributes. We also set `__instancecheck__` so that `isinstance(x, typing_extensions.TypeVar)` continues to work.

Also fix a few other minor test failures on current CPython main.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -4476,10 +4476,13 @@ class BufferTests(BaseTestCase):
                 return memoryview(b'')
 
         # On 3.12, collections.abc.Buffer does a structural compatibility check
-        if not TYPING_3_12_0:
+        if TYPING_3_12_0:
+            self.assertIsInstance(MyRegisteredBuffer(), Buffer)
+            self.assertIsSubclass(MyRegisteredBuffer, Buffer)
+        else:
             self.assertNotIsInstance(MyRegisteredBuffer(), Buffer)
             self.assertNotIsSubclass(MyRegisteredBuffer, Buffer)
-            Buffer.register(MyRegisteredBuffer)
+        Buffer.register(MyRegisteredBuffer)
         self.assertIsInstance(MyRegisteredBuffer(), Buffer)
         self.assertIsSubclass(MyRegisteredBuffer, Buffer)
 
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -47,6 +47,9 @@ TYPING_3_10_0 = sys.version_info[:3] >= (3, 10, 0)
 # 3.11 makes runtime type checks (_type_check) more lenient.
 TYPING_3_11_0 = sys.version_info[:3] >= (3, 11, 0)
 
+# 3.12 changes the representation of Unpack[] (PEP 692)
+TYPING_3_12_0 = sys.version_info[:3] >= (3, 12, 0)
+
 # https://github.com/python/cpython/pull/27017 was backported into some 3.9 and 3.10
 # versions, but not all
 HAS_FORWARD_MODULE = "module" in inspect.signature(typing._type_check).parameters
@@ -3531,7 +3534,9 @@ class UnpackTests(BaseTestCase):
 
     def test_repr(self):
         Ts = TypeVarTuple('Ts')
-        if TYPING_3_11_0:
+        if TYPING_3_12_0:
+            self.assertEqual(repr(Unpack[Ts]), 'typing.Unpack[Ts]')
+        elif TYPING_3_11_0:
             self.assertEqual(repr(Unpack[Ts]), '*Ts')
         else:
             self.assertEqual(repr(Unpack[Ts]), 'typing_extensions.Unpack[Ts]')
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -3531,7 +3534,9 @@ def test_basic_plain(self):
 
     def test_repr(self):
         Ts = TypeVarTuple('Ts')
-        if TYPING_3_11_0:
+        if TYPING_3_12_0:
+            self.assertEqual(repr(Unpack[Ts]), 'typing.Unpack[Ts]')
+        elif TYPING_3_11_0:
```
AlexWaygood, line 3539: Hmm, I feel like the test failure is pointing to a real issue here. We should backport https://github.com/python/cpython/pull/104048 -- the current `typing_extensions.Unpack` repr doesn't make any sense if you use `Unpack` with a `TypedDict` as per PEP 692

src/test_typing_extensions.py
```diff
@@ -4229,8 +4234,8 @@ def test_signature_on_37(self):
     @skipUnless(TYPING_3_9_0, "NamedTuple was a class on 3.8 and lower")
     def test_same_as_typing_NamedTuple_39_plus(self):
         self.assertEqual(
-            set(dir(NamedTuple)),
-            set(dir(typing.NamedTuple)) | {"__text_signature__"}
+            set(dir(NamedTuple)) - {"__text_signature__"},
+            set(dir(typing.NamedTuple))
```
AlexWaygood, line 4238: Interesting, what was the failure here?

src/test_typing_extensions.py
```diff
@@ -4341,9 +4363,11 @@ class MyRegisteredBuffer:
...
                 return memoryview(b'')
 
-        self.assertNotIsInstance(MyRegisteredBuffer(), Buffer)
-        self.assertNotIsSubclass(MyRegisteredBuffer, Buffer)
-        Buffer.register(MyRegisteredBuffer)
+        # On 3.12, collections.abc.Buffer does a structural compatibility check
+        if not TYPING_3_12_0:
+            self.assertNotIsInstance(MyRegisteredBuffer(), Buffer)
+            self.assertNotIsSubclass(MyRegisteredBuffer, Buffer)
+            Buffer.register(MyRegisteredBuffer)
```
AlexWaygood, line 4370: How about:

```suggestion
        # On 3.12, collections.abc.Buffer does a structural compatibility check
        if TYPING_3_12_0:
            self.assertIsInstance(MyRegisteredBuffer(), Buffer)
            self.assertIsSubclass(MyRegisteredBuffer, Buffer)
        else:
            self.assertNotIsInstance(MyRegisteredBuffer(), Buffer)
            self.assertNotIsSubclass(MyRegisteredBuffer, Buffer)
        Buffer.register(MyRegisteredBuffer)
```

src/typing_extensions.py
```diff
@@ -1331,39 +1332,53 @@ def __repr__(self):
...
+        return typevar
+
+    def __instancecheck__(self, __instance: Any) -> bool:
+        return isinstance(__instance, typing.TypeVar)
+
+
+class TypeVar(metaclass=_TypeVarMeta):
+    """Type variable."""
+
+    __module__ = 'typing'
```
AlexWaygood, line 1381: heh

AlexWaygood (commented):
The code looks good! A few comments about the tests. The only big concern I have is about `typing.Unpack` -- I think we should backport https://github.com/python/cpython/pull/104048 rather than just fiddling with the test to make it pass on 3.12.

Feels like we could probably share more code between `_TypeVarMeta`, `_ParamSpecMeta` and `_TypeVarTupleMeta`, but that can definitely be tackled in a followup PR; no need to worry about it now.

src/test_typing_extensions.py
JelleZijlstra: I didn't look too hard honestly, something about `__text_signature__` unexpectedly not being in one of the sets. I assumed it was about whether or not Argument Clinic had been applied to the tuple constructor or something.

AlexWaygood: the change seems fine, I'm just surprised the test broke!

src/test_typing_extensions.py
AlexWaygood: See #163

src/test_typing_extensions.py
AlexWaygood, line 1: Apparently we now need to add `import textwrap` at the top of this file, because I added a use of `textwrap.dedent` in this file in #161 but forgot to add an import... Sorry about that 🙃

JelleZijlstra:
> Feels like we could probably share more code between `_TypeVarMeta`, `_ParamSpecMeta` and `_TypeVarTupleMeta`, but that can definitely be tackled in a followup PR; no need to worry about it now.

I'm not sure that's feasible. The code is similar for all three, but not quite the same: they all take different arguments and map to different `typing` classes.

AlexWaygood (approved):
The changes look good, and I confirmed that the tests all pass on CPython `main`.

It looks like the tests _do_ currently crash on 3.12 alpha7, due to this not being valid syntax until very recently:

https://github.com/python/typing_extensions/blob/40dbc09dbb01982f3ef58db05f38ba9d61cd19e9/src/test_typing_extensions.py#L2404-L2405

(See https://github.com/python/typing_extensions/actions/runs/5029362366/jobs/9020946297?pr=162)

So I suppose we could possibly make this `sys.version_info` check more fine-grained:

https://github.com/python/typing_extensions/blob/40dbc09dbb01982f3ef58db05f38ba9d61cd19e9/src/test_typing_extensions.py#L2399

But also, the beta is about to be released, so maybe it doesn't matter too much?

JelleZijlstra:
Thanks! I'll push something to make it check for `hasattr(typing, 'TypeAliasType')` instead. I remember @carljm saying at the language summit that it's impossible to test any third-party libraries on early alphas; maybe we can make typing-extensions a partial exception :)

PR #163, opened 2023-05-19
Backport changes to the repr of `typing.Unpack` that were made in Python 3.12

AlexWaygood:
Also, sync the docstring with the CPython docstring, which is much more extensive.

Backport of https://github.com/python/cpython/pull/104048

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -3531,10 +3531,7 @@ class UnpackTests(BaseTestCase):
 
     def test_repr(self):
         Ts = TypeVarTuple('Ts')
-        if TYPING_3_11_0:
-            self.assertEqual(repr(Unpack[Ts]), '*Ts')
-        else:
-            self.assertEqual(repr(Unpack[Ts]), 'typing_extensions.Unpack[Ts]')
+        self.assertEqual(repr(Unpack[Ts]), 'typing_extensions.Unpack[Ts]')
 
     def test_cannot_subclass_vars(self):
         with self.assertRaises(TypeError):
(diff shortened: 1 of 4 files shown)


src/test_typing_extensions.py
```diff
@@ -3531,10 +3531,7 @@ def test_basic_plain(self):
 
     def test_repr(self):
         Ts = TypeVarTuple('Ts')
-        if TYPING_3_11_0:
-            self.assertEqual(repr(Unpack[Ts]), '*Ts')
-        else:
-            self.assertEqual(repr(Unpack[Ts]), 'typing_extensions.Unpack[Ts]')
+        self.assertEqual(repr(Unpack[Ts]), 'typing_extensions.Unpack[Ts]')
```
JelleZijlstra, line 3534: This will still fail on 3.12 because this will say typing, not typing_extensions

AlexWaygood:
```suggestion
        self.assertEqual(repr(Unpack[Ts]), f'{Unpack.__module__}.Unpack[Ts]')
```

src/typing_extensions.py
```diff
@@ -1960,14 +1960,60 @@ class Movie(TypedDict):
...
+    year: int
+
+  # This function expects two keyword arguments - *name* of type `str` and
+  # *year* of type `int`.
+  def foo(**kwargs: Unpack[Movie]): ...
+
+Note that there is only some runtime checking of this operator. Not
+everything the runtime allows may be accepted by static type checkers.
+
+For more information, see PEP 646.
```
JelleZijlstra, line 2001: and 692. Not sure that's worth a CPython PR, but let's change it here.
```suggestion
For more information, see PEP 646 and PEP 692.
```

PR #165, opened 2023-05-21
Bring over TypeVarTests from CPython

JelleZijlstra:
I started out trying to backport python/cpython#104571, but
realized that it makes sense to backport CPython's whole TypeVarTests
class since we now have our own implementation of TypeVar.

I dropped test_var_substitution and test_bad_var_substitution
since they rely on the internal __typing_subst__ method, and the
type substitution logic is generally very hard to get precisely
the same across versions.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -20,7 +20,7 @@ from pathlib import Path
 from unittest import TestCase, main, skipUnless, skipIf
 from unittest.mock import patch
 import typing
-from typing import TypeVar, Optional, Union, AnyStr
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
CI is failing because we didn't bring over the fix for python/cpython#84123. It doesn't seem worth the effort to fix that for early versions of 3.7 and 3.8, so I'm just going to skip that test before 3.9.

AlexWaygood (approved):
Haven't done a line-by-line review, but I trust in your copy-and-paste abilities :)

Issue #110, opened 2023-01-23
Backport allowing passing an ellipsis as the last argument of Concatenate to 3.10


FasterSpeeding:
https://github.com/python/cpython/pull/30969 allowed and documented this behaviour for 3.11.

This seemingly can't be supported for <=3.9 according to https://github.com/python/typing_extensions/issues/48 but changes made by https://github.com/python/cpython/pull/30959 to the stdlib typing module lead to this behaviour erroring on later 3.10 releases when using `typing_extensions.Concatenate`.

JelleZijlstra:
This is now documented as a limitation at https://typing-extensions.readthedocs.io/en/latest/#Concatenate. If there is a reasonable way to fix it, I'd accept a PR.

Issue #48, opened 2022-06-09
`Concatenate` implementation generates runtime exception if parameterized by `...`


erictraut:
The following code should work without a runtime exception.

```python
from typing import Any, Callable
from typing_extensions import ParamSpec, Concatenate


P = ParamSpec("P")
MyAlias = Callable[Concatenate[int, P], Any]

x: MyAlias[...]
```

It runs fine on Python 3.10 if using `typing.Concatenate` but fails on Python 3.9 if using `typing_extensions.Concatenate`. The runtime exception is:
```
TypeError: Parameters to generic types must be types. Got Ellipsis.
```

Fidget-Spinner:
This seems like a bug in Python's typing module itself.
```python
# In 3.9
typing.Callable[T, int][...]

TypeError: Parameters to generic types must be types. Got Ellipsis.
```
The problem is that `Callable` uses the default `_GenericAlias`'s substitution behavior pre 3.10. Which means its checks are too restrictive. We had to intentionally loosen the checks in 3.10 to get around this.

I don't think we can fix this from `typing_extensions` unless we monkey patch `typing._GenericAlias.__getitem__`. I don't feel too good doing that though.

JelleZijlstra:
This is now documented as a limitation at https://typing-extensions.readthedocs.io/en/latest/#Concatenate. If there is a reasonable way to fix it, I'd accept a PR.

Daraan:
I am currently looking at this again and the statement that "It runs fine on Python 3.10 if using typing.Concatenate" confused be a bit.
In hindsight this should only hold for 3.10.0-3.10.2 afterwards it also does not run on the other 3.10 `typing` versions either.

---

This should be fixable for the rest of the 3.10 versions if `typing.Concatenate` is monkey patched to use the `typing_extensions.Concatenate` variant combined with `_concatenate_getitem` from #442 that accepts an `...` argument.

Whats then left to do is to use `typing_extensions._ConcatenateGenericAlias` for 3.10.3+ < 3.11 instead of `typing._ConcatenateGenericAlias`.
This will fix the usability aspect

This then will cause a regression in the `__args__` attribute as `typing_extensions._ConcatenateGenericAlias` inherits from list which will be unpacked instead of staying an instance in the typing module (iirc there is an iterable check).
From the comment this should only be necessary for Python 3.9.(<2?).

In short the regression will be these two tests:
```python
                # skip for 3.10.3+ < 3.11
[...]

PR #166, opened 2023-05-21
Add documentation page for typing-extensions

JelleZijlstra:
Currently, typing-extensions is documented only through the README and CHANGELOG. This is hard to navigate, so I feel it's time for a documentation page dedicated to typing-extensions.

We can host it at typing-extensions.readthedocs.io.


doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+   See :py:class:`typing.NamedTuple`.
+
+   ``typing_extensions`` backports several changes
+   to ``NamedTuple`` on Python 3.11 and lower: in 3.11,
+   support for generic ``NamedTuple`` was added, and
+   in 3.12, the ``__orig_bases__`` attribute was added.
+
+   .. versionadded:: 4.3.0
+
+      Added to provide support for generic ``NamedTuple``.
```
AlexWaygood, line 108:
```suggestion
      Added to provide support for generic ``NamedTuple``s.
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+   See :py:data:`typing.Never`. In ``typing`` since 3.11.
+
+   .. versionadded:: 4.1.0
+
+.. class:: NewType(name, tp)
+
+   See :py:class:`typing.NewType`. In ``typing`` since 3.5.2.
+
+   ``NewType`` was made picklable in 3.10 and an error message was
```
AlexWaygood, line 124:
```suggestion
   Instances of ``NewType`` were made picklable in 3.10 and an error message was
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+   See :py:class:`typing.Protocol`. In ``typing`` since 3.8.
+
+   Python 3.12 improves the performance of runtime-checkable protocols;
+   ``typing_extensions`` backports this improvement.
+
+   .. versionchanged:: 4.6.0
+
+      Backported the ability to define ``__init__`` methods on Protocol classes.
+
+   .. versionchanged:: 4.6.0
```
AlexWaygood, line 174: Could possibly also mention https://github.com/python/typing_extensions/pull/140 and #132 in versionchanged notes for `Protocol` — they bring the implementation in line with Python 3.12, and they're strictly-speaking bugfixes, but they're also minor compatibility breaks with previous versions

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+   honor the ``total`` keyword with old-style ``TypedDict()`` in Python
+   3.9.0 and 3.9.1. :py:class:`typing.TypedDict` also does not support multiple inheritance
+   with :py:class:`typing.Generic` on Python <3.11, and :py:class:`typing.TypedDict` classes do not
+   consistently have the ``__orig_bases__`` attribute on Python <3.12. The
+   ``typing_extensions`` backport provides all of these features and bugfixes on
+   all Python versions.
+
+   .. versionadded:: 4.3.0
+
+      Added to provide support for generic ``TypedDict``.
```
AlexWaygood, line 225: I don't think that's true — it existed in typing_extensions before we modified the implementation to provide support for generic TypedDicts

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+      Added in order to support the new ``default=`` and
+      ``infer_variance=`` arguments.
+
+   .. versionchanged:: 4.6.0
+
+      The implementation was changed for compatibility with Python 3.12.
+
+.. class:: TypeVarTuple(name, *, default=...)
+
+   See :py:class:`typing.TypeVarTuple`. In ``typing`` since 3.10.
```
AlexWaygood, line 257:
```suggestion
   See :py:class:`typing.TypeVarTuple`. In ``typing`` since 3.11.
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+      Added support for the ``default=`` argument.
+
+   .. versionchanged:: 4.6.0
+
+      The implementation was changed for compatibility with Python 3.12.
+
+.. data:: Unpack
+
+   See :py:data:`typing.Unpack`. In ``typing`` since 3.10.
```
AlexWaygood, line 274:
```suggestion
   See :py:data:`typing.Unpack`. In ``typing`` since 3.11.
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+Protocols
+~~~~~~~~~
+
+.. class:: SupportsAbs
+
+   See :py:class:`typing.SupportsAbs`.
+
+   ``typing_extensions`` backports a more performant version of this
+   protocol in Python 3.11 and lower.
```
AlexWaygood, line 354: I think I'd go for "on" here (and in the other Protocol descriptions below)?

```suggestion
   protocol on Python 3.11 and lower.
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+      The ``field_descriptors`` parameter was renamed to ``field_specifiers``.
+      For compatibility, the decorator now accepts arbitrary keyword arguments.
+
+   .. versionchanged:: 4.5.0
+
+      The ``frozen_default`` parameter was added.
+
+.. decorator:: deprecated(msg, *, category=DeprecationWarning, stacklevel=1)
+
+   See :pep:`702`.
```
AlexWaygood, line 439:
```suggestion
   See :pep:`702`. Experimental; not yet part of the standard library.
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+   .. versionadded:: 4.5.0
+
+.. decorator:: final
+
+   See :py:func:`typing.final`. In ``typing`` since 3.8.
+
+   Since Python 3.11, this decorator supports runtime introspection
+   through the ``__final__`` attribute; ``typing_extensions.final``
```
AlexWaygood, line 448:
```suggestion
   by setting the ``__final__`` attribute wherever possible; ``typing_extensions.final``
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+   backports this feature.
+
+   .. versionchanged:: 4.1.0
+
+      The ``__final__`` attribute is now set on decorated objects.
```
AlexWaygood, line 453: Maybe something like:

> The decorator now attempts to set the `__final__` attribute on decorated objects. AttributeErrors and TypeErrors are caught when trying to set this attribute.

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+.. decorator:: override
+
+   See :py:func:`typing.override`. In ``typing`` since 3.12.
+
+   .. versionadded:: 4.4.0
+
+   .. versionchanged:: 4.5.0
+
+      The decorator now sets the ``__override__`` attribute on the decorated
```
AlexWaygood, line 475: See my comment above about `@final`

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+.. function:: get_original_bases(cls)
+
+   See :py:func:`types.get_original_bases`. Added to the standard library
+   in Python 3.12.
+
+   This function should always produce correct results when called on classes
+   constructed using features from ``typing_extensions``. However, it may
+   produce incorrect results when called on some :py:class:`NamedTuple` or
+   :py:class:`TypedDict` on Python <=3.11.
```
AlexWaygood, line 531:
```suggestion
   :py:class:`TypedDict` classes on Python <=3.11.
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+.. function:: get_type_hints(obj, globalns=None, localns=None, include_extras=False)
+
+   See :py:func:`typing.get_type_hints`.
+
+   In Python 3.11, this function was changed to support the new
+   :py:data:`typing.Required`. ``typing_extensions`` backports these
+   fixes.
+
+   .. versionchanged:: 4.1.0
```
AlexWaygood, line 552: Presumably we must also have backported the `include_extras` parameter at some point after PEP 593 was implemented in Python 3.9

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+
+      Interaction with :data:`Required` and :data:`NotRequired`.
+
+.. function:: is_typeddict(tp)
+
+   See :py:func:`typing.is_typeddict`. In ``typing`` since 3.10.
+
+   On versions where :class:`TypedDict` is not the same as
+   :py:class:`typing.TypedDict`, this function recognizes
+   ``TypedDict`` classes created through either.
```
AlexWaygood, line 562:
```suggestion
   ``TypedDict`` classes created through either mechanism.
```

AlexWaygood (commented):
This is great, and very much needed — thank you!

Some comments, mostly pretty minor:

AlexWaygood (commented):
This is great, and very much needed — thank you!

Some comments, mostly pretty minor:

doc/index.rst
JelleZijlstra:
```suggestion
      Added to provide support for generic ``NamedTuple``\ s.
```

doc/index.rst
```diff
@@ -0,0 +1,599 @@
...
+   Python 3.12 improves the performance of runtime-checkable protocols;
+   ``typing_extensions`` backports this improvement.
+
+   .. versionchanged:: 4.6.0
+
+      Backported the ability to define ``__init__`` methods on Protocol classes.
+
+   .. versionchanged:: 4.6.0
+
+      Backported changes to runtime-checkable protocols from Python 3.12.
```
JelleZijlstra, line 176:
```suggestion
      Backported changes to runtime-checkable protocols from Python 3.12,
      including :py:gh:`103034` and :py:gh:`26067`.
```

doc/index.rst
JelleZijlstra:
```suggestion
      Added support for generic ``TypedDict``\ s.
```
Oops yes, confused myself with NamedTuple

doc/index.rst
JelleZijlstra:
```suggestion
      The decorator now attempts to set the ``__final__`` attribute on decorated objects. 
```
Left out the second sentence for brevity; details of behavior can be in the CPython docs.

doc/index.rst
JelleZijlstra:
```suggestion
      The decorator now attempts to set the ``__override__`` attribute on the decorated
```

doc/index.rst
JelleZijlstra: I think that was before we started a changelog. I haven't gone through the git archeology to figure out version changes from that time.

doc/index.rst
AlexWaygood: It needs to be "versionchanged" as well as "versionadded", it existed in `typing_extensions` before 4.3.0. I don't know when it was added.

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+   :py:class:`TypedDict` does not store runtime information
+   about which (if any) keys are non-required in Python 3.8, and does not
+   honor the ``total`` keyword with old-style ``TypedDict()`` in Python
+   3.9.0 and 3.9.1. :py:class:`typing.TypedDict` also does not support multiple inheritance
+   with :py:class:`typing.Generic` on Python <3.11, and :py:class:`typing.TypedDict` classes do not
+   consistently have the ``__orig_bases__`` attribute on Python <3.12. The
+   ``typing_extensions`` backport provides all of these features and bugfixes on
+   all Python versions.
+
+   .. versionadded:: 4.3.0
```
AlexWaygood, line 224: Don't know when it was added, but it wasn't in 4.3.0 :)

```suggestion
   .. versionchanged:: 4.3.0
```

JelleZijlstra: Thanks, just pushed that

doc/index.rst
JelleZijlstra: Done, thanks for all the comments!

doc/index.rst
```diff
@@ -0,0 +1,600 @@
+
+Welcome to typing_extensions's documentation!
+=============================================
+
+``typing_extensions`` provides common building blocks for type annotations
+that are understood by static type checkers.
+
+The ``typing_extensions`` module serves two related purposes:
```
AlexWaygood, line 8: It would be nice to mention (and link to) the `typing` module in the first paragraph, I think. Maybe something like:

```suggestion
``typing_extensions`` complements the standard-library :py:mod:`typing` module,
providing runtime support for type hints as specified by PEP 484 and subsequent
PEPs. The module serves two related purposes:
```

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+
+- Enable use of new type system features on older Python versions. For example,
+  :py:data:`typing.TypeGuard` is new in Python 3.10, but ``typing_extensions`` allows
+  users on previous Python versions to use it too.
+- Enable experimentation with new type system PEPs before they are accepted and
+  added to the :py:mod:`typing` module.
```
AlexWaygood, line 14: "Type system PEPs" sounds slightly strange to me -- an alternative solution would be "type-system PEPs", but that also feels slightly awkward

```suggestion
- Enable experimentation with new PEPs relating to the type system before they
  are accepted and added to the :py:mod:`typing` module.
```

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+
+New features may be added to ``typing_extensions`` as soon as they are specified
+in a PEP that has been added to the `python/peps <https://github.com/python/peps>`_
+repository. If the PEP is accepted, the feature will then be added to :py:mod:`typing`
+for the next CPython release. No typing PEP that affected ``typing_extensions``
+has been rejected so far, so we haven't yet figured out how to deal with that possibility.
```
AlexWaygood, line 20:
```suggestion
repository. If the PEP is accepted, the feature will then be added to the
:py:mod:`typing` module for the next CPython release. No typing PEP that
affected ``typing_extensions`` has been rejected so far, so we haven't yet
figured out how to deal with that possibility.
```

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+
+   See :py:func:`typing.overload`.
+
+   Since Python 3.11, this decorator supports runtime introspection
+   through :func:`get_overloads`; ``typing_extensions.overload``
+   backports this feature.
+
+   .. versionchanged:: 4.2.0
+
+      Added to provide support for introspection through :func:`get_overloads` was added.
```
AlexWaygood, line 466:
```suggestion
      Introspection support via :func:`get_overloads` was added.
```

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+
+   .. versionadded:: 4.2.0
+
+.. function:: get_type_hints(obj, globalns=None, localns=None, include_extras=False)
+
+   See :py:func:`typing.get_type_hints`.
+
+   In Python 3.11, this function was changed to support the new
+   :py:data:`typing.Required`. ``typing_extensions`` backports these
+   fixes.
```
AlexWaygood, line 551:
```suggestion
   :py:data:`typing.Required` and :py:data:`typing.NotRequired`.
   ``typing_extensions`` backports these fixes.
```

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+   See :py:class:`typing.Text`. In ``typing`` since 3.5.2.
+
+.. data:: TYPE_CHECKING
+
+   See :py:data:`typing.TYPE_CHECKING`. In ``typing`` since 3.5.2.
+
+Limitations
+-----------
+
+There are a few types whose interface was modified between different
```
AlexWaygood, line 587: Or, if we keep "whose", then "interface" needs to be plural ("interfaces were" rather than "interface was")

```suggestion
There are a few types where the interface was modified between different
```

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+versions of typing. For example, :py:class:`typing.Sequence` was modified to
+subclass :py:class:`typing.Reversible` as of Python 3.5.3.
+
+These changes are *not* backported to prevent subtle compatibility
+issues when mixing the differing implementations of modified classes.
+
+Certain types have incorrect runtime behavior due to limitations of older
+versions of the typing module:
```
AlexWaygood, line 595: Make it slightly clearer that this is a distinct point to the previous paragraph:

```suggestion
As well as this, certain types have incorrect runtime behavior due to
limitations of older versions of the typing module:
```

AlexWaygood (commented):
A few more small points:

doc/index.rst
AlexWaygood: Yeah, I tried digging through the git archaeology, and quickly gave up :)

doc/index.rst
JelleZijlstra: It would be nice to mention (and link to) the `typing` module in the first paragraph, I think. Maybe something like:
Also add a link:
```suggestion
``typing_extensions`` complements the standard-library :py:mod:`typing` module,
providing runtime support for type hints as specified by :pep:`484` and subsequent
PEPs. The module serves two related purposes:
```

doc/index.rst
JelleZijlstra: How about:

```suggestion
- Enable experimentation with type system features proposed in new PEPs before they are accepted and
  added to the :py:mod:`typing` module.
```

doc/index.rst
JelleZijlstra: There was no `typing_extensions.overload` before this, so the note is here to explain why we added. I should actually change the line above to "versionadded", not "versionchanged".

doc/index.rst
```diff
@@ -0,0 +1,600 @@
...
+
+.. decorator:: overload
+
+   See :py:func:`typing.overload`.
+
+   Since Python 3.11, this decorator supports runtime introspection
+   through :func:`get_overloads`; ``typing_extensions.overload``
+   backports this feature.
+
+   .. versionchanged:: 4.2.0
```
JelleZijlstra, line 464:
```suggestion
   .. versionadded:: 4.2.0
```

doc/index.rst
AlexWaygood: Yes, that works well!

doc/index.rst
AlexWaygood: Are you sure? I can see it as far back as 3.10.1: https://github.com/python/typing_extensions/blob/0c70227e221866b74b5cf6b3e251bc67864781ef/src_py3/typing_extensions.py#L693-L719

doc/index.rst
JelleZijlstra: I actually think we should get rid of this catch-all section and document these limitations only under specific objects. I'll move a version of this text to the ParamSpec docs instead, and get rid of the sentence about Sequence/Reversible which I think is only relevant to 3.5-era changes.

doc/index.rst
JelleZijlstra: Oh you're right, I misremembered. I wonder why; according to https://docs.python.org/3.10/library/typing.html#typing.overload it's always been in `typing`.

doc/index.rst
AlexWaygood: Yes, that makes sense. Sounds like a plan.

doc/index.rst
AlexWaygood: The past is a foreign country: they do things differently there

doc/index.rst
JelleZijlstra: I couldn't figure out what's supposedly wrong with Concatenate, so I'm going to add a note to ParamSpec only.

```
Python 3.8.16 (default, Dec 26 2022, 20:36:56) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> from typing_extensions import Concatenate, get_args, get_origin, ParamSpec
>>> P = ParamSpec("P")
>>> c = Concatenate[int, P]
>>> get_origin(c)
typing_extensions.Concatenate
>>> get_args(c)
(<class 'int'>, ~P)
```

AlexWaygood: > I couldn't figure out what's supposedly wrong with Concatenate, so I'm going to add a note to ParamSpec only.

Makes sense, though #110 might be worth a mention on the Concatenate entry. I took a brief look at it a while back, and it looks... challenging to backport.

AlexWaygood: #48 as well...

AlexWaygood (approved):
Looks great!

doc/index.rst
JelleZijlstra: Good point, added those. I also noticed that the :gh: links to CPython weren't working, so I added a custom Sphinx extension to make them work (possibly overkill?).

AlexWaygood: > possibly overkill?

Definitely a Choice, but I'll support you in it

JelleZijlstra:
Next step: get this deployed. Other followups:

- Migrate changelog to RST and put it on this docs page
- Trim down the README to remove redundancy with this page
- Consider extending the introduction, maybe a section specifically on the versioning scheme and/or backward compatibility

PR #167, opened 2023-05-21
Improve some docstrings

AlexWaygood:
`help(typing_extensions.TypeVar)` gives quite poor results at the moment.

~~N.B. I haven't included a test for `TypeVar.__doc__`, as it feels like the test I'd add would fit most naturally into the `TestCase` subclass @JelleZijlstra is adding in #165.~~

^Tests added for all three classes.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -4526,6 +4526,9 @@ class TypeVarTests(BaseTestCase):
         with self.assertRaises(ValueError):
             TypeVar('T', contravariant=True, infer_variance=True)
 
+    def test_docstring(self):
+        self.assertEqual(TypeVar.__doc__, typing.TypeVar.__doc__)
+
 
 class TypeVarLikeDefaultsTests(BaseTestCase):
     def test_typevar(self):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -3425,6 +3425,10 @@ class ParamSpecTests(BaseTestCase):
         # won't be the same.
         self.assertNotEqual(hash(ParamSpec('P')), hash(P))
 
+    @skipUnless(hasattr(typing, "ParamSpec"), "Only relevant if typing.ParamSpec exists")
+    def test_consistent_docstring(self):
+        self.assertEqual(ParamSpec.__doc__, typing.ParamSpec.__doc__)
+
 
 class ConcatenateTests(BaseTestCase):
     def test_basics(self):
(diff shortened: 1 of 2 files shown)


src/test_typing_extensions.py
```diff
@@ -3426,6 +3426,10 @@ def test_eq(self):
         # won't be the same.
         self.assertNotEqual(hash(ParamSpec('P')), hash(P))
 
+    @skipUnless(hasattr(typing, "ParamSpec"), "Only relevant if typing.ParamSpec exists")
+    def test_consistent_docstring(self):
+        self.assertEqual(ParamSpec.__doc__, typing.ParamSpec.__doc__)
```
JelleZijlstra, line 3431: I don't necessarily want this test to pass in the future

JelleZijlstra (changes requested):
We should actually have our own docstrings, since we add the `default=` argument to all of these, and it should be covered in the docstring.

AlexWaygood:
Hmm, and that makes me realise that `typing.TypeVar.__doc__` is pretty out of date on the CPython `main` branch:

```pycon
>>> import typing
>>> print(typing.TypeVar.__doc__)
Type variable.

Usage::

  T = TypeVar('T')  # Can be anything
  A = TypeVar('A', str, bytes)  # Must be str or bytes

Type variables exist primarily for the benefit of static type
checkers.  They serve as the parameters for generic types as well
as for generic function definitions.  See class Generic for more
information on generic types.  Generic functions work as follows:

  def repeat(x: T, n: int) -> List[T]:
      '''Return a list containing n references to x.'''
      return [x]*n

  def longest(x: A, y: A) -> A:
      '''Return the longest of two strings.'''
      return x if len(x) >= len(y) else y

The latter example's signature is essentially the overloading
of (str, str) -> str and (bytes, bytes) -> bytes.  Also note
that if the arguments are instances of some subclass of str,
the return type is still plain str.

At runtime, isinstance(x, T) and issubclass(C, T) will raise TypeError.

Type variables defined with covariant=True or contravariant=True
can be used to declare covariant or contravariant generic types.
See PEP 484 for more details. By default generic types are invariant
in all type variables.

Type variables can be introspected. e.g.:

  T.__name__ == 'T'
  T.__constraints__ == ()
  T.__covariant__ == False
  T.__contravariant__ = False
  A.__constraints__ == (str, bytes)

Note that only type variables defined in global scope can be pickled.
```

JelleZijlstra:
I suppose I can improve it in my PEP 695 docs PR.

JelleZijlstra:
Do you want to carry over the docstring I pushed to python/cpython#104642?

AlexWaygood:
This has merge conflicts now, and I'll need to basically start on scratch once we've finished haggling over the wording of the docstring over at https://github.com/python/cpython/pull/104642 :)

So, closing for now

PR #169, opened 2023-05-21
Shorten README, link to docs page

JelleZijlstra:


AlexWaygood (approved):
The "Overview" section could probably also be shorter, but this is great

AlexWaygood (approved):

PR #174, opened 2023-05-22
Improve the repr() of `_marker`

AlexWaygood:
This makes the output from `help(typing_extensions.TypeVar)` much more readable.

Before:

```pycon
>>> help(typing_extensions.TypeVar)
Help on class TypeVar in module typing:

class TypeVar(builtins.object)
 |  TypeVar(name, *constraints, bound=None, covariant=False, contravariant=False, default=<object object at 0x0000028D453A8760>, infer_variance=False)
```

After:

```pycon
>>> help(typing_extensions.TypeVar)
Help on class TypeVar in module typing:

class TypeVar(builtins.object)
 |  TypeVar(name, *constraints, bound=None, covariant=False, contravariant=False, default=<sentinel>, infer_variance=False)
```

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -95,6 +95,7 @@ GenericMeta = type
 # The functions below are modified copies of typing internal helpers.
 # They are needed by _ProtocolMeta and they provide support for PEP 646.
 
+
 class _Sentinel:
     def __repr__(self):
         return "<sentinel>"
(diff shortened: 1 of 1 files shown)


AlexWaygood:
The "help on class TypeVar in module typing" thing, when it's actually giving you help on a class from typing_extensions, is a bit weird. But that's a separate issue...

src/typing_extensions.py
```diff
@@ -95,7 +95,12 @@
 # The functions below are modified copies of typing internal helpers.
 # They are needed by _ProtocolMeta and they provide support for PEP 646.
 
-_marker = object()
+class _Sentinel:
```
AlexWaygood, line 98:
```suggestion

class _Sentinel:
```

PR #175, opened 2023-05-22
Remove __module__ assignment for TypeVar and friends

JelleZijlstra:


AlexWaygood (approved):
LGTM, considering no tests fail!

Seems like the `__module__ = 'typing'` thing was added in fafb5c5 (#77), but I can't find any rationale for why it was done in that PR thread. (Don't know if @cdce8p remembers?)

For context for other readers, the current situation results in some slightly bizarre behaviour when you call `help()` or `repr()` on `typing_extensions.TypeVar` (it claims to be from `typing`, even though it has some important differences to `typing.TypeVar`):

```pycon
Python 3.12.0a7+ (heads/main:be0c106789, May 21 2023, 12:25:53) [MSC v.1932 64 bit (AMD64)] on win32
>>> import typing_extensions, typing
>>> typing_extensions.TypeVar
<class 'typing.TypeVar'>
>>> typing_extensions.TypeVar("T", default=int)
~T
>>> typing.TypeVar
<class 'typing.TypeVar'>
>>> typing.TypeVar("T", default=int)
TypeError: 'default' is an invalid keyword argument for typevar()
>>> typing_extensions.TypeVar is typing.TypeVar
False
>>> help(typing_extensions.TypeVar)
Help on class TypeVar in module typing:

class TypeVar(builtins.object)
 |  TypeVar(name, *constraints, bound=None, covariant=False, contravariant=False, default=<sentinel>, infer_variance=False)
 |
 |  Type variable.
 |
 |  Class methods defined here:
 |
 |  __init_subclass__() -> None from typing_extensions._TypeVarMeta
 |      This method is called when a class is subclassed.
 |
 |      The default implementation does nothing. It may be
 |      overridden to extend subclasses.
 |
 |  ----------------------------------------------------------------------
 |  Data descriptors defined here:
 |
 |  __dict__
 |      dictionary for instance variables (if defined)
 |
 |  __weakref__
 |      list of weak references to the object (if defined)
```

cdce8p:
> Seems like the `__module__ = 'typing'` thing was added in [fafb5c5](https://github.com/python/typing_extensions/commit/fafb5c597f586e8109a1c668c706264060f5456a) (#77), but I can't find any rationale for why it was done in that PR thread. (Don't know if @cdce8p remembers?)

Not sure anymore unfortunately. It's likely I missed that while copying the implementation from cpython.

AlexWaygood (approved):

PR #176, opened 2023-05-22
Further simplify the implementations of the TypeVarLikes

AlexWaygood:
Use one metaclass instead of three. Do the magic in `__new__` on the class instead of `__call__` on the metaclass.


JelleZijlstra (approved):
good call

PR #184, opened 2023-05-23
Have `Protocol` inherit from `typing.Generic` on 3.8+

### LINKED ISSUE #181: Cannot parameterise `typing_extensions.Protocol` with a `ParamSpec` or `TypeVarTuple` on py310+
> This change seems to have broken the latest version of [pydash](https://github.com/dgilland/pydash) when used with typing-extensions 4.6.0. There's a bug report about this here: https://github.com/dgilland/pydash/issues/197.

> I did a bisect of every change between typing-extensions 4.5.0 and 4.6.0, and this was pointed out as the culprit. I've had a look at the code changes here but don't have a good enough grasp of what's going wrong that is causing this to fail.

> This seems to be the relevant Pydash code that relates to the error:
https://github.com/dgilland/pydash/blob/051fe69c3e523f903a0a0ea6ca6a5c2d4b83a3e7/src/pydash/utilities.py#L577-L638

_Originally posted by @Tenzer in https://github.com/python/typing_extensions/issues/137#issuecomment-1558823709_

AlexWaygood:
Fixes #181.

The diff GitHub is displaying for this PR is a mess to read, but it's actually a fairly simple change. We branch on whether `sys.version_info >= (3, 8)` or not. If we're on 3.8+, we inherit from `typing.Generic` (like `Protocol` does in CPython). If we're on 3.7, we don't (because `typing.Generic` won't let us on Python 3.7).

Inheriting from `typing.Generic` is _very_ useful for us, because it means that we no longer have to worry about keeping `typing_extensions.Protocol.__class_getitem__` in sync with the logic in `typing.Generic.__class_getitem__`. The root cause of #181 was that the two methods got out of sync; changes that were made to `typing.Generic.__class_getitem__` were never backported to `typing_extensions.Protocol.__class_getitem__`.

Other than the fact that `Protocol` defines `__class_getitem__` on 3.7, but doesn't on 3.8, the implementations of `Protocol` across the two branches are broadly the same (and broadly the same as they were before this PR).

The one complication is that I had to work around this line in CPython here: https://github.com/python/cpython/blob/08b4eb83aadcbdb389b5970b51cac9be95146c2a/Lib/typing.py#L1024. `typing.py` does this to determine whether a class is `Generic` or `Protocol`, and we go down the wrong code path if `typing.py` determines that `typing_extensions.Protocol` is a different class to `typing.Protocol`

```py
is_generic_or_protocol = cls in (Generic, Protocol)
```

To work around that, I define `__eq__` on `_ProtocolMeta` so that `typing.Protocol == typing_extensions.Protocol` evaluates to `True`. This is a ghastly hack, but, well, it works. (As a result of defining `__eq__`, I end up also having to define `__hash__`, or Python starts complaining that `typing_extensions.Protocol` isn't hashable, and loads of tests start failing.)

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -3827,9 +3827,11 @@ class UnpackTests(BaseTestCase):
                 self.assertEqual(
                     klass[int, str, float, bool].__args__, (int, str, float, bool)
                 )
-                # TODO This should probably also fail on 3.11,
-                # pending changes to CPython.
-                if not TYPING_3_11_0:
+                # A bug was fixed in 3.11.1
+                # (https://github.com/python/cpython/commit/74920aa27d0c57443dd7f704d6272cca9c507ab3)
+                # That means this assertion doesn't pass on 3.11.0,
+                # but it passes on all other Python versions
+                if sys.version_info[:3] != (3, 11, 0):
                     with self.assertRaises(TypeError):
                         klass[int]
 
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -2625,16 +2625,19 @@ class ProtocolTests(BaseTestCase):
         self.assertEqual(MemoizedFunc.__parameters__, (P, T, T2))
         self.assertTrue(MemoizedFunc._is_protocol)
 
-        with self.assertRaisesRegex(TypeError, "Too few arguments"):
+        with self.assertRaises(TypeError):
             MemoizedFunc[[int, str, str]]
 
-        X = MemoizedFunc[[int, str, str], T, T2]
-        self.assertEqual(X.__parameters__, (T, T2))
-        self.assertEqual(X.__args__, ((int, str, str), T, T2))
+        if sys.version_info >= (3, 10):
+            # These unfortunately don't pass on <=3.9,
+            # due to typing._type_check on older Python versions
+            X = MemoizedFunc[[int, str, str], T, T2]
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -2613,6 +2613,57 @@ class CustomProtocolWithoutInitB(Protocol):
...
+            cache: typing.Dict[T2, T]
+            def __call__(self, *args: Unpack[Ts]) -> T: ...
+
+        self.assertEqual(MemoizedFunc.__parameters__, (Ts, T, T2))
+        self.assertTrue(MemoizedFunc._is_protocol)
+
+        things = "arguments" if sys.version_info >= (3, 11) else "parameters"
+
+        with self.assertRaisesRegex(TypeError, f"Too few {things}"):
+            MemoizedFunc[int]
```
AlexWaygood, line 2657:
```suggestion
        # A bug was fixed in 3.11.1
        # (https://github.com/python/cpython/commit/74920aa27d0c57443dd7f704d6272cca9c507ab3)
        # That means this assertion doesn't pass on 3.11.0,
        # but it passes on all other Python versions
        if sys.version_info[:3] != (3, 11, 0):
            with self.assertRaisesRegex(TypeError, f"Too few {things}"):
                MemoizedFunc[int]
```

JelleZijlstra (approved):

dolfinus:
This PR caused #188

PR #188, opened 2023-05-24
Fix isinstance check for Generic classes

dolfinus:
After upgrading to `typing-extensions==4.6.1` this code example started failing on Python 3.7:

```python
from typing_extensions import Protocol, runtime_checkable
from typing import Generic, TypeVar
from pathlib import Path
from dataclasses import dataclass

@runtime_checkable
class ContainsException(Protocol):
    """
    Protocol for objects containing ``.exception`` attribute
    """

    @property
    def exception(self) -> Exception:
        """
        Exception object with traceback
        """

T = TypeVar("T")

@dataclass(eq=False, frozen=True)
class PathContainer(Generic[T]):
    path: T
    exception: Exception

@dataclass(eq=False, frozen=True)
class PathWithFailure(PathContainer[Path]):
    pass


file = PathWithFailure(path=Path('/some'), exception=RuntimeError("some")
isinstance(file, ContainsException)
```

```
  File "/home/maxim/Repo/typing_extensions/1.py", line 33, in <module>
    isinstance(file, ContainsException)
  File "/home/maxim/Repo/typing_extensions/src/typing_extensions.py", line 599, in __instancecheck__
    if super().__instancecheck__(instance):
  File "/home/maxim/.pyenv/versions/3.7.8/lib/python3.7/abc.py", line 139, in __instancecheck__
    return _abc_instancecheck(cls, instance)
  File "/home/maxim/Repo/typing_extensions/src/typing_extensions.py", line 583, in __subclasscheck__
    return super().__subclasscheck__(other)
  File "/home/maxim/.pyenv/versions/3.7.8/lib/python3.7/abc.py", line 143, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
  File "/home/maxim/Repo/typing_extensions/src/typing_extensions.py", line 661, in _proto_hook
    and other._is_protocol
AttributeError: type object 'PathWithFailure' has no attribute '_is_protocol'
```

Who said that objects of generic classes must have `_is_protocol` property?


src/typing_extensions.py
```diff
@@ -659,7 +658,7 @@ def _proto_hook(cls, other):
                     isinstance(annotations, collections.abc.Mapping)
                     and attr in annotations
                     and issubclass(other, (typing.Generic, _ProtocolMeta))
-                    and other._is_protocol
+                    and getattr(other, "_is_protocol", False)
```
AlexWaygood, line 661: Sorry for the breakage here. This was my mistake.

Subclasses of `typing.Generic` _are_ guaranteed to have an `_is_proto` class attribute on Python 3.8+, but not on Python 3.7. It was my fault for not spotting that when writing #184.

Could you possibly add an explanatory comment, something like this? It should help us maintain the code in the future

```suggestion
                    # All subclasses of Generic have an _is_proto attribute on 3.8+
                    # But not on 3.7
                    and getattr(other, "_is_protocol", False)
```

AlexWaygood (commented):
Thank you! I think this probably also needs a changelog entry (under a new "Unreleased" heading at the top of the file)

src/typing_extensions.py
dolfinus: Fixed

AlexWaygood:
(A test would also be great, to ensure that this doesn't regress again)

dolfinus:
Added test and changelog item

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Fix failing `type object ... has no attribute '_is_protocol'` on Python 3.7
```
AlexWaygood, line 3: Feel free to give yourself credit in the CHANGELOG entry, as well!

```suggestion
- Fix regression in version 4.6.1 where comparing a generic class against a 
  runtime-checkable protocol using `isinstance()` would cause `AttributeError`
  to be raised if using Python 3.7
```

src/test_typing_extensions.py
```diff
@@ -4320,6 +4320,24 @@ class Y(Generic[T], NamedTuple):
...
+        class Proto(Protocol):
+            @property
+            def x(self):
+                ...
+
+        z = Z(3)
+        self.assertIs(type(z), Z)
+        self.assertIsInstance(z, Z)
+        self.assertIsInstance(z, Proto)
+        self.assertEqual(z.x, 3)
```
AlexWaygood, line 4339: This feels like an odd place to put a test for this issue, as this class is dedicated for `NamedTuple` tests, and this issue has nothing to do with `NamedTuple`s.

I would create a new method in the `ProtocolTests` class here, and add your test in the new method:

https://github.com/python/typing_extensions/blob/f62fa3f37ce1949e8a536691b30c59c4889445eb/src/test_typing_extensions.py#L1678

AlexWaygood (approved):
LGTM. I confirmed that the new test fails on `main`, and passes with this PR. Sorry again for the breakage!

JelleZijlstra:
This fix was released as part of version 4.6.2 just now.

Issue #50, opened 2022-06-23
Make typing_extensions a drop in for typing


Gobot1234:
Currently I find it annoying trying to remember which types from typing_extensions are and aren't in the versions of typing that I have access to. I think this could be solved by making typing_extensions a fully drop in replacement. I think this would also help with some older issues like subclassing Generic with a ParamSpec as a parameter used to be.

https://github.com/python/typing_extensions/issues/50#tasklist-block-080df15f-51ca-4e49-9aad-58c84588c2ee

srittau:
I would be fine with this, but I am mostly indifferent to it.

vwxyzjn:
Yeah, this is very inconvenient. I tried using it as a drop-in replacement because we have dozens of files, but it doesn't work: 

`from typing_extensions import Literal, Mapping, Optional, Tuple, Union`

yields

`ImportError: cannot import name 'Mapping' from 'typing_extensions'`

vwxyzjn:
Effectively we needed to do things like

```
from typing import Literal, Mapping, Optional, Tuple, Union
try:
    import Literal
except ImportError:
    from typing extensions import Literal
```

which is a labor-intensive process and not that maintainable.

JelleZijlstra:
I'm sympathetic to this idea; it's hard to remember what you need to import from typing-extensions depending on the Python version. At runtime we can simply add `from typing import *` to typing_extensions.py. The stubs for typing-extensions can work similarly. Some type checkers may also need minor changes.

For reference, these are the names in `typing.__all__` but not `typing_extensions.__all__`.

```
{'AbstractSet',
 'Any',
 'AnyStr',
 'BinaryIO',
 'ByteString',
 'Callable',
 'Collection',
 'Container',
 'Dict',
 'ForwardRef',
 'FrozenSet',
 'Generator',
 'Generic',
 'Hashable',
 'IO',
 'ItemsView',
 'Iterable',
 'Iterator',
 'KeysView',
 'List',
 'Mapping',
 'MappingView',
 'Match',
 'MutableMapping',
 'MutableSequence',
 'MutableSet',
 'NamedTuple',
 'Optional',
 'Pattern',
 'Reversible',
 'Sequence',
 'Set',
 'Sized',
 'SupportsAbs',
 'SupportsBytes',
 'SupportsComplex',
 'SupportsFloat',
 'SupportsInt',
 'SupportsRound',
 'TextIO',
 'Tuple',
 'TypeVar',
 'Union',
[...]

PR #189, opened 2023-05-24
Add guidance on robust runtime use

JelleZijlstra:


AlexWaygood (approved):
Looks great!

hauntsaninja (approved):
Looks good! We should definitely do #50 to make this easier.

PR #194, opened 2023-05-25
Don't reimplement `runtime_checkable` on py38+

AlexWaygood:
Re-export it from `typing` instead.

Prior to #184, `typing.runtime_checkable` didn't work on `typing_extensions.Protocol`. Now that `typing_extensions.Protocol` inherits from `typing.Generic` on py38+, however, it does! This means that we can just re-export `typing.runtime_checkable` on py38+, rather than reimplementing it on all versions <3.12.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+- Due to changes in the implementation of `typing_extensions.Protocol`,
+  `typing.runtime_checkable` can now be used on `typing_extensions.Protocol`
+  (previously, users had to use `typing_extensions.runtime_checkable` if they
+  were using `typing_extensions.Protocol`).
+
 # Release 4.6.2 (May 25, 2023)
 
 - Fix use of `@deprecated` on classes with `__new__` but no `__init__`.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# Unreleased
+- Due to changes in the implementation of `typing_extensions.Protocol`,
+  `typing.runtime_checkable` can now be used on `typing_extensions.Protocol`
```
JelleZijlstra, line 3: Can you add a versionchanged to the documentation for this?

AlexWaygood: I had a go in 00f568b, but I'm not really sure we should be advertising this too widely. Users shouldn't really be depending on being able to use `typing.runtime_checkable` on `typing_extensions.Protocol`; it's more of a "nice-to-have" (and it will simplify the implementation for _us_ when we drop support for Python 3.7).

JelleZijlstra: Yes, that documentation is very confusing to read, we're probably better off without it. Sorry for the churn!

AlexWaygood: No worries!

JelleZijlstra (approved):

PR #196, opened 2023-05-25
Re-export all names from typing

### LINKED ISSUE #50: Make typing_extensions a drop in for typing

JelleZijlstra:
Fixes #50.

I decided to exclude top-level names that have a definite removal
timeline (io, re, ByteString); there's no point in re-exporting them
now.

Many other names are deprecated without a definite removal plan
(PEP 585). If they do get removed eventually, typing-extensions can
simply re-export the builtin/collections.abc names.

In the documentation, I add a separate section listing all the names
that are pure aliases for typing. This makes it easier to distinguish
the cases where there is something more interesting going on in
typing-extensions.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -4118,6 +4118,12 @@ class AllTests(BaseTestCase):
         te_all = set(typing_extensions.__all__)
         exceptions = {"ByteString"}
         self.assertGreaterEqual(te_all, t_all - exceptions)
+        # Deprecated, to be removed in 3.14
+        self.assertFalse(hasattr(typing_extensions, "ByteString"))
+        # These were never included in `typing.__all__`,
+        # and have been removed in Python 3.13
+        self.assertNotIn('re', te_all)
+        self.assertNotIn('io', te_all)
 
     def test_typing_extensions_includes_standard(self):
         a = typing_extensions.__all__
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -4111,6 +4111,14 @@ class DataclassTransformTests(BaseTestCase):
 
 class AllTests(BaseTestCase):
 
+    def test_drop_in_for_typing(self):
+        # Check that the typing_extensions.__all__ is a superset of
+        # typing.__all__.
+        t_all = set(typing.__all__)
+        te_all = set(typing_extensions.__all__)
+        exceptions = {"ByteString"}
+        self.assertGreaterEqual(te_all, t_all - exceptions)
+
     def test_typing_extensions_includes_standard(self):
         a = typing_extensions.__all__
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
(Let's wait at least a few days to merge this in case we need another bugfix release.)

AlexWaygood (approved):

src/typing_extensions.py
```diff
@@ -2873,3 +2912,43 @@ def __ror__(self, left):
                 if not _is_unionable(left):
                     return NotImplemented
                 return typing.Union[left, self]
+
+
+# Aliases for items that have always been in typing
```
AlexWaygood, line 2917: Could be good to add a comment here outlining (one reason) why we're not going with the apparently-more-simple approach

```suggestion
# Aliases for items that have always been in typing
# Explicitly assign these (rather than using `from typing import *` at the top),
# so that we get a CI error if one of these is deleted from typing.py
# in a future version of Python
```

src/test_typing_extensions.py
```diff
@@ -4111,6 +4111,14 @@ class CustomerModel(ModelBase, init=False):
 
 class AllTests(BaseTestCase):
 
+    def test_drop_in_for_typing(self):
+        # Check that the typing_extensions.__all__ is a superset of
+        # typing.__all__.
+        t_all = set(typing.__all__)
+        te_all = set(typing_extensions.__all__)
+        exceptions = {"ByteString"}
+        self.assertGreaterEqual(te_all, t_all - exceptions)
```
AlexWaygood, line 4120:
```suggestion
        self.assertGreaterEqual(te_all, t_all - exceptions)
        # Deprecated, to be removed in 3.14
        self.assertFalse(hasattr(typing_extensions, "ByteString"))
        # These were never included in `typing.__all__,
        # and have been removed in Python 3.13
        self.assertNotIn('re', te_all)
        self.assertNotIn('io', te_all)
```

src/typing_extensions.py
JelleZijlstra: There's actually a more concrete reason: `ByteString` is in `__all__`, and we don't want it here.

AlexWaygood (approved):
LGTM (agree on holding off on merging for now, though, in case we need another bugfix release)

hauntsaninja (approved):
Thanks for working on this!

PR #197, opened 2023-05-26
Add more detailed versioning policy

### LINKED ISSUE #185: Use pre releases to allow community find bug or patch its packages.
The `typing-extension` package is widely used in many projects. It happens that sometimes some package does to strong assumptions (https://github.com/pydantic/pydantic/issues/5821). 

It would be nice if `typing_extension` could create a pre (rc) release and give a few days (for example week) for the community to test it. (example of a project that partially uses this approach is, for example, numpy https://pypi.org/project/numpy/#history)

JelleZijlstra:
In particular, introduce a release candidate phase.

Fixes #185

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,10 +23,13 @@ CPython's `main` branch.
 # Versioning scheme
 
 Starting with version 4.0.0, `typing_extensions` uses
-[Semantic Versioning](https://semver.org/). The major version is incremented for all
-backwards-incompatible changes.
+[Semantic Versioning](https://semver.org/). See the documentation
+for more detail.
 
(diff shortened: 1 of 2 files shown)


srittau:
I would personally only do rc releases if existing functionality changes. When a release only contains new classes and functions (for new PEPs etc.), regressions are very unlikely. But as you are the RM I am happy with any policy that you are happy with.

srittau (approved):
See my comment above, but I am also fine with this.

JelleZijlstra:
I think it's useful to do for new features too (e.g., there was a bug in `@deprecated` that we fixed this week), and in any case recently we've made a lot more changes to existing functionality than adding new stuff.

PR #201, opened 2023-05-26
Minor improvements to CI workflows

AlexWaygood:
- Use the new `allow-prereleases` key for `actions/setup-python`. If we set this to `true`, then we can just use `3.12` in our CI matrix, instead of `3.12-dev`. That means we won't have to worry about updating the matrix from `3.12-dev` to `3.12` when 3.12.0 comes out in the autumn; our CI will automatically start using the latest version of Python 3.12.
- Colorise the output of flake8.
- Don't bother upgrading pip in CI. It's not really necessary, and wastes a few seconds. Disable the "you're using an older version of pip" message instead, using the `PIP_DISABLE_PIP_VERSION_CHECK` environment variable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,9 @@ on:
 permissions:
   contents: read
 
+env:
+  PIP_DISABLE_PIP_VERSION_CHECK: 1
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Docs on `allow-prereleases`: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases

srittau (approved):

PR #211, opened 2023-06-02
Document how to run tests

### LINKED ISSUE #8: Better protections for typing-extensions tests
Followup from python/typing#935:
- Add assertion to make sure the typing-extensions we're testing is the one in the repo, not the one that happens to be installed in your venv.
- Document how to test typing-extensions locally (probably, we should recommend using an empty venv).

JelleZijlstra:
Fixes #8

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,25 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
+# Running tests
+
+Testing `typing-extensions` can be tricky because many development tools depend on
+`typing-extensions`, so you may end up testing some installed version of the library,
+rather than your local code.
+
+The simplest way to run the tests locally is:
+
+- `cd src/`
+- `python test_typing_extensions.py`
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -26,6 +26,25 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
+# Running tests
+
+Testing `typing-extensions` can be tricky because many development tools depend on
```
AlexWaygood, line 31:
```suggestion
Testing `typing_extensions` can be tricky because many development tools depend on
```

CONTRIBUTING.md
```diff
@@ -26,6 +26,25 @@ Starting with version 4.0.0, `typing_extensions` uses
...
+`typing-extensions`, so you may end up testing some installed version of the library,
```
AlexWaygood, line 32:
```suggestion
`typing_extensions`, so you may end up testing some installed version of the library,
```

AlexWaygood (commented):
Nit: we refer to it as `typing_extensions` (underscore, not hyphen) elsewhere in this document

srittau (approved):

PR #220, opened 2023-06-04
Allow Protocols to inherit from typing_extensions.Buffer

AlexWaygood:
Backport of https://github.com/python/cpython/pull/104827

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -2745,6 +2745,23 @@ class ProtocolTests(BaseTestCase):
         class CustomContextManager(typing.ContextManager, Protocol):
             pass
 
+    def test_typing_extensions_protocol_allowlist(self):
+        @runtime_checkable
+        class ReleasableBuffer(Buffer, Protocol):
+            def __release_buffer__(self, mv: memoryview) -> None: ...
+
+        class C: pass
+        class D:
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -2745,6 +2745,23 @@ class CustomProtocol(TestCase, Protocol):
         class CustomContextManager(typing.ContextManager, Protocol):
             pass
 
+    def test_typing_extensions_protocol_allowlist(self):
```
AlexWaygood, line 2748:
```suggestion
    @skip_if_py312b1
    def test_typing_extensions_protocol_allowlist(self):
```

src/test_typing_extensions.py
```diff
@@ -2737,6 +2737,28 @@ def close(self):
         self.assertIsSubclass(B, Custom)
         self.assertNotIsSubclass(A, Custom)
 
+    @skipUnless(
+        hasattr(collections.abc, "Buffer"),
+        "needs collections.abc.Buffer to exist"
+    )
+    @skip_if_py312b1
```
AlexWaygood, line 2744: This test isn't run in CI at all right now, but I've confirmed locally that it passes with the CPython `main` branch (so when 3.12b2 comes along, it should pass in CI for us here)

srittau (approved):

PR #224, opened 2023-06-07
Fix pydantic tests

### LINKED ISSUE #222: Third-party tests failed on Wed Jun 07 2023

AlexWaygood:
I can't say I really understand what's going on here, but this seems to fix the CI errors.

Resolves #222

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -53,9 +53,7 @@ jobs:
       - name: Setup pdm for pydantic tests
         uses: pdm-project/setup-pdm@v3
         with:
-          version: 1.7.0
           python-version: ${{ matrix.python-version }}
-          cache: true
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
I guess we may want to put the cache back later?

AlexWaygood:
> I guess we may want to put the cache back later?

The cache presumably speeds things up, but also increases the chance of breakages like this. Tbh, since we don't have much control over these third-party tests, I'm inclined to go for the option that minimises the risk of breakages like this, so perhaps we're better off without the cache for good? We only run these tests on a daily basis anyway, so it's not like the test taking a little longer has a big impact on our developer experience -- it just increases our carbon footprint a little.

JelleZijlstra:
> We only run these tests on a daily basis anyway

They also get run on PRs, but I think your point still stands.

PR #225, opened 2023-06-07
Docs: Add a recipe for robust runtime introspection

AlexWaygood:
Followup to #203

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -73,6 +73,54 @@ the risk of compatibility issues:
   attributes directly. If some information is not available through a public
   attribute, consider opening an issue in CPython to add such an API.
 
+Here is an example recipe for a general-purpose function that could be used for
+reasonably performant runtime introspection of typing objects. The function
+will be resilient against any potential changes in ``typing_extensions`` that
+alter whether an object is reimplemented in ``typing_extensions``, rather than
+simply being re-exported from the :mod:`typing` module::
+
+   import functools
+   import typing
+   import typing_extensions
+   from typing import Tuple, Any
+
+   # Use an unbounded cache for this function, for optimal performance
+   @functools.lru_cache(maxsize=None)
+   def get_typing_objects_by_name_of(name: str) -> Tuple[Any, ...]:
+       result = tuple(
+           getattr(module, name)
+           # You could potentially also include mypy_extensions here,
+           # if your library supports mypy_extensions
+           for module in (typing, typing_extensions)
+           if hasattr(module, name)
+       )
+       if not result:
+           raise ValueError(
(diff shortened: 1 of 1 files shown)


srittau (approved):

AlexWaygood:
I did some benchmarking anyway, using this script (I called it `utils.py` and put it in `typing_extensions/src/`):

```py
import functools
import typing
import typing_extensions
from typing import Tuple, Any
from typing_extensions import get_origin

# Use an unbounded cache for this function, for optimal performance
# @functools.lru_cache(maxsize=None)
def get_typing_objects_by_name_of(name: str) -> Tuple[Any, ...]:
   result = tuple(
       getattr(module, name)
       # You could potentially also include mypy_extensions here,
       # if your library supports mypy_extensions
       for module in (typing, typing_extensions)
       if hasattr(module, name)
   )
   if not result:
       raise ValueError(
           f"Neither typing nor typing_extensions has an object called {name!r}"
       )
   return result


def is_typing_name(obj: object, name: str) -> bool:
   return any(obj is thing for thing in get_typing_objects_by_name_of(name))


is_literal = functools.partial(is_typing_name, name="Literal")


def _bench():
   is_literal(typing.Literal)
   is_literal(typing_extensions.Literal)
   is_literal(typing.Any)
   is_literal(get_origin(typing.Literal[42]))
   is_literal(get_origin(typing_extensions.Final[42]))
```

I ran these commands from the `typing_extensions/src/` directory:

```
>python -m timeit "from utils import _bench" "_bench()"
50000 loops, best of 5: 5.3 usec per loop
```

With the `@lru_cache(maxsize=None)` line commented out:

```
>python -m timeit "from utils import _bench" "_bench()"
50000 loops, best of 5: 8.27 usec per loop
```

doc/index.rst
```diff
@@ -73,6 +73,54 @@ the risk of compatibility issues:
...
+           if hasattr(module, name)
+       )
+       if not result:
+           raise ValueError(
+               f"Neither typing nor typing_extensions has an object called {name!r}"
+           )
+       return result
+
+
+   def is_typing_name(obj: object, name: str) -> bool:
```
AlexWaygood, line 104:
```suggestion
   # Use a cache here as well, but make it a bounded cache
   # (the default cache size is 128)
   @functools.lru_cache()
   def is_typing_name(obj: object, name: str) -> bool:
```

AlexWaygood:
Adding the second cache speeds it up further, to `1.99 usec per loop`. I think that's reasonable, though I think it makes sense to make that one a bounded cache rather than an unbounded one.

srittau (approved):

PR #237, opened 2023-06-13
Allow typing_extensions.Protocol and typing.Protocol to mix

### LINKED ISSUE #236: Can't mix `typing.Protocol` and `typing_extensions.Protocol` anymore as base classes
When inheriting from a typing-based Protocol and a typing-extensions-based Protocol, I get a metaclass error:

```
Python 3.10.11 (main, Apr 20 2023, 19:02:41) [GCC 11.2.0] on linux
>>> import typing
>>> import typing_extensions
>>> class P1(typing.Protocol): ...
... 
>>> class P2(typing_extensions.Protocol): ...
... 
>>> class P3(P1, P2, typing.Protocol): ...
... 
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
>>> class P3(P1, P2, typing_extensions.Protocol): ...
... 
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

Is this expected?

(This is with typing-extensions 4.6.3.)

AlexWaygood:
Fixes #236.

I don't really like this patch; it feels like it makes the code hackier and more fragile, and I'm not sure what the use case is. I feel like I'd prefer to just document that you can't mix the two `Protocol` versions. This is just a proof of concept to show how it could be fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,8 @@
 - Allow `Protocol` classes to inherit from `typing_extensions.Buffer` or
   `collections.abc.Buffer`. Patch by Alex Waygood (backporting
   https://github.com/python/cpython/pull/104827, by Jelle Zijlstra).
-- Allow classes to have both `typing.Protocol` and `typing_extensions.Protocol`
-  in their mro. Since v4.6.0, this caused `TypeError` to be raised due to a
+- Allow classes to inherit from both `typing.Protocol` and `typing_extensions.Protocol`
+  simultaneously. Since v4.6.0, this caused `TypeError` to be raised due to a
   metaclass conflict. Patch by Alex Waygood.
 
 # Release 4.6.3 (June 1, 2023)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
 - Allow `Protocol` classes to inherit from `typing_extensions.Buffer` or
   `collections.abc.Buffer`. Patch by Alex Waygood (backporting
   https://github.com/python/cpython/pull/104827, by Jelle Zijlstra).
+- Allow classes to have both `typing.Protocol` and `typing_extensions.Protocol`
+  in their mro. Since v4.6.0, this cause `TypeError` to be raised due to a
+  metaclass conflict. Patch by Alex Waygood.
 
 # Release 4.6.3 (June 1, 2023)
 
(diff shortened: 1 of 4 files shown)


AlexWaygood:
I still don't much like the patch, but it does sound like this is something we should probably fix.

AlexWaygood:
[No breakage](https://github.com/AlexWaygood/typing_extensions/actions/runs/5258080810) to any of the third-party test suites we run in the daily workflow.

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@
 - Allow `Protocol` classes to inherit from `typing_extensions.Buffer` or
   `collections.abc.Buffer`. Patch by Alex Waygood (backporting
   https://github.com/python/cpython/pull/104827, by Jelle Zijlstra).
+- Allow classes to have both `typing.Protocol` and `typing_extensions.Protocol`
+  in their mro. Since v4.6.0, this caused `TypeError` to be raised due to a
```
JelleZijlstra, line 28:
```suggestion
- Allow classes to inherit from both `typing.Protocol` and `typing_extensions.Protocol`
  simultaneously. Since v4.6.0, this caused `TypeError` to be raised due to a
```
I feel like "mro" is too jargon-y

doc/index.rst
```diff
@@ -282,6 +282,12 @@ Special typing primitives
       Backported changes to runtime-checkable protocols from Python 3.12,
       including :pr-cpy:`103034` and :pr-cpy:`26067`.
 
+   .. versionchanged:: 4.7.0
+
+      Classes can now have both :py:class:`typing.Protocol` and
+      ``typing_extensions.Protocol`` in their mro. Previously, this led to
```
JelleZijlstra, line 288:
```suggestion
      Classes can now inherit from both :py:class:`typing.Protocol` and
      ``typing_extensions.Protocol`` simultaneously. Previously, this led to
```

src/test_typing_extensions.py
```diff
@@ -1799,6 +1799,26 @@ def __init__(self):
...
+        "Test is only relevant if typing.Protocol exists"
+    )
+    def test_typing_Protocol_and_extensions_Protocol_can_mix(self):
+        class TypingProto(typing.Protocol):
+            x: int
+
+        class ExtensionsProto(Protocol):
+            y: int
+
+        class SubProto(TypingProto, ExtensionsProto, typing.Protocol):
```
JelleZijlstra, line 1813: Also add a test where ExtensionsProto comes before TypingProto.

JelleZijlstra (commented):
Thanks!

src/test_typing_extensions.py
AlexWaygood: Done!

PR #240, opened 2023-06-16
Backport NamedTuple and TypedDict deprecations from Python 3.13

JelleZijlstra:


src/typing_extensions.py
```diff
@@ -1075,13 +1083,14 @@ def __subclasscheck__(cls, other):
 
         __instancecheck__ = __subclasscheck__
 
-    def TypedDict(__typename, __fields=None, *, total=True, **kwargs):
+    def TypedDict(__typename, __fields=_sentinel, *, total=True, **kwargs):
         """A simple typed namespace. At runtime it is equivalent to a plain dict.
 
-        TypedDict creates a dictionary type that expects all of its
+        TypedDict creates a dictionary type such that a type checker will expect all
```
JelleZijlstra, line 1089: I took this opportunity to copy-paste the docstring from CPython.

JelleZijlstra:
Seems like there's some issue with generic namedtuples on 3.12, I'll fix that in a separate PR.

JelleZijlstra:
Actually the failure is due to this PR (it's because we now re-implement NamedTuple on 3.12). Fixed it.

src/typing_extensions.py
```diff
@@ -980,10 +980,19 @@ def __round__(self, ndigits: int = 0) -> T_co:
     # Generic TypedDicts are also impossible using typing.TypedDict on Python <3.11.
     # Aaaand on 3.12 we add __orig_bases__ to TypedDict
     # to enable better runtime introspection.
+    # On 3.13 we deprecate some odd ways of creating TypedDicts.
     TypedDict = typing.TypedDict
     _TypedDictMeta = typing._TypedDictMeta
     is_typeddict = typing.is_typeddict
 else:
+    class _Sentinel:
```
AlexWaygood, line 988: We actually already have this class in typing_extensions; `typing_extensions._marker` is an instance of it. I think we can just reuse `_marker` as our sentinel object for these deprecations

src/typing_extensions.py
```diff
@@ -2624,12 +2658,67 @@ def __new__(cls, typename, bases, ns):
                 nm_tpl.__init_subclass__()
             return nm_tpl
 
-    def NamedTuple(__typename, __fields=None, **kwargs):
-        if __fields is None:
-            __fields = kwargs.items()
+    def NamedTuple(__typename, __fields=_sentinel, **kwargs):
+        """Typed version of namedtuple.
```
AlexWaygood, line 2662: If we're adding a docstring the normal way here, we no longer need to monkey-patch it onto the function afterwards:

https://github.com/python/typing_extensions/blob/f4adac5cb359f74ccbab47a36195f4d5e69a558a/src/typing_extensions.py#L2726

JelleZijlstra: Oh I missed that. I think it's better if we present the same docstring across all versions, so I'll remove the monkey-patching.

src/test_typing_extensions.py
```diff
@@ -5196,7 +5283,6 @@ def test_copy_and_pickle(self):
                 self.assertIsInstance(jane2, cls)
 
     def test_docstring(self):
-        self.assertEqual(NamedTuple.__doc__, typing.NamedTuple.__doc__)
         self.assertIsInstance(NamedTuple.__doc__, str)
```
AlexWaygood, line 5286: Tbh this test method feels pretty redundant now; I'd just delete the whole method :p

AlexWaygood (approved):

src/test_typing_extensions.py
AlexWaygood: I'd hope we'd catch it in manual review if somebody proposed deleting the entire docstring in a PR 😄

When I added this test method, my intent was just to test that the monkeypatching in the backport was working correctly. But I guess we can keep it; it doesn't do any harm :)

The PR LGTM!

PR #242, opened 2023-06-16
Fix edge-case `Protocol` bug on Python 3.7

AlexWaygood:
This line is buggy. As a result of the bug, several of the assertions added in https://github.com/python/cpython/pull/105835 don't pass on Python 3.7:

https://github.com/python/typing_extensions/blob/bc9ce4f9a4e1f38f85fab58328e8dd86741dab45/src/typing_extensions.py#L743

It should be

```py
issubclass(other, typing.Generic) or isinstance(other, _ProtocolMeta)
```

But we can also rewrite the condition more elegantly using the new `is_protocol` function, so that's what I've done in this PR.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -608,7 +608,6 @@ else:
         _typing_Protocol = _marker
         _ProtocolMetaBase = abc.ABCMeta
 
-
     class _ProtocolMeta(_ProtocolMetaBase):
         # This metaclass is somewhat unfortunate,
         # but is necessary for several reasons...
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -604,22 +604,10 @@ def _no_init(self, *args, **kwargs):
...
-
-        def _is_protocol(cls):
-            return (
-                isinstance(cls, type)
-                and issubclass(cls, typing.Generic)
-                and getattr(cls, "_is_protocol", False)
-            )
     else:
         _typing_Protocol = _marker
         _ProtocolMetaBase = abc.ABCMeta
```
AlexWaygood, line 610:
```suggestion
```

PR #244, opened 2023-06-17
Add a few more tests for interactions between `typing.Protocol` and `typing_extensions.Protocol`

AlexWaygood:
I was looking at lines like these:

https://github.com/python/typing_extensions/blob/af89916a21b0b720b413e5618dd119f55c53f7b7/src/typing_extensions.py#L747-L749

https://github.com/python/typing_extensions/blob/af89916a21b0b720b413e5618dd119f55c53f7b7/src/typing_extensions.py#L759-L761

And realised that I wasn't at all sure how the interaction between `typing.Protocol.__init_subclass__` and `typing_extensions.Protocol.__init_subclass__` would work out. It _looks_ to me like it's fine, but I thought a few extra tests might be good anyway.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1842,7 +1842,7 @@ class ProtocolTests(BaseTestCase):
             with self.subTest(cls=cls.__name__):
                 self.assertFalse(is_protocol(cls))
                 # Check that this doesn't raise:
-                cls()
+                self.assertIsInstance(cls(), cls)
                 with self.assertRaises(TypeError):
                     runtime_checkable(cls)
                 with self.assertRaises(TypeError):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1829,6 +1829,9 @@ class ProtocolTests(BaseTestCase):
                     TypeError, "Protocols cannot be instantiated"
                 ):
                     proto()
+                # check these don't raise
+                runtime_checkable(proto)
+                typing.runtime_checkable(proto)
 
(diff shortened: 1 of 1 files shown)


AlexWaygood:
(There is a bug there, actually. But I'll fix it in another PR.)

src/test_typing_extensions.py
```diff
@@ -1817,11 +1817,37 @@ class SubProto3(ExtensionsProto, TypingProto, typing.Protocol):
...
         class Concrete(SubProto): pass
         class Concrete2(SubProto2): pass
         class Concrete3(SubProto3): pass
         class Concrete4(SubProto4): pass
 
+        for cls in Concrete, Concrete2, Concrete3, Concrete4:
+            with self.subTest(cls=cls.__name__):
+                self.assertFalse(is_protocol(cls))
+                # Check that this doesn't raise:
+                cls()
```
JelleZijlstra, line 1845:
```suggestion
                self.assertIsInstance(cls(), cls)
```
Not the most useful thing to test, but this could fail if we did something sufficiently crazy in `__new__`.

JelleZijlstra (approved):
Looks good, thanks.

Should I merge or wait for the bugfix you alluded to?

AlexWaygood:
Merge this -- the bugfix probably isn't fixable? I think we'll have to document it. Writing up an issue now :/

AlexWaygood:
> Merge this -- the bug probably isn't fixable? I think we'll have to document it. Writing up an issue now :/

#245

PR #246, opened 2023-06-17
Document that `typing_extensions.Protocol` and `typing.Protocol` don't always play well together

### LINKED ISSUE #245: `typing_extensions.Protocol` and `typing.Protocol` don't always play well together
If you're just using `typing_extensions.Protocol`, we backport https://github.com/python/cpython/pull/31628 from Python 3.11. That means you can do this (the `__init__` method on a protocol is retained, and can be used in concrete subclasses of that protocol):

```pycon
>>> import typing_extensions as te
>>> class P(te.Protocol):
...     x: int
...     def __init__(self, x: int) -> None:
...         self.x = x
...
>>> class C(P): pass
...
>>> C(1).x
1
```

_But_, if you have both `typing.Protocol` and `typing_extensions.Protocol` in the mro, it doesn't work. On Python 3.8 (and, I assume, 3.9/3.10), there's this behaviour currently:

```pycon
>>> import typing as t, typing_extensions as te
>>> class P(t.Protocol): pass
...
>>> class Q(P, te.Protocol):
...     x: int
...     def __init__(self, x: int) -> None:
...         self.x = x
...
>>> class C(Q): pass
...
>>> C(1).x
AttributeError: 'C' object has no attribute 'x'
```

I've been looking at it, and I think this may not be fixable, unfortunately. Should we document that this is a known limitation?

AlexWaygood:
Closes #245

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -304,6 +304,12 @@ Special typing primitives
       ``typing_extensions.Protocol`` simultaneously. Previously, this led to
       :py:exc:`TypeError` being raised due to a metaclass conflict.
 
+      It is recommended to avoid doing this if possible. Not all features and
+      bugfixes that ``typing_extensions.Protocol`` backports from newer Python
+      versions are guaranteed to work if :py:class:`typing.Protocol` is also
+      present in a protocol class's :py:term:`method-resolution-order`. See
+      :issue:`245` for some examples.
+
 .. data:: Required
 
    See :py:data:`typing.Required` and :pep:`655`. In ``typing`` since 3.11.
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -307,7 +307,7 @@ Special typing primitives
       It is recommended to avoid doing this if possible. Not all features and
       bugfixes that ``typing_extensions.Protocol`` backports from newer Python
       versions are guaranteed to work if :py:class:`typing.Protocol` is also
-      present in a protocol class's :py:term:`method-resolution-order`. See
+      present in a protocol class's :py:term:`method resolution order`. See
       :issue:`245` for some examples.
 
 .. data:: Required


AlexWaygood:
@JelleZijlstra would it be possible to set up readthedocs previews for docs PRs? :)

JelleZijlstra:
> @JelleZijlstra would it be possible to set up readthedocs previews for docs PRs? :)

I've procrastinated figuring out how to do that :)

JelleZijlstra:
Building locally gives

```
/Users/jelle/py/typing_extensions/doc/index.rst:307: WARNING: term not in glossary: 'method-resolution-order'
```

doc/index.rst
```diff
@@ -304,6 +304,12 @@ Special typing primitives
       ``typing_extensions.Protocol`` simultaneously. Previously, this led to
       :py:exc:`TypeError` being raised due to a metaclass conflict.
 
+      It is recommended to avoid doing this if possible. Not all features and
+      bugfixes that ``typing_extensions.Protocol`` backports from newer Python
+      versions are guaranteed to work if :py:class:`typing.Protocol` is also
+      present in a protocol class's :py:term:`method-resolution-order`. See
```
AlexWaygood, line 310:
```suggestion
      present in a protocol class's :py:term:`method resolution order`. See
```

AlexWaygood:
> Building locally gives
> 
> ```
> /Users/jelle/py/typing_extensions/doc/index.rst:307: WARNING: term not in glossary: 'method-resolution-order'
> ```

I'm on my phone, but I _think_ my last push should fix that!

JelleZijlstra:
Screen Shot 2023-06-17 at 1 03 00 PM

Looks great!

PR #249, opened 2023-06-21
Update CONTRIBUTING.md

JelleZijlstra:
- Tell future me how to use twine
- Start using GitHub releases. Somebody (forgot who) asked for this. The
  advantage is that it gives a new channel to notify people about new releases.


AlexWaygood (approved):
LGTM

PR #252, opened 2023-06-23
Backport CPython PR 105976

AlexWaygood:
Backport of https://github.com/python/cpython/pull/105976

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# Unreleased
+
+- Fix bug where a `typing_extensions.Protocol` class that had one or more
+  non-callable members would raise `TypeError` when `issubclass()`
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
Thanks! There's a test failure on 3.12.

Also, let's land this after 4.7.0 final goes out next week?

AlexWaygood:
> Also, let's land this after 4.7.0 final goes out next week?

It fixes a regression introduced in the 4.6 series... But I'm happy to defer to your judgement as RM :)

I do _appear_ to have been the _only_ user of `typing_extensions.Protocol` affected by the regression I introduced 😆

JelleZijlstra:
Ah you're right, we should fix this in 4.7.0 then.

PR #257, opened 2023-06-29
Stop running third-party tests on 3.7

### LINKED ISSUE #256: Third-party tests failed on Thu Jun 29 2023

JelleZijlstra:
typed-argument-parser already dropped support. We will drop support soon too (just waiting for a bit to see whether we need a bugfix release for the 4.7 series), so let's just drop 3.7 from all third-party runs.

Fixes #256.

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -41,7 +41,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
Makes sense, since I imagine several of the others will also be dropping support for 3.7 in short order

hauntsaninja (approved):

Issue #258, opened 2023-06-30
4.7.0: new implementation of TypedDict breaks pypy-3.7 and pypy-3.8


AlexWaygood:
See e.g. a CI run on black PR https://github.com/psf/black/pull/3757: https://github.com/psf/black/actions/runs/5420281960/jobs/9854265747?pr=3757

I'm inclined to think this is a bug in PyPy-3.7 (and it looks like it was fixed in PyPy-3.8). It looks like PyPy-3.7 isn't taking any notice of the `__mro_entries__` method we set on TypedDict. But it might be worth explicitly calling out in our docs that the latest version isn't compatible with PyPy-3.7?

AlexWaygood:
Ah, looks like even PyPy-3.8 is broken: https://github.com/python/typing_extensions/actions/runs/5421296757/jobs/9856554127

A5rocks:
Hi, I've encountered this too! Only on pypy 3.7 here cause turns out pylint subclasses `typing_extensions.TypedDict` when python version <3.8! https://github.com/python-trio/trio/actions/runs/5405876855/jobs/9822036832#step:5:1587 (I'm going to pin `typing_extensions` to below the current version just until we get out our final release that supports 3.7, so this isn't a major problem for us.)

JelleZijlstra:
This looks like a bug in pypy: it doesn't properly support `__mro_entries__`.

```
% python3.8
Python 3.8.16 (default, Dec 26 2022, 20:36:56) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def f(): pass
... 
>>> f.__mro_entries__ = lambda _: (int,)
>>> class x(f): pass
... 
>>> x.__bases__
(<class 'int'>,)
>>> 
% ~/.pyenv/versions/pypy3.8-7.3.11/bin/python
Python 3.8.16 (a9dbdca6fc3286b0addd2240f11d97d8e8de187a, Dec 29 2022, 11:45:30)
[PyPy 7.3.11 with GCC Apple LLVM 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>> def f(): pass
>>>> f.__mro_entries__ = lambda _: (int,)
>>>> class x(f): pass
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #259, opened 2023-06-30
Run tests on more pypy versions in CI

AlexWaygood:
An attempt to make breakages like #258 less likely. I haven't included pypy-3.7, since our tests are currently failing on pypy-3.7 (#258). For the third-party tests we run, I've only included pypy in the matrix if the upstream package actually tests with pypy (and I've only included the specific pypy versions that they test with)


.github/workflows/ci.yml
```diff
@@ -41,7 +41,21 @@ jobs:
...
+          - "3.8"
+          - "3.8.0"
+          - "3.9"
+          - "3.9.0"
+          - "3.10"
+          - "3.10.0"
+          - "3.11"
+          - "3.11.0"
+          - "3.12"
+          - "pypy3.8"
```
AlexWaygood, line 56:
```suggestion
```

.github/workflows/third_party.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "pypy3.8", "pypy3.9"]
```
AlexWaygood, line 44:
```suggestion
        python-version: ["3.8", "3.9", "3.10", "3.11", "pypy3.9"]
```

.github/workflows/third_party.yml
```diff
@@ -307,7 +308,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "pypy3.8"]
```
AlexWaygood, line 311:
```suggestion
        python-version: ["3.8", "3.9", "3.10", "3.11", "pypy3.9"]
```

AlexWaygood:
Our tests are also failing on pypy-3.8, it looks like, so I've also removed that from the matrix.

srittau (approved):

PR #267, opened 2023-07-06
Drop support for Python 3.7

AlexWaygood:


src/test_typing_extensions.py
```diff
@@ -5429,15 +5389,6 @@ def test_copy_and_pickle(self):
     def test_docstring(self):
         self.assertIsInstance(NamedTuple.__doc__, str)
 
-    @skipUnless(TYPING_3_8_0, "NamedTuple had a bad signature on <=3.7")
-    def test_signature_is_same_as_typing_NamedTuple(self):
-        self.assertEqual(inspect.signature(NamedTuple), inspect.signature(typing.NamedTuple))
```
AlexWaygood, line 5434: I deleted this test, as it no longer makes sense for us to have the same signature as `typing.NamedTuple` on <3.13, following #240

src/typing_extensions.py
```diff
@@ -535,46 +495,6 @@ def _get_protocol_attrs(cls):
     return attrs
 
 
-def _maybe_adjust_parameters(cls):
```
AlexWaygood, line 538: This function used to be used for our implementation of `TypedDict` and `Protocol`, but it is now only used for our implementation of `Protocol` on Python <3.8

src/typing_extensions.py
```diff
@@ -1077,7 +870,7 @@ def __subclasscheck__(cls, other):
     _TypedDict = type.__new__(_TypedDictMeta, 'TypedDict', (), {})
 
     @_ensure_subclassable(lambda bases: (_TypedDict,))
-    def TypedDict(__typename, __fields=_marker, *, total=True, **kwargs):
+    def TypedDict(typename, fields=_marker, /, *, total=True, **kwargs):
```
AlexWaygood, line 873: Strictly speaking, switching to using PEP-570 syntax here (and elsewhere) is a backwards-incompatible change. Not sure how we feel about that.

src/typing_extensions.py
```diff
@@ -1199,12 +992,12 @@ def greet(name: str) -> None:
         At runtime this returns the first argument unchanged and otherwise
         does nothing.
         """
-        return __val
+        return val
 
 
-if hasattr(typing, "Required"):
+if hasattr(typing, "Required"):  # 3.11+
```
AlexWaygood, line 998: These kinds of comments are very useful when doing this kind of PR, so I added them in various places that didn't have them

src/typing_extensions.py
```diff
@@ -2709,27 +2503,17 @@ class Employee(NamedTuple):
...
-    # On 3.8+, alter the signature so that it matches typing.NamedTuple.
-    # The signature of typing.NamedTuple on >=3.8 is invalid syntax in Python 3.7,
-    # so just leave the signature as it is on 3.7.
-    if sys.version_info >= (3, 8):
-        _new_signature = '(typename, fields=None, /, **kwargs)'
-        if isinstance(NamedTuple, _types.FunctionType):
-            NamedTuple.__text_signature__ = _new_signature
-        else:
-            NamedTuple.__call__.__text_signature__ = _new_signature
-
```
AlexWaygood, line 2732: As mentioned above, monkey-patching the signature so that it appears to be the same as `typing.NamedTuple` on all versions no longer makes sense following #240

src/typing_extensions.py
```diff
@@ -737,146 +648,33 @@ def _proto_hook(cls, other):
                 return NotImplemented
         return True
 
-    if sys.version_info >= (3, 8):
```
AlexWaygood, line 740: It makes me so happy to drop the 3.7 branch here :)

CONTRIBUTING.md
```diff
@@ -18,7 +18,7 @@ standard library, so that users can experiment with them before they are added t
 standard library. Such features should already be specified in a PEP or merged into
 CPython's `main` branch.
 
-`typing_extensions` supports Python versions 3.7 and up.
+`typing_extensions` supports Python versions 3.8 and up.
```
JelleZijlstra, line 21: Let's remove this line, it's redundant.

src/typing_extensions.py
JelleZijlstra: I'm fine with it.

src/typing_extensions.py
```diff
@@ -2323,14 +2116,14 @@ def reveal_type(__obj: T) -> T:
...
+        return obj
 
 
-if hasattr(typing, "assert_never"):
+if hasattr(typing, "assert_never"):  # 3.11+
     assert_never = typing.assert_never
-else:
-    def assert_never(__arg: Never) -> Never:
+else:  # <=3.10
+    def assert_never(arg: Never) -> Never:
```
JelleZijlstra, line 2126:
```suggestion
    def assert_never(arg: Never, /) -> Never:
```

JelleZijlstra (commented):
Thanks!

CONTRIBUTING.md
AlexWaygood: Should I delete the one in the README as well? It's probably fine us just stating this in `doc/index.rst`

JelleZijlstra: Yes, I think we can get rid of that one too. Thanks!

AlexWaygood: Okay, done!

PR #269, opened 2023-07-07
Fix third-party workflow

### LINKED ISSUE #268: Third-party tests failed on Fri Jul 07 2023

AlexWaygood:
pdm now won't let us add the `main` branch of typing_extensions as a dependency to pydantic or cattrs, because they both still have `requires-python = ">= 3.7"` in their pyproject.toml files, and we now have `requires-python = ">= 3.8"` in ours. To me it seems silly that pdm's locking mechanism is working this way, since we're not trying to actually install or run pydantic or cattrs on Python 3.7. But anyway, here's a hacky workaround to get things green again.

Fixes #268

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -49,6 +49,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: pydantic/pydantic
+      - name: Edit pydantic pyproject.toml
+        # pydantic's python-requires means pdm won't let us add typing-extensions-latest
+        # as a requirement unless we do this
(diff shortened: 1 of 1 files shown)


AlexWaygood:
> To me it seems silly that pdm's locking mechanism is working this way, since we're not trying to actually install or run pydantic or cattrs on Python 3.7.

Oh, I guess it's so that `pip install pydantic` is guaranteed to come with the same version of `typing_extensions` for any given version of pydantic, meaning that you get fully reproducible installs. I guess that makes sense. Still annoying for us, though.

PR #277, opened 2023-08-28
Add Doc from PEP 727: https://peps.python.org/pep-0727/

tiangolo:
Add doc and DocInfo from PEP 727: https://peps.python.org/pep-0727/

I'm not too familiar with the process in this codebase and what else would be expected, not sure if I'm missing something else, or if there's anything else I should test, order things in a different way, etc. 🤓

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -722,8 +722,8 @@ Functions
    .. versionadded:: 4.1.0
 
 
-Helper classes
-~~~~~~~~~~~~~~
+Annotation metadata
+~~~~~~~~~~~~~~~~~~~
 
 .. class:: Doc(documentation, /)
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -616,29 +616,6 @@ Functions
 
    .. versionadded:: 4.2.0
 
-.. function:: doc(documentation)
-
-   Define the documentation of a type annotation using :data:`Annotated`, to be
-   used in class attributes, function and method parameters, return values,
-   and variables.
-
-   The value should be a string literal to allow static tools like editors
-   to use it.
-
-   This complements docstrings.
-
-   It returns a :class:`DocInfo` instance containing the documentation value in
-   the ``documentation`` attribute.
-
-   Example::
-
-      >>> from typing_extensions import doc, Annotated
-      >>> def hi(to: Annotated[str, doc("Who to say hi to")]) -> None: ...
-
-   .. versionadded:: 4.8.0
-
-      See :pep:`727`.
-
 .. function:: get_args(tp)
 
    See :py:func:`typing.get_args`. In ``typing`` since 3.8.
@@ -748,15 +725,23 @@ Functions
 Helper classes
 ~~~~~~~~~~~~~~
 
-.. class:: DocInfo(documentation)
+.. class:: Doc(/, documentation)
 
-   Container for documentation information as returned by :func:`doc`.
+   Define the documentation of a type annotation using :data:`Annotated`, to be
+   used in class attributes, function and method parameters, return values,
+   and variables.
 
-   It is expected this class wouldn't be manually created but instead returned
-   by :func:`doc` inside of type annotations using :data:`Annotated`.
+   The value should be a positional-only string literal to allow static tools
+   like editors and documentation generators to use it.
 
-   The ``documentation`` attribute contains the documentation string passed
-   to :func:`doc`.
+   This complements docstrings.
+
+   The string value passed is available in the attribute ``documentation``.
+
+   Example::
+
+      >>> from typing_extensions import Annotated, Doc
+      >>> def hi(to: Annotated[str, Doc("Who to say hi to")]) -> None: ...
 
    .. versionadded:: 4.8.0
 
@@ -764,7 +749,7 @@ Helper classes
 
    .. attribute:: documentation
 
-      The documentation string passed to :func:`doc`.
+      The documentation string passed to :class:`Doc`.
 
 
 Pure aliases
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -616,6 +616,29 @@ Functions
 
    .. versionadded:: 4.2.0
 
+.. function:: doc(documentation)
+
+   Define the documentation of a type annotation using :data:`Annotated`, to be
+   used in class attributes, function and method parameters, return values,
+   and variables.
+
+   The value should be a string literal to allow static tools like editors
+   to use it.
+
+   This complements docstrings.
+
+   It returns a :class:`DocInfo` instance containing the documentation value in
+   the ``documentation`` attribute.
+
+   Example::
+
+      >>> from typing_extensions import doc, Annotated
+      >>> def hi(to: Annotated[str, doc("Who to say hi to")]) -> None: ...
+
+   .. versionadded:: 4.8.0
+
+      See :pep:`727`.
+
 .. function:: get_args(tp)
 
    See :py:func:`typing.get_args`. In ``typing`` since 3.8.
@@ -720,6 +743,29 @@ Functions
    See :py:func:`typing.reveal_type`. In ``typing`` since 3.11.
 
    .. versionadded:: 4.1.0
+
+
+Helper classes
+~~~~~~~~~~~~~~
+
+.. class:: DocInfo(documentation)
+
+   Container for documentation information as returned by :func:`doc`.
+
+   It is expected this class wouldn't be manually created but instead returned
+   by :func:`doc` inside of type annotations using :data:`Annotated`.
+
+   The ``documentation`` attribute contains the documentation string passed
+   to :func:`doc`.
+
+   .. versionadded:: 4.8.0
+
+      See :pep:`727`.
+
+   .. attribute:: documentation
+
+      The documentation string passed to :func:`doc`.
+
 
 Pure aliases
 ~~~~~~~~~~~~
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -725,7 +725,7 @@ Functions
 Helper classes
 ~~~~~~~~~~~~~~
 
-.. class:: Doc(/, documentation)
+.. class:: Doc(documentation, /)
 
    Define the documentation of a type annotation using :data:`Annotated`, to be
    used in class attributes, function and method parameters, return values,
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -60,8 +60,7 @@ __all__ = [
     'clear_overloads',
     'dataclass_transform',
     'deprecated',
-    'doc',
-    'DocInfo',
+    'Doc',
     'get_overloads',
     'final',
     'get_args',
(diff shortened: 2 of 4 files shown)


tiangolo:
Sorry for the lint errors. 🤦 

After the first one, I ran the command locally that exposed the error in tests but I didn't realize there was a different one to expose the error in the source. I hope I got them all now. 😅

tiangolo:
BTW, if you want and/or is easier, feel free to discard this PR and make a new one, or take over and commit on top, etc. It's all good. 😄

src/typing_extensions.py
```diff
@@ -2809,6 +2811,44 @@ def get_protocol_members(tp: type, /) -> typing.FrozenSet[str]:
...
+    class DocInfo:
+        """Container for documentation information as returned by ``doc()``.
+
+        It is expected this class wouldn't be manually created but instead returned
+        by ``doc()`` inside of type annotations using ``Annotated``.
+
+        The ``documentation`` attribute contains the documentation string passed
+        to ``doc()``.
+        """
+        def __init__(self, documentation: str):
```
JelleZijlstra, line 2827:
```suggestion
        def __init__(self, documentation: str) -> None:
```

src/typing_extensions.py
```diff
@@ -2809,6 +2811,44 @@ def get_protocol_members(tp: type, /) -> typing.FrozenSet[str]:
...
+            self.documentation = documentation
+
+        def __repr__(self) -> str:
+            return f"DocInfo({self.documentation!r})"
```
JelleZijlstra, line 2831:
```suggestion
            return f"DocInfo({self.documentation!r})"
       
       def __hash__(self) -> int:
            return hash(self.documentation)
```
Let's make it hashable (hope I got the indentation right)

src/typing_extensions.py
```diff
@@ -2809,6 +2811,44 @@ def get_protocol_members(tp: type, /) -> typing.FrozenSet[str]:
...
+
+        It returns a class ``DocInfo`` instance containing the documentation value in
+        the ``documentation`` attribute.
+
+        Example::
+
+            >>> from typing_extensions import doc, Annotated
+            >>> def hi(to: Annotated[str, doc("Who to say hi to")]) -> None: ...
+        """
+        return DocInfo(documentation)
```
JelleZijlstra, line 2851: Just mentioned this on Discuss, but it feels a little redundant to have this function when it doesn't do anything except construct the class. I don't feel too strongly about exactly what color to paint the shed, but obvious alternatives include:

- Make `doc` itself a class, getting rid of `DocInfo` (disadvantage: weird to have a class with a lowercase name)
- Get rid of `doc` and have users call `DocInfo` directly (disadvantage: even more verbose)
- Rename it to `Doc` and have users write `Annotated[param, Doc("my doc")]` (disadvantage: the lowercase `doc` looks nicer to my eye)

I'm sure people on Discuss will come up with other similar ideas. Ultimately it's your call as the PEP author which one of those you prefer.

My main concern here is backward compat for typing-extensions itself: if I merge and release this PR, code that uses `typing_extensions.DocInfo` and `typing_extensions.doc` in a sensible way should keep working essentially in perpetuity. (Because typing-extensions is a low-level dependency for so much of the ecosystem, backward compatibility is very important.) So I want to make sure that whatever we release will keep working even if the PEP's naming scheme changes.

Concretely, the following the operations should keep working, even if we make another typing-extensions release that spells the proposal differently:
- Using `typing_extensions.doc()` to create an instance of the info class
- Using `isinstance(..., typing_extensions.DocInfo)` to check for documentation
- Using `typing_extensions.DocInfo("some doc")` to construct docs (this is not exactly documented but I'm sure people will do it)

I think we can keep these things working with any of my ideas from above, simply by maintaining aliases like `DocInfo = Doc` in a future version of typing-extensions. The one problem is that if only one of `DocInfo` and `doc` ends up in `typing`, then the `if hasattr` above won't work properly, so I think we should check for both separately.

JelleZijlstra (commented):
This should also be documented in the changelog and the documentation page. I'll push some changes in a moment.

doc/index.rst
```diff
@@ -721,6 +744,29 @@ Functions
 
    .. versionadded:: 4.1.0
 
+
+Helper classes
```
JelleZijlstra, line 748: Wasn't sure where to put this so I made a new section. @AlexWaygood do you have any suggestions?

JelleZijlstra:
FYI I pushed documentation, a few more dunder methods, and more tests. (I don't know why people like pickling typing objects but they do, so let's test that it works.)

src/typing_extensions.py
Gobot1234: FWIW this very much looks to me similar to property in this regard and PEP 8 is fine with classes that look like functions if they are meant to be called like functions.  (I did read the PEP and think it was an odd choice to have a function that does basically nothing)

AlexWaygood: I agree that it seems needlessly duplicative to have a `doc` function and a `DocInfo` class. My personal preference would just be to have a `Doc` class; `Annotated[int, Doc('the number of apples')]` doesn't look too bad to me, and I think we'd quickly get used to it. I'd also be fine if it were a `doc` class, though; I agree with @Gobot1234 that `property`, `classmethod` and `enumerate` are all good precedents here for "function-like" classes that have lowercase names

doc/index.rst
AlexWaygood: In a way, it's a similar class to `ParamSpecArgs` and `ParamSpecKwargs` -- all three are simple data classes that have no special meaning to a type checker, and just exist to hold information

AlexWaygood:
> I don't know why people like pickling typing objects

though not as much as people [like making weakrefs to typing objects](https://github.com/python/cpython/issues/108295), apparently ;)

src/typing_extensions.py
```diff
@@ -2810,6 +2812,56 @@ def get_protocol_members(tp: type, /) -> typing.FrozenSet[str]:
...
+        to ``doc()``.
+        """
+        def __init__(self, documentation: str) -> None:
+            self.documentation = documentation
+
+        def __repr__(self) -> str:
+            return f"DocInfo({self.documentation!r})"
+
+        def __hash__(self) -> int:
+            return hash(self.documentation)
```
AlexWaygood, line 2834: I believe it's generally discouraged to have classes where this happens:

```pycon
>>> hash(DocInfo('foo')) == hash('foo')
True
>>> DocInfo('foo') == 'foo'
False
```

doc/index.rst
AlexWaygood: "Metadata classes"?

src/typing_extensions.py
tiangolo: Agreed! And thanks for the feedback regarding this in the discussion as well. I just removed the function and made it a single `Doc` class. I also made it take a positional-only parameter, taking feedback from Eric Traut.

I will update the PEP accordingly later, after finishing with this PR https://github.com/python/peps/pull/3316

src/typing_extensions.py
```diff
@@ -2810,6 +2811,41 @@ def get_protocol_members(tp: type, /) -> typing.FrozenSet[str]:
...
+        Example::
+
+            >>> from typing_extensions import Annotated, Doc
+            >>> def hi(to: Annotated[str, Doc("Who to say hi to")]) -> None: ...
+        """
+        def __init__(self, /, documentation: str) -> None:
+            self.documentation = documentation
+
+        def __repr__(self) -> str:
+            return f"DocInfo({self.documentation!r})"
```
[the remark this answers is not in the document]
tiangolo: Oh, right!

doc/index.rst
```diff
@@ -721,6 +744,29 @@ Functions
...
+   The ``documentation`` attribute contains the documentation string passed
+   to :func:`doc`.
+
+   .. versionadded:: 4.8.0
+
+      See :pep:`727`.
+
+   .. attribute:: documentation
+
+      The documentation string passed to :func:`doc`.
```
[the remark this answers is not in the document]
tiangolo: I'm on it.

Is the docstring of the class itself okay? Or should I use

```
:data:`Annotated`
```

instead of:

```
`Annotated`
```
?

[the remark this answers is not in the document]
AlexWaygood: ReST markup is preferred here, so that Sphinx can add a link to the docs for `Annotated`. But it doesn't make sense to include ReST markup in the docstring, since we don't autogenerate docs from the docstring 👍

doc/index.rst
```diff
@@ -721,6 +721,37 @@ Functions
 
    .. versionadded:: 4.1.0
 
+
+Helper classes
+~~~~~~~~~~~~~~
+
+.. class:: Doc(/, documentation)
```
tiangolo, line 728: I'm not sure this is the right syntax in Sphinx to show that it's a positional-only parameter, I assumed so infering from `dataclass_transform(*, eq_default=False, etc=True)`.

src/typing_extensions.py
```diff
@@ -2810,6 +2811,41 @@ def get_protocol_members(tp: type, /) -> typing.FrozenSet[str]:
...
+        This complements docstrings.
+
+        The string value passed is available in the attribute ``documentation``.
+
+        Example::
+
+            >>> from typing_extensions import Annotated, Doc
+            >>> def hi(to: Annotated[str, Doc("Who to say hi to")]) -> None: ...
+        """
+        def __init__(self, /, documentation: str) -> None:
```
AlexWaygood, line 2834: In the docs you say that this should be positional-only, which means the slash has to come after the argument:

```suggestion
        def __init__(self, documentation: str, /) -> None:
```

doc/index.rst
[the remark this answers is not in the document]
tiangolo: Ah, perfect. Thank you.

src/typing_extensions.py
tiangolo: 🤦

doc/index.rst
```diff
@@ -721,6 +721,37 @@ Functions
 
    .. versionadded:: 4.1.0
 
+
+Helper classes
+~~~~~~~~~~~~~~
+
+.. class:: Doc(documentation, /)
```
tiangolo, line 728: I'm not sure this is the right syntax in Sphinx to show that it's a positional-only parameter, I assumed so inferring from `dataclass_transform(*, eq_default=False, etc=True)`.

doc/index.rst
tiangolo: This was already wrong, just updated it and added the comment again in the new version.

doc/index.rst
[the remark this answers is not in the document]
AlexWaygood: No problem, thanks for the update!

doc/index.rst
JelleZijlstra: Now that "Doc" is a class intended to be use by normal (non-introspection) users, I don't know if "Helper classes" or "Metadata classes" still makes sense. Maybe we should say "Annotation metadata"?

JelleZijlstra (approved):
Looks good! (Some concerns about the name of the doc section, but that's something we can always change later.)

doc/index.rst
hugovk: Tip: enable docs previews for PRs:

https://docs.readthedocs.io/en/stable/pull-requests.html

doc/index.rst
AlexWaygood: "Annotation metadata" sounds reasonable to me

JelleZijlstra (approved):

tiangolo:
We haven't discussed this anywhere yet, but I just realizing (right after you merged this one 🤦), b now that there's only one class `Doc` there's probably not a strong argument for having the attribute named `documentation`, do you think it would be better to name it `value`?

The main argument I can think of for that is that it's shorter and would be somewhat similar to an enum where the actual value is stored in an attribute `value`. What do you think?

JelleZijlstra:
I don't feel too strongly. I think the precise name is nice to make it clear what you are working with (e.g., you can grep for `\.documentation` more easily than for `\.value`), and there's some chance we'll add other attributes to the `Doc` class in the future.

(Just for backward compatibility in typing-extensions, if we do release with `.documentation` and decide to rename it later, we can preserve compatibility by adding a `@property` delegating from `.documentation` to the new name.)

tiangolo:
Thank you for the quick feedback! Perfect. 🤓🚀

PR #278, opened 2023-08-30
Docs improvements

JelleZijlstra:
- Move OrderedDict into pure aliases as we dropped support for 3.7
- Explicitly state there is no BC before a final release. Hopefully this is
  obvious, but just in case we get people complaining about this.


hauntsaninja (approved):

AlexWaygood (commented):
LGTM!

PR #279, opened 2023-09-06
Fix ParamSpec ellipsis default for <3.10

Gobot1234:


AlexWaygood (commented):
CHANGELOG entry/docs? :-)

If this fixes an issue you hit, other people might have hit it too, so they probably deserve to know about the fix!

JelleZijlstra:
The change itself looks good, but please add a changelog/docs entry as Alex requested. For reference, PEP 696 explicitly allows this: https://peps.python.org/pep-0696/#paramspec-defaults

Gobot1234:
Oh weird, it doesn't seem to have pushed the changes

JelleZijlstra (approved):
Looks good but @AlexWaygood please take a look too.

I'll likely cut a new release once the PEP 727 (`Doc()`) changes are in.

doc/index.rst
```diff
@@ -278,6 +278,11 @@ Special typing primitives
 
       The implementation was changed for compatibility with Python 3.12.
 
+   .. versionchanged:: 4.8.0
+
+      Fixed a bug where passing an ellipsis literal to ``default`` would raise a 
+      :py:exc:`TypeError` on Python versions <3.11.
+
```
AlexWaygood, line 285: Over at CPython we usually _wouldn't_ advertise pure bugfixes in the main docs (only in the changelog -- in the main docs we generally pretend the implementation has been perfect all along ;) but I'm not sure what our policy is at `typing_extensions`.

(The line between bugfixes and features is also slightly blurrier in `typing_extensions`, since "backporting a bugfix from CPython" can count as a "`typing_extensions` feature". But the feature we're fixing a bug in here hasn't been implemented in CPython yet, of course.)

JelleZijlstra: I'm OK with adding it, but would reword it as "Passing an ellipsis literal (`...`) to `default` now works on Python 3.10 and lower."

src/typing_extensions.py
```diff
@@ -1277,7 +1277,10 @@ def _set_default(type_param, default):
         type_param.__default__ = tuple((typing._type_check(d, "Default must be a type")
                                         for d in default))
     elif default != _marker:
-        type_param.__default__ = typing._type_check(default, "Default must be a type")
+        if isinstance(type_param, ParamSpec) and default is ...:  # ... not valid <3.11
+            type_param.__default__ = default
+        else:
+            type_param.__default__ = typing._type_check(default, "Default must be a type")
```
AlexWaygood, line 1283: Aha, so the issue is that `typing._type_check` changed in Python 3.11 so that it would accept an ellipsis, but didn't allow it on previous versions

AlexWaygood (approved):
Looks reasonable to me, other than my point about the docs.

Another possible fix could be to vendor the current version of `_type_check` from the CPython `main` branch, since we use that function quite a lot in `typing_extensions`, and there might have been other changes to `_type_check` in recent versions of CPython that are causing bugs we don't know about yet in `typing_extensions`. It would be a more invasive change, though; I think this fix is fine for now

doc/index.rst
```diff
@@ -278,6 +278,11 @@ Special typing primitives
 
       The implementation was changed for compatibility with Python 3.12.
 
+   .. versionchanged:: 4.8.0
+
+      Fixed a bug where passing an ellipsis literal (``...``) to ``default`` now
+      works on Python 3.10 and lower.
```
AlexWaygood, line 284:
```suggestion
      Passing an ellipsis literal (``...``) to *default* now works on Python
      3.10 and lower.
```

Gobot1234:
> Looks reasonable to me, other than my point about the docs.
> 
> Another possible fix could be to vendor the current version of `_type_check` from the CPython `main` branch, since we use that function quite a lot in `typing_extensions`, and there might have been other changes to `_type_check` in recent versions of CPython that are causing bugs we don't know about yet in `typing_extensions`. It would be a more invasive change, though; I think this fix is fine for now

I think its this change FWIW https://github.com/python/cpython/commit/870b22b9c442d035190d2b8fb82256cd9a03da48

AlexWaygood:
> I think its this change FWIW [python/cpython@870b22b](https://github.com/python/cpython/commit/870b22b9c442d035190d2b8fb82256cd9a03da48)

Riiight... so we probably can't use `typing_extensions.Annotated` to wrap `typing_extensions.ParamSpecArgs` on lower versions of Python... probably not the biggest problem in the world...

AlexWaygood (approved):
thanks, lgtm!

JelleZijlstra (approved):

AlexWaygood:
> > I think its this change FWIW [python/cpython@870b22b](https://github.com/python/cpython/commit/870b22b9c442d035190d2b8fb82256cd9a03da48)
> 
> Riiight... so we probably can't use `typing_extensions.Annotated` to wrap `typing_extensions.ParamSpecArgs` on lower versions of Python... probably not the biggest problem in the world... :)

lol, confirmed:

```pycon
Python 3.9.12 (main, Apr  4 2022, 05:22:27) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> from typing_extensions import *
>>> P = ParamSpec("P")
>>> Annotated[P.args, 'foo']
  File "C:\Users\alexw\.conda\envs\py39\lib\typing.py", line 277, in inner
    return func(*args, **kwds)
  File "C:\Users\alexw\.conda\envs\py39\lib\typing.py", line 1340, in __class_getitem__
    origin = _type_check(params[0], msg, allow_special_forms=True)
  File "C:\Users\alexw\.conda\envs\py39\lib\typing.py", line 166, in _type_check
    raise TypeError(f"{msg} Got {arg!r:.100}.")
TypeError: Annotated[t, ...]: t must be a type. Got P.args.
```

Works fine with `typing` on later Python versions:

```pycon
Python 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)] on win32
>>> from typing import *
>>> P = ParamSpec("P")
>>> Annotated[P.args, 'foo']
typing.Annotated[P.args, 'foo']
```

JelleZijlstra:
We probably should vendor `_type_check` unconditionally. It's not a particularly likely thing to run into, but it is technically a bug in typing-extensions. Do you want to make a PR?

AlexWaygood:
Sure

AlexWaygood:
> We probably should vendor `_type_check` unconditionally.

hmm... seems tricky to just copy over the py311+ implementation without also backporting `_type_convert()` from py311+, and that in turn requires backporting `ForwardRef` as it exists on py311+.

JelleZijlstra:
Thanks for investigating! That feels like it might be more trouble than a purely hypothetical use case is worth. Adding our own ForwardRef would potentially break users who do `isinstance(..., typing.ForwardRef)`. (They should technically be checking for `typing_extensions.ForwardRef` anyway, but we don't have to break them unnecessarily.)

PR #281, opened 2023-09-08
Run tests on py312 for more third-party projects

AlexWaygood:
~~cattrs,~~ mypy and pyanalyze all now support Python 3.12, and run tests on Python 3.12 in CI.

Also, get rid of a hack that's no longer necessary for running the cattrs tests, since cattrs now requires Python 3.8+

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -128,7 +128,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
There's something wrong with `cattrs`, apparently while trying to install the `pendulum` dependency but the error isn't very helpful.

AlexWaygood:
> There's something wrong with `cattrs`, apparently while trying to install the `pendulum` dependency but the error isn't very helpful.

It's odd: the cattrs test suite appears to be passing on py312 upstream: https://github.com/python-attrs/cattrs/actions/runs/6043137162/job/16399595172. But they don't appear to be installing pendulum upstream; not 100% sure why we are with our invocation of the test suite.

AlexWaygood:
I've backed out the change for cattrs; we can leave it for now

PR #284, opened 2023-09-18
Add support for PEP 705

JelleZijlstra:


JelleZijlstra:
Let's not merge this until python/peps#3440 is merged.

doc/index.rst
```diff
@@ -366,6 +372,45 @@ Special typing primitives
...
+
+   .. attribute:: __other_keys__
+
+      Boolean indicating the value of the ``other_keys=`` argument for
+      the current class, which indicates whether the ``TypedDict`` accepts
+      keys other than those explicitly declared in the class body.
+
+      Note that if a base class has ``other_keys=False``,
+      but the current class does not, semantically the class will not accept
+      other keys, but this attribute will still be ``False``.
```
srittau, line 410: I don't understand this sentence. So,

```python
class Super(TypedDict, other_keys=False): ...
class Sub(TypedDict, other_keys=True): ...
```

Does that mean that `Sub` semantically(?) doesn't allow other keys (but practically it does?) and `Sub.__other_keys__` is `False`?

JelleZijlstra: Your understanding is correct, but I agree this would be confusing. I'll leave the decision to @alicederyn as the PEP author, but I'd recommend one of these two:

- Make it an error for an `other_keys=True` TypedDict to inherit from an `other_keys=False` TypedDict. This would mean that any subclass of an `other_keys=False` TypedDict must also set `other_keys=False`. This is my preference as it's most explicit.
- Automatically set `other_keys=False` if any base class has `other_keys=False`.

doc/index.rst
```diff
@@ -318,6 +318,12 @@ Special typing primitives
       present in a protocol class's :py:term:`method resolution order`. See
       :issue:`245` for some examples.
 
+.. data:: ReadOnly
+
+   See :pep:`705`. Indicates that a :class:`TypedDict` key may not be modified.
```
alicederyn, line 323: I've been trying to stick to saying that the "entry" is readonly, rather than "key", to avoid possible confusion.

doc/index.rst
alicederyn: Good point. I've also gone for explicit, and updated PEP-705.

src/typing_extensions.py
```diff
@@ -829,35 +853,73 @@ def __new__(cls, name, bases, ns, total=True):
...
+                        )
+                if (
+                    readonly
+                    and is_typeddict(base)
+                    and base is not _TypedDict
+                    and not getattr(base, "__readonly__", False)
+                ):
+                    raise TypeError(
+                        "read-only TypedDict cannot extend non-read-only "
+                        f"TypedDict {base}"
```
AlexWaygood, line 883:
```suggestion
                        f"TypedDict {base.__name__!r}"
```

(we know that `base` is an instance of `type` at this point in the function, so the `.__name__` attribute access should be safe)

src/typing_extensions.py
```diff
@@ -829,35 +853,73 @@ def __new__(cls, name, bases, ns, total=True):
...
             for annotation_key, annotation_type in own_annotations.items():
-                annotation_origin = get_origin(annotation_type)
-                if annotation_origin is Annotated:
-                    annotation_args = get_args(annotation_type)
-                    if annotation_args:
-                        annotation_type = annotation_args[0]
-                        annotation_origin = get_origin(annotation_type)
-
-                if annotation_origin is Required:
+                qualifiers = list(_get_typeddict_qualifiers(annotation_type))
```
AlexWaygood, line 888: Very minor, but a `set` might be better here (it would make the `if Required in qualifiers` check below O(1) )

```suggestion
                qualifiers = set(_get_typeddict_qualifiers(annotation_type))
```

src/typing_extensions.py
```diff
@@ -829,35 +853,73 @@ def __new__(cls, name, bases, ns, total=True):
...
                 annotations.update(base.__dict__.get('__annotations__', {}))
                 required_keys.update(base.__dict__.get('__required_keys__', ()))
                 optional_keys.update(base.__dict__.get('__optional_keys__', ()))
+                readonly_keys.update(base.__dict__.get('__readonly_keys__', ()))
+                mutable_keys.update(base.__dict__.get('__mutable_keys__', ()))
+
+                if not getattr(base, "__other_keys__", True):
+                    if (
+                        set(own_annotations) -
+                        set(base.__dict__.get('__annotations__', {}))
```
AlexWaygood, line 869: runs about twice as fast:

```suggestion
                        own_annotations.keys() -
                        base.__dict__.get('__annotations__', {}).keys()
```

doc/index.rst
alicederyn: The PEP got rewritten to use "item" instead of "entry", as that's what the Python docs/API use elsewhere.

src/typing_extensions.py
```diff
@@ -829,35 +853,73 @@ def __new__(cls, name, bases, ns, total=True):
...
+                    if annotation_key in mutable_keys:
+                        raise TypeError(
+                            f"Cannot override mutable key {annotation_key!r}"
+                            " with read-only key"
+                        )
+                    readonly_keys.add(annotation_key)
+                elif readonly:
+                    readonly_keys.add(annotation_key)
+                else:
+                    mutable_keys.add(annotation_key)
```
alicederyn, line 912: Do you need to remove the key from readonly_keys too here? It might be inherited.

alicederyn:
PEP-705 has dropped both `readonly` and `other_keys`; only the `typing.ReadOnly` annotation needs to be added now.

doc/index.rst
JelleZijlstra: No longer relevant as `other_keys` was removed

src/typing_extensions.py
JelleZijlstra: This code is gone

src/typing_extensions.py
JelleZijlstra: Done

src/typing_extensions.py
JelleZijlstra: Removed this code

src/typing_extensions.py
JelleZijlstra: This is actually already buggy for Required/NotRequired, even in CPython:

```
Python 3.13.0a2+ (heads/pep702:6ab4584ac4, Nov 26 2023, 17:26:43) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> from typing import *
>>> class A(TypedDict):
...     a: NotRequired[int]
... 
>>> class B(A):
...     a: Required[int]
... 
>>> B.__required_keys__
frozenset({'a'})
>>> B.__optional_keys__
frozenset({'a'})
```

I'll leave fixing that to a separate PR.

JelleZijlstra: (But I'll fix this for ReadOnly.)

alicederyn: I think that example is illegal though. You can't change the required/notrequired on a key like that; it's banned by PEP-589. It's only once we have ReadOnly that this becomes an issue.

src/typing_extensions.py
```diff
@@ -1925,6 +1961,51 @@ class Movie(TypedDict):
...
+
+            class Movie(TypedDict):
+                title: ReadOnly[str]
+                year: int
+
+            def mutate_movie(m: Movie) -> None:
+                m["year"] = 1992  # allowed
+                m["title"] = "The Matrix"  # typechecker error
+
+        There is no runtime checking for this propery.
```
alicederyn, line 1980:
```suggestion
        There is no runtime checking for this property.
```

src/typing_extensions.py
```diff
@@ -1925,6 +1961,51 @@ class Movie(TypedDict):
         """)
 
 
+if hasattr(typing, 'ReadOnly'):
+    ReadOnly = typing.ReadOnly
+elif sys.version_info[:2] >= (3, 9):  # 3.9-3.12
+    @_ExtensionsSpecialForm
+    def ReadOnly(self, parameters):
+        """A special typing construct to mark a key of a TypedDict as read-only.
```
alicederyn, line 1969:
```suggestion
        """A special typing construct to mark an item of a TypedDict as read-only.
```
If we want to keep to the naming approach of the PEP

src/typing_extensions.py
JelleZijlstra: Replied about this on CPython

src/typing_extensions.py
```diff
@@ -1925,6 +1961,51 @@ class Movie(TypedDict):
         """)
 
 
+if hasattr(typing, 'ReadOnly'):
+    ReadOnly = typing.ReadOnly
+elif sys.version_info[:2] >= (3, 9):  # 3.9-3.12
+    @_ExtensionsSpecialForm
+    def ReadOnly(self, parameters):
+        """A special typing construct to mark an item of a TypedDict as read-only.
+        For example:
```
AlexWaygood, line 1970:
```suggestion

        For example:
```

src/typing_extensions.py
```diff
@@ -1925,6 +1961,51 @@ class Movie(TypedDict):
...
+    class _ReadOnlyForm(_ExtensionsSpecialForm, _root=True):
+        def __getitem__(self, parameters):
+            item = typing._type_check(parameters,
+                                      f'{self._name} accepts only a single type.')
+            return typing._GenericAlias(self, (item,))
+
+    ReadOnly = _ReadOnlyForm(
+        'ReadOnly',
+        doc="""A special typing construct to mark a key of a TypedDict as read-only.
+        For example:
```
AlexWaygood, line 1995:
```suggestion

        For example:
```

AlexWaygood (approved):
Some minor suggestions (mostly optional), but overall this looks good!

AlexWaygood (approved):
LGTM

PR #288, opened 2023-09-28
Make `NewType.__call__` params positional-only

AlexWaygood:
This is really minor, but it means that the signature of `typing_extensions.NewType.__call__` exactly matches that of `typing.NewType.__call__` on all Python versions we support:

```pycon
Python 3.8.16 (default, Mar  2 2023, 03:18:16) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> from typing import NewType
>>> x = NewType("x", int)
>>> x(obj=42)
TypeError: new_type() got an unexpected keyword argument 'obj'
```

```pycon
Python 3.10.8 | packaged by conda-forge | (main, Nov 24 2022, 14:07:00) [MSC v.1916 64 bit (AMD64)] on win32
>>> x = NewType("x", int)
NameError: name 'NewType' is not defined
>>> from typing import NewType
>>> x = NewType("x", int)
>>> x(obj=42)
TypeError: NewType.__call__() got an unexpected keyword argument 'obj'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Release 4.9.0
+# Release 4.9.0 (???)
 
 - All parameters on `NewType.__call__` are now positional-only. This means that
   the signature of `typing_extensions.NewType.__call__` now exactly matches the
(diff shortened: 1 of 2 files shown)


AlexWaygood:
(I don't think this needs a changelog entry, but happy to add one if folks disagree!)

JelleZijlstra:
Thanks! We should document this in the changelog as it is a user-visible change, but no need to mention it in the docs.

PR #291, opened 2023-10-18
Fix tests on Python 3.13

hroncok:
This is a followup for d95cc228ea96feec105592a9902e5b2d6cc048a9

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -3383,11 +3383,6 @@ class TypedDictTests(BaseTestCase):
         with self.assertRaises(TypeError):
             TypedDict('Emp', [('name', str)], None)
 
-        with self.assertWarns(DeprecationWarning):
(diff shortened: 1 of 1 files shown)


hroncok:
Alternatively, this can be split into a skipped test. Let me know if you prefer that.

hroncok:
The if-ed block seems to be already tested in `test_typeddict_special_keyword_names` so I am not sure what was the purpose :confused:

hroncok:
The `test_keywords_syntax_raises_on_3_13` test already asserts that type error.

JelleZijlstra:
You're right. Let's just remove this block of tests, it duplicates `test_basics_keywords_syntax` exactly.

hroncok:
Done. Explained the reasons for removal in the commit message.

JelleZijlstra (approved):

hroncok:
Thanks for the quick feedback loop :)

PR #293, opened 2023-10-27
Fix runtime behaviour of PEP 696

NCPlayz:
In [PEP 696](https://peps.python.org/pep-0696/) there are two failure modes specified:

```py
DefaultStrT = TypeVar("DefaultStrT", default=str)
DefaultIntT = TypeVar("DefaultIntT", default=int)
DefaultBoolT = TypeVar("DefaultBoolT", default=bool)
T = TypeVar("T")
T2 = TypeVar("T2")

 # 1. Invalid: non-default TypeVars cannot follow ones with defaults
class NonDefaultFollowsDefault(Generic[DefaultStrT, T]): ...

class AllTheDefaults(Generic[T1, T2, DefaultStrT, DefaultIntT, DefaultBoolT]): ...

# 2. Invalid: expected 2 arguments to AllTheDefaults
AllTheDefaults[int]
```

This PR aims to implement the runtime behaviour for this PEP that was missed out in the initial support PR #77. 
It also brings support for this to Python 3.11+.

One note I had was that the current error will say `Too few parameters for AllTheDefaults; actual 1, expected 5`.
For TypeVarLikes with defaults they aren't actually required, so is it worth computing the minimum required then saying `expected at least 2` for this case?

Let me know if you have any suggestions!

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5605,6 +5605,23 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         class A(Generic[Unpack[Ts]]): ...
         Alias = Optional[Unpack[Ts]]
 
+    def test_erroneous_generic(self):
+        DefaultStrT = TypeVar('DefaultStrT', default=str)
+        T = TypeVar('T')
+
+        with self.assertRaises(TypeError):
+            Test = Generic[DefaultStrT, T]
+
+    def test_need_more_params(self):
+        DefaultStrT = typing_extensions.TypeVar('DefaultStrT', default=str)
+        T = typing_extensions.TypeVar('T')
+        U = typing_extensions.TypeVar('U')
+
+        class A(Generic[T, U, DefaultStrT]): ...
+
+        with self.assertRaises(TypeError):
+            Test = A[int]
+
     def test_pickle(self):
         global U, U_co, U_contra, U_default  # pickle wants to reference the class by name
         U = typing_extensions.TypeVar('U')
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -3262,10 +3262,7 @@ class ProtocolTests(BaseTestCase):
         self.assertEqual(MemoizedFunc.__parameters__, (Ts, T, T2))
         self.assertTrue(MemoizedFunc._is_protocol)
 
-        # 3.11+ moves where this exception is thrown
-        # 3.11: TypeVarTuple.__typing_prepare_subst__
-        # 3.12+: typing._typevartuple_prepare_subst
-        things = "arguments" if sys.version_info >= (3, 11) else "parameters"
+        things = "arguments"
 
         # A bug was fixed in 3.11.1
         # (https://github.com/python/cpython/commit/74920aa27d0c57443dd7f704d6272cca9c507ab3)
@@ -5714,7 +5711,7 @@ class NamedTupleTests(BaseTestCase):
                 self.assertIsInstance(a, G)
                 self.assertEqual(a.x, 3)
 
-                with self.assertRaisesRegex(TypeError, 'Too many parameters'):
+                with self.assertRaisesRegex(TypeError, 'Too many arguments'):
                     G[int, str]
 
     @skipUnless(TYPING_3_9_0, "tuple.__class_getitem__ was added in 3.9")
@@ -6229,8 +6226,12 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         U = typing_extensions.TypeVar('U')
 
         class A(Generic[T, U, DefaultStrT]): ...
+        A[int, bool]
+        A[int, bool, str]
 
-        with self.assertRaises(TypeError):
+        with self.assertRaises(
+            TypeError, "Too few arguments for .+; actual 1, expected at least 2"
+        ):
             Test = A[int]
 
     def test_pickle(self):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -3262,14 +3262,12 @@ class ProtocolTests(BaseTestCase):
         self.assertEqual(MemoizedFunc.__parameters__, (Ts, T, T2))
         self.assertTrue(MemoizedFunc._is_protocol)
 
-        things = "arguments"
-
         # A bug was fixed in 3.11.1
         # (https://github.com/python/cpython/commit/74920aa27d0c57443dd7f704d6272cca9c507ab3)
         # That means this assertion doesn't pass on 3.11.0,
         # but it passes on all other Python versions
         if sys.version_info[:3] != (3, 11, 0):
-            with self.assertRaisesRegex(TypeError, f"Too few {things}"):
+            with self.assertRaisesRegex(TypeError, "Too few arguments"):
                 MemoizedFunc[int]
 
         X = MemoizedFunc[int, T, T2]
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -6230,7 +6230,7 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         A[int, bool, str]
 
         with self.assertRaises(
-            TypeError, "Too few arguments for .+; actual 1, expected at least 2"
+            TypeError, msg="Too few arguments for .+; actual 1, expected at least 2"
         ):
             Test = A[int]
 
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5714,7 +5714,7 @@ class NamedTupleTests(BaseTestCase):
                 self.assertIsInstance(a, G)
                 self.assertEqual(a.x, 3)
 
-                with self.assertRaisesRegex(TypeError, f'Too many parameters'):
+                with self.assertRaisesRegex(TypeError, 'Too many parameters'):
                     G[int, str]
 
     @skipUnless(TYPING_3_9_0, "tuple.__class_getitem__ was added in 3.9")
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -3096,6 +3096,9 @@ class ProtocolTests(BaseTestCase):
         self.assertEqual(MemoizedFunc.__parameters__, (Ts, T, T2))
         self.assertTrue(MemoizedFunc._is_protocol)
 
+        # 3.11+ moves where this exception is thrown
+        # 3.11: TypeVarTuple.__typing_prepare_subst__
+        # 3.12+: typing._typevartuple_prepare_subst
         things = "arguments" if sys.version_info >= (3, 11) else "parameters"
 
         # A bug was fixed in 3.11.1
@@ -5231,9 +5234,7 @@ class NamedTupleTests(BaseTestCase):
                 self.assertIsInstance(a, G)
                 self.assertEqual(a.x, 3)
 
-                things = "arguments" if sys.version_info >= (3, 11) else "parameters"
-
-                with self.assertRaisesRegex(TypeError, f'Too many {things}'):
+                with self.assertRaisesRegex(TypeError, f'Too many parameters'):
                     G[int, str]
 
     @skipUnless(TYPING_3_9_0, "tuple.__class_getitem__ was added in 3.9")
@@ -5606,7 +5607,7 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         Alias = Optional[Unpack[Ts]]
 
     def test_erroneous_generic(self):
-        DefaultStrT = TypeVar('DefaultStrT', default=str)
+        DefaultStrT = typing_extensions.TypeVar('DefaultStrT', default=str)
         T = TypeVar('T')
 
         with self.assertRaises(TypeError):
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2700,9 +2700,6 @@ else:
             expect_val = elen
             if hasattr(cls, "__parameters__"):
                 parameters = [p for p in cls.__parameters__ if not _is_unpack(p)]
-                num_tv_tuples = sum(isinstance(p, TypeVarTuple) for p in parameters)
-                if (num_tv_tuples > 0) and (alen >= elen - num_tv_tuples):
-                    return
 
                 # deal with TypeVarLike defaults
                 # required TypeVarLikes cannot appear after a defaulted one.
@@ -2788,9 +2785,9 @@ else:
 
                     parameters.append(t)
             else:
-                if _should_collect_from_parameters(t):
-                    parameters.extend(
-                        [t for t in t.__parameters__ if t not in parameters])
+                for x in getattr(t, '__parameters__', ()):
+                    if x not in parameters:
+                        parameters.append(x)
 
         return tuple(parameters)
 
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2682,7 +2682,8 @@ if not hasattr(typing, "TypeVarTuple"):
 
                     expect_val = f"at least {elen}"
 
-            raise TypeError(f"Too {'many' if alen > elen else 'few'} arguments"
+            things = "arguments" if sys.version_info >= (3, 10) else "parameters"
+            raise TypeError(f"Too {'many' if alen > elen else 'few'} {things}"
                             f" for {cls}; actual {alen}, expected {expect_val}")
 else:
     # Python 3.11+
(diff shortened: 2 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -5605,6 +5605,23 @@ def test_typevartuple(self):
         class A(Generic[Unpack[Ts]]): ...
         Alias = Optional[Unpack[Ts]]
 
+    def test_erroneous_generic(self):
+        DefaultStrT = TypeVar('DefaultStrT', default=str)
```
JelleZijlstra, line 5609: A bit weird that this one uses plain TypeVar while the next test uses `typing_extensions.TypeVar`. They are the same (we do `from typing_extensions import TypeVar`), so we should stick to just one variation. I think it is clearer to use `typing_extensions.TypeVar` but don't feel strongly.

(No need to change this in other tests.)

src/typing_extensions.py
```diff
@@ -2378,9 +2336,139 @@ def wrapper(*args, **kwargs):
...
+                    for collected in _collect_parameters([x]):
+                        if collected not in parameters:
+                            parameters.append(collected)
+            elif hasattr(t, '__typing_subst__'):
+                if t not in parameters:
+                    if getattr(t, '__default__', None) is not None:
+                        if not default_encountered:
+                            default_encountered = True
+                    elif default_encountered:
+                        raise TypeError('expected TypeVar with default type, found'
```
JelleZijlstra, line 2460: I think this error message could be clearer. Also, TypeVar seems wrong as it could be a ParamSpec or TypeVarTuple instead.

For inspiration:
```
% python -c 'def f(x=1, y): ...'
  File "<string>", line 1
    def f(x=1, y): ...
               ^
SyntaxError: parameter without a default follows parameter with a default
```

So maybe "type parameter {t!r} without a default follows type parameter with a default".

src/typing_extensions.py
```diff
@@ -2378,9 +2336,139 @@ def wrapper(*args, **kwargs):
...
+                # deal with TypeVarLike defaults
+                # required TypeVarLikes cannot appear after a defaulted one.
+                if alen < elen:
+                    # since we validate TypeVarLike default in _collect_type_vars
+                    # or _collect_parameters we can safely check parameters[alen]
+                    if getattr(parameters[alen], '__default__', None) is not None:
+                        return
+
+            raise TypeError(f"Too {'many' if alen > elen else 'few'} parameters"
+                            f" for {cls}; actual {alen}, expected {elen}")
```
JelleZijlstra, line 2366: Agree the "expected {elen}" part here is problematic, as the precise number of type parameters accepted isn't necessarily fixed. I would look at the error messages from function calls in Python for how to make this better.

JelleZijlstra (commented):
Thanks! cc @Gobot1234

src/typing_extensions.py
```diff
@@ -145,27 +145,6 @@ def __repr__(self):
...
-            raise TypeError(f"{cls} is not a generic class")
-        elen = len(cls.__parameters__)
-    alen = len(parameters)
-    if alen != elen:
-        if hasattr(cls, "__parameters__"):
-            parameters = [p for p in cls.__parameters__ if not _is_unpack(p)]
-            num_tv_tuples = sum(isinstance(p, TypeVarTuple) for p in parameters)
-            if (num_tv_tuples > 0) and (alen >= elen - num_tv_tuples):
-                return
-        raise TypeError(f"Too {'many' if alen > elen else 'few'} parameters for {cls};"
```
NCPlayz, line 165: Was trying to figure out why the tests failed when I reuse this error and it is because typing_extensions raises with "parameters" whereas Python has always raised with "arguments", and this conditional is placed everywhere in tests:

https://github.com/python/typing_extensions/blob/fda0c15e3b5ed05703420cfb4d0974edb5e39c46/src/test_typing_extensions.py#L5234

In Python 3.12:

https://github.com/python/cpython/blob/cf29a2f25e359790c0c0f6a4ab0ab49ca5ffab5d/Lib/typing.py#L297

In Python 3.10:

https://github.com/python/cpython/blob/dcb16c98be61630369227f0d893f8d9262d25cac/Lib/typing.py#L231

Is this phrase change on purpose and should the tests be fixed accordingly?

src/typing_extensions.py
NCPlayz: Using https://github.com/python/cpython/blob/244567398370bfa62a332676762bb1db395c02fc/Lib/typing.py#L974-L975 as inspiration to change the error message

cdce8p:
What's the status of this one? With basic / experimental support for PEP 696 coming in the next mypy release this is basically necessary to use TypeVar defaults.

src/test_typing_extensions.py
```diff
@@ -5711,9 +5714,7 @@ class Y(Generic[T], NamedTuple):
                 self.assertIsInstance(a, G)
                 self.assertEqual(a.x, 3)
 
-                things = "arguments" if sys.version_info >= (3, 11) else "parameters"
-
-                with self.assertRaisesRegex(TypeError, f'Too many {things}'):
+                with self.assertRaisesRegex(TypeError, f'Too many parameters'):
```
Gobot1234, line 5717:
```suggestion
                with self.assertRaisesRegex(TypeError, 'Too many parameters'):
```
To fix CI

Gobot1234 (commented):
I don't have the necessary internal knowledge of how typing.py works anymore really to be able to review this properly

cdce8p:
> It would be good to get more review (if @Gobot1234 or you could review, I'd appreciate it).

👍🏻  I'll look at it tomorrow.

src/typing_extensions.py
```diff
@@ -2689,9 +2647,154 @@ def wrapper(*args, **kwargs):
 #   counting generic parameters, so that when we subscript a generic,
 #   the runtime doesn't try to substitute the Unpack with the subscripted type.
 if not hasattr(typing, "TypeVarTuple"):
+    def _check_generic(cls, parameters, elen=_marker):
+        """Check correct count for parameters of a generic cls (internal helper).
+        This gives a nice error message in case of count mismatch.
+        """
```
cdce8p, line 2653:
```suggestion
        """Check correct count for parameters of a generic cls (internal helper).

        This gives a nice error message in case of count mismatch.
        """
```
Insert an empty line to match the docstring in the `else` case.

src/typing_extensions.py
```diff
@@ -2689,9 +2647,154 @@ def wrapper(*args, **kwargs):
...
+        default_encountered = False
+        for t in types:
+            if (
+                isinstance(t, typevar_types) and
+                t not in tvars and
+                not _is_unpack(t)
+            ):
+                if getattr(t, '__default__', None) is not None:
+                    if not default_encountered:
+                        default_encountered = True
```
cdce8p, line 2746:
```suggestion
                    default_encountered = True
```
This can be simplified. The `if` check is unnecessary.

src/typing_extensions.py
```diff
@@ -2689,9 +2647,154 @@ def wrapper(*args, **kwargs):
...
+                # `[T, int]`, or `[int, *Ts]`, etc.
+                for x in t:
+                    for collected in _collect_parameters([x]):
+                        if collected not in parameters:
+                            parameters.append(collected)
+            elif hasattr(t, '__typing_subst__'):
+                if t not in parameters:
+                    if getattr(t, '__default__', None) is not None:
+                        if not default_encountered:
+                            default_encountered = True
```
cdce8p, line 2784:
```suggestion
                        default_encountered = True
```

src/typing_extensions.py
```diff
@@ -2689,9 +2647,154 @@ def wrapper(*args, **kwargs):
...
+                isinstance(t, typevar_types) and
+                t not in tvars and
+                not _is_unpack(t)
+            ):
+                if getattr(t, '__default__', None) is not None:
+                    if not default_encountered:
+                        default_encountered = True
+                elif default_encountered:
+                    raise TypeError(f'type parameter {t!r} without a default'
+                                    ' follows type parameter with a default')
```
cdce8p, line 2749:
```suggestion
                    raise TypeError(f'Type parameter {t!r} without a default'
                                    ' follows type parameter with a default')
```
I would suggest to capitalize the first word to match the other error messages.

src/typing_extensions.py
```diff
@@ -2689,9 +2647,154 @@ def wrapper(*args, **kwargs):
...
+                        if collected not in parameters:
+                            parameters.append(collected)
+            elif hasattr(t, '__typing_subst__'):
+                if t not in parameters:
+                    if getattr(t, '__default__', None) is not None:
+                        if not default_encountered:
+                            default_encountered = True
+                    elif default_encountered:
+                        raise TypeError(f'type parameter {t!r} without a default'
+                                        ' follows type parameter with a default')
```
cdce8p, line 2787:
```suggestion
                        raise TypeError(f'Type parameter {t!r} without a default'
                                        ' follows type parameter with a default')
```

src/typing_extensions.py
```diff
@@ -2689,9 +2647,154 @@ def wrapper(*args, **kwargs):
...
+
+                    num_default_tv = sum(getattr(p, '__default__', None)
+                                         is not None for p in parameters)
+
+                    elen -= num_default_tv
+
+                    expect_val = f"at least {elen}"
+
+            raise TypeError(f"Too {'many' if alen > elen else 'few'} parameters"
+                            f" for {cls}; actual {alen}, expected {expect_val}")
```
cdce8p, line 2685:
```suggestion
            raise TypeError(f"Too {'many' if alen > elen else 'few'} arguments"
                            f" for {cls}; actual {alen}, expected {expect_val}")
```
If we unify `parameters` and `arguments`, I believe it makes sense to use the "newer" one. I.e. Python 3.11+ uses `arguments`.

src/typing_extensions.py
```diff
@@ -2689,9 +2647,154 @@ def wrapper(*args, **kwargs):
...
```
cdce8p, line 2721:
```suggestion
            raise TypeError(f"Too {'many' if alen > elen else 'few'} arguments"
                            f" for {cls}; actual {alen}, expected {expect_val}")
```

src/test_typing_extensions.py
```diff
@@ -3262,6 +3262,9 @@ def __call__(self, *args: Unpack[Ts]) -> T: ...
         self.assertEqual(MemoizedFunc.__parameters__, (Ts, T, T2))
         self.assertTrue(MemoizedFunc._is_protocol)
 
+        # 3.11+ moves where this exception is thrown
+        # 3.11: TypeVarTuple.__typing_prepare_subst__
+        # 3.12+: typing._typevartuple_prepare_subst
```
cdce8p, line 3267:
```suggestion
```
Unrelated?

src/test_typing_extensions.py
```diff
@@ -3262,6 +3262,9 @@ def __call__(self, *args: Unpack[Ts]) -> T: ...
...
         things = "arguments" if sys.version_info >= (3, 11) else "parameters"
```
cdce8p, line 3268: This is no longer necessary if we use `arguments` for all versions.
The `assertRaisesRegex` below should than also be updated to use `arguments` instead.
https://github.com/python/typing_extensions/blob/4fdc09ddb54be26580f68e26443a422c6024364c/src/test_typing_extensions.py#L3272-L3273

src/test_typing_extensions.py
```diff
@@ -5711,9 +5714,7 @@ class Y(Generic[T], NamedTuple):
                 self.assertIsInstance(a, G)
                 self.assertEqual(a.x, 3)
 
-                things = "arguments" if sys.version_info >= (3, 11) else "parameters"
-
-                with self.assertRaisesRegex(TypeError, f'Too many {things}'):
+                with self.assertRaisesRegex(TypeError, 'Too many parameters'):
```
cdce8p, line 5717:
```suggestion
                with self.assertRaisesRegex(TypeError, 'Too many arguments'):
```

src/test_typing_extensions.py
cdce8p, line 3268:
```suggestion
        things = "arguments"
```

src/test_typing_extensions.py
```diff
@@ -6215,6 +6216,23 @@ def test_typevartuple(self):
...
+
+        with self.assertRaises(TypeError):
+            Test = Generic[DefaultStrT, T]
+
+    def test_need_more_params(self):
+        DefaultStrT = typing_extensions.TypeVar('DefaultStrT', default=str)
+        T = typing_extensions.TypeVar('T')
+        U = typing_extensions.TypeVar('U')
+
+        class A(Generic[T, U, DefaultStrT]): ...
```
cdce8p, line 6231:
```suggestion
        class A(Generic[T, U, DefaultStrT]): ...
        A[int, bool]
        A[int, bool, str]
```
I would add two good examples here, just so we test them too.

src/test_typing_extensions.py
```diff
@@ -6215,6 +6216,23 @@ def test_typevartuple(self):
...
+
+        with self.assertRaises(TypeError):
```
cdce8p, line 6233:
```suggestion
        with self.assertRaises(
            TypeError, "Too few arguments for .+; actual 1, expected at least 2"
        ):
```

cdce8p (commented):
Left a few comments. Most of them are just some minor style ones.
Overall this looks good to me! Thanks @NCPlayz 👍🏻 

--
@JelleZijlstra I left suggestions everywhere I could. If you like feel free to merge them. Would be awesome to get this released soon.

NCPlayz:
Thanks for the reviews! Fixed a minor bug but everything seems to be passing.

src/typing_extensions.py
```diff
@@ -2689,9 +2647,153 @@ def wrapper(*args, **kwargs):
...
+        """
+        if not elen:
+            raise TypeError(f"{cls} is not a generic class")
+        alen = len(parameters)
+        if alen != elen:
+            expect_val = elen
+            if hasattr(cls, "__parameters__"):
+                parameters = [p for p in cls.__parameters__ if not _is_unpack(p)]
+                num_tv_tuples = sum(isinstance(p, TypeVarTuple) for p in parameters)
+                if (num_tv_tuples > 0) and (alen >= elen - num_tv_tuples):
```
JelleZijlstra, line 2703: This condition isn't present on main, why do we need it here?

https://github.com/python/cpython/blob/2339e7cff745271f0e4a919573a347ab2bc1c2e9/Lib/typing.py#L289

src/typing_extensions.py
```diff
@@ -2689,9 +2647,153 @@ def wrapper(*args, **kwargs):
...
+                if t not in parameters:
+                    if getattr(t, '__default__', None) is not None:
+                        default_encountered = True
+                    elif default_encountered:
+                        raise TypeError(f'Type parameter {t!r} without a default'
+                                        ' follows type parameter with a default')
+
+                    parameters.append(t)
+            else:
+                if _should_collect_from_parameters(t):
```
JelleZijlstra, line 2790: This is different from the behavior on CPython main: https://github.com/python/cpython/blob/2339e7cff745271f0e4a919573a347ab2bc1c2e9/Lib/typing.py#L283

JelleZijlstra (commented):
Thanks. We need to be extra careful in the cases where we monkeypatch CPython, so I left some comments there.

I'll look into adding a test to this repo that runs the test_typing suite with the typing_extensions monkeypatches enabled.

JelleZijlstra:
Ci didn't run

JelleZijlstra:
Ci didn't run

JelleZijlstra:
I may have made CI not run any more somehow, will investigate.

src/typing_extensions.py
JelleZijlstra: Looks like this leads to a failure in the test_typing suite, could you make this mirror the CPython logic more closely?

NCPlayz: Gonna go do that now :pray:

NCPlayz:
Oh.. just noticed that some changes to unit tests might have broken in one of the merge commits

JelleZijlstra:
What specifically is broken? Feel free to send another PR with any fixes.

src/test_typing_extensions.py
```diff
@@ -5712,7 +5712,6 @@ class Y(Generic[T], NamedTuple):
                 self.assertEqual(a.x, 3)
 
                 things = "arguments" if sys.version_info >= (3, 10) else "parameters"
```
NCPlayz, line 5714: Seems like these lines got re-added -- but maybe they were breaking for 3.8/3.9?

JelleZijlstra: I think so, yes. I see you commented on this in https://github.com/python/typing_extensions/pull/293#discussion_r1375254093 before; I don't feel too strongly but I think it's best to match CPython's behavior.

PR #294, opened 2023-10-28
Fix deprecating a mixin

### LINKED ISSUE #251: @deprecated breaking __init__ inheritance when used with mixins/multiple inheritance
When a class without `__init__` (mixin) is used in multiple inheritance and marked as deprecated, inheritance of `__init__` doesn't work.

Example:

```py
from typing_extensions import deprecated


class Service:
    def __init__(self, foo: str, bar: str) -> None:
        print("Service.__init__")
        self.foo = foo
        self.bar = bar

    def echo(self) -> None:
        print("Service.echo", self.foo, self.bar)


class Base:
    def message(self) -> None:
        print("Base.message")


class Child(Base):
    def message(self) -> None:
        print("Child.message")
        super().message()


@deprecated("Use Child instead")	# removing this will work normally
class FooBarPrinter(Child):
    def message(self) -> None:
        print("FooBarPrinter.message")
        super().message()


class Spam(Service, FooBarPrinter):
    def do_work(self):
        self.echo()
        self.message()


if __name__ == "__main__":
    Spam("foo", "bar").do_work()
```

JelleZijlstra:
Fixes #251

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2337,7 +2337,6 @@ else:
                 return arg
             elif isinstance(arg, type):
                 original_new = arg.__new__
-                original_init_subclass = arg.__init_subclass__
 
                 @functools.wraps(original_new)
                 def __new__(cls, *args, **kwargs):
@@ -2353,12 +2352,28 @@ else:
 
                 arg.__new__ = staticmethod(__new__)
 
-                @functools.wraps(original_init_subclass)
-                def __init_subclass__(*args, **kwargs):
-                    warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
-                    return original_init_subclass(*args, **kwargs)
+                original_init_subclass = arg.__init_subclass__
+                # We need slightly different behavior if __init_subclass__
+                # is a bound method (likely if it was implemented in Python)
+                if isinstance(original_init_subclass, _types.MethodType):
+                    original_init_subclass = original_init_subclass.__func__
 
-                arg.__init_subclass__ = __init_subclass__
+                    @functools.wraps(original_init_subclass)
+                    def __init_subclass__(*args, **kwargs):
+                        warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
+                        return original_init_subclass(*args, **kwargs)
+
+                    arg.__init_subclass__ = classmethod(__init_subclass__)
+                # Or otherwise, which likely means it's a builtin such as
+                # type's implementation of __init_subclass__.
+                else:
+                    @functools.wraps(original_init_subclass)
+                    def __init_subclass__(*args, **kwargs):
+                        warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
+                        return original_init_subclass(*args, **kwargs)
+
+                    arg.__init_subclass__ = __init_subclass__
+
                 arg.__deprecated__ = __new__.__deprecated__ = msg
                 __init_subclass__.__deprecated__ = msg
                 return arg
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2359,7 +2359,7 @@ else:
                     return original_init_subclass(*args, **kwargs)
 
                 arg.__init_subclass__ = __init_subclass__
-                arg.__deprecated__ = __new__.__deprecated__ = arg
+                arg.__deprecated__ = __new__.__deprecated__ = msg
                 __init_subclass__.__deprecated__ = msg
                 return arg
             elif callable(arg):
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2337,24 +2337,29 @@ else:
                 return arg
             elif isinstance(arg, type):
                 original_new = arg.__new__
-                has_init = arg.__init__ is not object.__init__
+                original_init_subclass = arg.__init_subclass__
 
                 @functools.wraps(original_new)
                 def __new__(cls, *args, **kwargs):
-                    warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
+                    if cls is arg:
+                        warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
                     if original_new is not object.__new__:
                         return original_new(cls, *args, **kwargs)
                     # Mirrors a similar check in object.__new__.
-                    elif not has_init and (args or kwargs):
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
I realized we could do something smarter: inject `__init_subclass__` to warn when creating subclasses of a deprecated class, and then change the injected `__new__` to warn only if the class being instantiated is the deprecated class, not a subclass.

src/typing_extensions.py
```diff
@@ -2337,21 +2337,30 @@ def decorator(arg: _T, /) -> _T:
...
                 arg.__new__ = staticmethod(__new__)
-                arg.__deprecated__ = __new__.__deprecated__ = msg
+
+                @functools.wraps(original_init_subclass)
+                def __init_subclass__(*args, **kwargs):
+                    warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
+                    return original_init_subclass(*args, **kwargs)
+
+                arg.__init_subclass__ = __init_subclass__
+                arg.__deprecated__ = __new__.__deprecated__ = arg
```
AlexWaygood, line 2362: This should fix the failing test:

```suggestion
                arg.__deprecated__ = __new__.__deprecated__ = msg
```

src/typing_extensions.py
```diff
@@ -2337,21 +2337,30 @@ def decorator(arg: _T, /) -> _T:
...
                 arg.__new__ = staticmethod(__new__)
+
+                @functools.wraps(original_init_subclass)
+                def __init_subclass__(*args, **kwargs):
+                    warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
+                    return original_init_subclass(*args, **kwargs)
+
+                arg.__init_subclass__ = __init_subclass__
                 arg.__deprecated__ = __new__.__deprecated__ = msg
+                __init_subclass__.__deprecated__ = msg
```
AlexWaygood, line 2363: This doesn't always do the right thing, because of the fact that `__init_subclass__` is an implicit classmethod. For example:

```pycon
Python 3.11.5 (tags/v3.11.5:cce6ba9, Aug 24 2023, 14:38:34) [MSC v.1936 64 bit (AMD64)] on win32
>>> import typing_extensions as t
>>> @t.deprecated('DEPRECATION')
... class Foo:
...     ATTR = 1
...     def __init_subclass__(cls, *args, **kwargs):
...         print(cls.ATTR)
...
>>> class Bar:
...     ATTR = 2
...
>>> class Baz(Bar, Foo): pass
...
<stdin>:1: DeprecationWarning: DEPRECATION
1
>>> Baz.ATTR
2
```

You can fix it by doing this (diff is relative to your PR branch):

```diff
diff --git a/src/typing_extensions.py b/src/typing_extensions.py
index 6a0015a..37ef575 100644
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2337,7 +2337,7 @@ else:
                 return arg
             elif isinstance(arg, type):
                 original_new = arg.__new__
-                original_init_subclass = arg.__init_subclass__
+                original_init_subclass = arg.__init_subclass__.__func__

                 @functools.wraps(original_new)
                 def __new__(cls, *args, **kwargs):
@@ -2358,7 +2358,7 @@ else:
                     warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
                     return original_init_subclass(*args, **kwargs)

-                arg.__init_subclass__ = __init_subclass__
+                arg.__init_subclass__ = classmethod(__init_subclass__)
                 arg.__deprecated__ = __new__.__deprecated__ = msg
                 __init_subclass__.__deprecated__ = msg
                 return arg
```

AlexWaygood (changes requested):

src/typing_extensions.py
JelleZijlstra: Thanks for catching this. Your solution isn't quite right as it doesn't work for classes without a Python `__init_subclass__`, but I got something to work.

src/typing_extensions.py
```diff
@@ -2343,21 +2343,45 @@ def decorator(arg: _T, /) -> _T:
...
+                    original_init_subclass = original_init_subclass.__func__
+
+                    @functools.wraps(original_init_subclass)
+                    def __init_subclass__(*args, **kwargs):
+                        warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
+                        return original_init_subclass(*args, **kwargs)
+
+                    arg.__init_subclass__ = classmethod(__init_subclass__)
+                # Or otherwise, which likely means it's a builtin such as
+                # type's implementation of __init_subclass__.
```
AlexWaygood, line 2374: nit: `type` doesn't implement `__init_subclass__`; I think it just inherits the implementation from `object`:

```pycon
>>> type.__dict__['__init_subclass__']
KeyError: '__init_subclass__'
>>> object.__dict__['__init_subclass__']
<method '__init_subclass__' of 'object' objects>
```

```suggestion
                # object's implementation of __init_subclass__.
```

src/typing_extensions.py
```diff
@@ -2343,21 +2343,45 @@ def decorator(arg: _T, /) -> _T:
...
+                else:
+                    @functools.wraps(original_init_subclass)
+                    def __init_subclass__(*args, **kwargs):
+                        warnings.warn(msg, category=category, stacklevel=stacklevel + 1)
+                        return original_init_subclass(*args, **kwargs)
+
+                    arg.__init_subclass__ = __init_subclass__
```
AlexWaygood, line 2381: Not sure how often we hit this branch? Is it only when `original_init_subclass` is `object.__init_subclass__`? But I have no actual problem to point out here, I guess I'm just slightly queasy about how well it's tested right now :)

AlexWaygood (approved):
LGTM, thanks!

src/typing_extensions.py
AlexWaygood: Ah right, thanks -- I checked my solution worked in the REPL, but didn't actually run the test suite with my changes :) My bad!

src/typing_extensions.py
JelleZijlstra: Quite a lot, I think five or so tests failed before I implemented this.

AlexWaygood: Yeah I know — my point was that they were all failing for the same reason (`object.__init_subclass__` not being a classmethod) — I was curious whether there were any other realistic situations that would lead to us ending up in this branch :)

PR #302, opened 2023-11-29
Update @deprecated implementation

JelleZijlstra:
Copying over aspects of python/cpython#104004

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2388,7 +2388,7 @@ else:
 
                     arg.__init_subclass__ = classmethod(__init_subclass__)
                 # Or otherwise, which likely means it's a builtin such as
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -2371,7 +2388,7 @@ def __init_subclass__(*args, **kwargs):
 
                     arg.__init_subclass__ = classmethod(__init_subclass__)
                 # Or otherwise, which likely means it's a builtin such as
-                # object's implementation of __init_subclass__.
+                # type's implementation of __init_subclass__.
```
AlexWaygood, line 2391: we discussed this before in https://github.com/python/typing_extensions/pull/294#discussion_r1381249381 ;)

```suggestion
                # object's implementation of __init_subclass__.
```

PR #306, opened 2023-11-29
Prepare release 4.9.0rc1

JelleZijlstra:


AlexWaygood:
From the CI in #307, it looks like your PEP-705 PR might have broken stubtest. Maybe we should fix stubtest before cutting the release? :)

JelleZijlstra:
I think I'd prefer to get the RC out first. We can fix stubtest next. (For context, I'll be mostly-unreachable for about the second half of December; I want to make sure we get through the RC phase for this release before that.)

AlexWaygood:
> I think I'd prefer to get the RC out first. We can fix stubtest next.

I think we have to do that anyway, actually, looking at it. It looks like we'll need a one-two punch of (1) adding these new `ClassVar`s to typeshed's `typing_extensions` stub, then (2) doing a typeshed sync at mypy, to get stubtest's tests green again.

JelleZijlstra:
Done: https://pypi.org/project/typing-extensions/4.9.0rc1/

PR #312, opened 2023-12-08
fix: Interaction with get_type_hints and Annotated <3.11.

DanCardin:


brianschubert:
I think this has been superseded by #482?

Daraan:
> I think this has been superseded by #482?

I think so as well. Its a possible reference in case more needs to be pulled back at some time.

PR #329, opened 2024-02-08
Support __extra__ for PEP 728.

PIG208:


PIG208:
I assume that I need to update `doc/index.rst` and `CHANGELOG.md`. Will get back to it later this day.

PIG208:
Added documentation on this feature. I used 4.10.0 as the next version number, as I think `__extra__` wouldn't pose backwards compatibility issues given that dunder attributes are generally reserved.

srittau:
Thanks! Currently, there's some discussion about this in https://discuss.python.org/t/pep-728-typeddict-with-typed-extra-items/45443/9. We should hold off merging this for a few days, until the discussion is resolved.

JelleZijlstra:
Agree. My plan is to wait for a week or so for both this one and PEP 742 to reach a rough consensus in the discussion, then make a new typing-extensions release for both.

JelleZijlstra:
@PIG208 What do you think of implementing Eric's proposal from https://discuss.python.org/t/pep-728-typeddict-with-typed-extra-items/45443/24 ? That seems the best option so far.

JelleZijlstra:
In particular, if we can get the implementation ready by Sunday, it can go into the upcoming release 4.10.0. If you don't have time in the next few days, I can work on an implementation tomorrow.

PIG208:
I was working on updating the PEP. Will probably have both this implementation and the PEP update ready today. Thanks for notifying me!

PIG208:
I'm thinking about if and how we should include the special "__extra_item__" for runtime introspection. Perhaps it is not ideal to leave it in `__annotations__`, because the regular `__extra_item__` can also end up there and get potentially overridden in a child class that has the special `__extra_item__` key.

Should we instead use an attribute `__closed__` to host the annotated type when the special `__extra_item__` is used? If that's the case, should we copy the value of `__closed__` to the subclass during inheritance?

PIG208:
An alternative would be defining `__closed__` as a boolean indicating if the current TypedDict type is closed, and finding somewhere else to store the annotation type of the special extra item. This is similar to how `__total__` works.

However, for introspection, `__required_keys__` and `__optional_keys__` are generally more appropriate for `__total__`, and its semantics could be [confusing](https://github.com/python/cpython/pull/109544#pullrequestreview-1631314105). So I'm not sure if the boolean flag will be necessary for PEP 728.

JelleZijlstra:
I think the general principle should be that we should allow runtime typing tools to reconstruct what the user wrote as much as possible, without deciding for them what the exact semantics are. After all, there may be edge cases (e.g. involving forward references) where the runtime typing tool is able to use more advanced logic than typing-extensions to figure out the right behavior. However, we have to balance that argument with creating an intuitive interface for users.

I do think we need two attributes: `__closed__` matching the value of the `closed=` class argument, and `__extra_keys__` with the annotation for the magic key `__extra_keys__`. If we didn't have those two keys, we couldn't distinguish between a class with `__extra_keys__: None` (weird but legal) and one with no `__extra_keys__` at all.

I'd want the following semantics:

```python
class TD1(TypedDict):
    a: int
assert TD1.__closed__ is False
assert TD1.__extra_keys__ is None
assert TD1.__annotations__ == {'a': int}

class TD2(TypedDict, closed=True):
    a: int
assert TD2.__closed__ is True
assert TD2.__extra_keys__ is Never
assert TD2.__annotations__ == {'a': int}

class TD3(TypedDict, closed=True):
    a: int
    __extra_keys__: str
assert TD3.__closed__ is True
assert TD3.__extra_keys__ is str
assert TD3.__annotations__ == {'a': int}

class TD4(TD2):
    b: int

assert TD4.__closed__ is True
assert TD4.__extra_keys__ is Never
assert TD4.__annotations__ == {'a': int, 'b': int}

class TD5(TD3):
    b: int

assert TD5.__closed__ is True
assert TD5.__extra_keys__ is str
assert TD5.__annotations__ == {'a': int, 'b': int}

class TD6(TD2, closed=False):  # resets inheritance of __extra_keys__
    b: int

assert TD6.__closed__ is False
assert TD6.__extra_keys__ is None
assert TD6.__annotations__ == {'a': int, 'b': int}

class TD7(TD3):
    __extra_keys__: str  # just a regular key

assert TD7.__closed__ is True
assert TD7.__extra_keys__ is int
assert TD7.__annotations__ == {'a': int, '__extra_keys__': str}
```

PIG208:
It makes sense to have both dunder attributes, but I'm surprised by the "resetting inheritance of __extra_keys__" behavior. 

I expected that `__extra_keys__` always gets inherited and `__closed__` depends only on the TypedDict type instead of its bases, like what `__total__` currently does:

```py
class TD6(TD2, closed=False):
    b: int

assert TD6.__closed__ is False
assert TD6.__extra_keys__ is Never  # instead of None
assert TD6.__annotations__ == {'a': int, 'b': int}

```

Was this decision driven by distinguishing TypedDict with `__extra_keys__: None` and one without `__extra_keys__`?

PIG208:
This current implementation inherits `__extra_items__` from the bases (assuming there are no conflicts) and sets `__closed__` only based on the current TypedDict type. A drawback is that we have to avoid setting `__extra_items__` on non-closed `TypedDict` types to distinguish missing`__extra_items__` from `__extra_items__: None`.

doc/index.rst
```diff
@@ -394,6 +394,32 @@ Special typing primitives
       are mutable if they do not carry the :data:`ReadOnly` qualifier.
 
       .. versionadded:: 4.9.0
+    
+    The experimental ``closed`` keyword argument and the special key
+    ``"__extra_items__"`` proposed in :pep:`728` are supported.
```
JelleZijlstra, line 399:
```suggestion
    ``__extra_items__`` proposed in :pep:`728` are supported.
```

doc/index.rst
```diff
@@ -394,6 +394,32 @@ Special typing primitives
...
+
+    When ``closed`` is unspecified or ``closed=False`` is given,
+    ``"__extra_items__"`` behave like a regular key. Otherwise, this becomes a
```
JelleZijlstra, line 402:
```suggestion
    ``__extra_items__`` behaves like a regular key. Otherwise, this becomes a
```

doc/index.rst
```diff
@@ -394,6 +394,32 @@ Special typing primitives
...
+    special key that does not show up in ``__readonly_keys__``,
+    ``__mutable_keys__``, ``__required_keys__``, ``__optional_keys``, or
+    ``__annotations__``.
+
+    For runtime introspection, two attributes can be looked at:
+
+    .. attribute:: __closed__
+
+        A boolean flag indicating the value of the keyword argument ``closed``
+        on the current ``TypedDict``.
```
JelleZijlstra, line 412:
```suggestion
        A boolean flag indicating whether the current ``TypedDict`` is considered closed.
```

doc/index.rst
```diff
@@ -394,6 +394,32 @@ Special typing primitives
...
+
+        .. versionadded:: 4.10.0
+
+    .. attribute:: __extra_items__
+
+        The type annotation of the extra items allowed on the ``TypedDict``.
+        This attribute does not appear on a TypedDict that has itself and all
```
JelleZijlstra, line 419: It should always exist, attributes that exist conditionally are a nuisance. The docs should specify that it is None both if `closed=False` and if it's explicitly set to None.

src/test_typing_extensions.py
```diff
@@ -4177,6 +4177,121 @@ class AllTheThings(TypedDict):
...
         self.assertEqual(AllTheThings.__mutable_keys__, frozenset({'d'}))
+    
+    def test_extra_keys_non_readonly(self):
+        class Base(TypedDict, closed=True):
+            __extra_items__: str
+        
+        class Child(Base):
+            a: NotRequired[int]
+        
+        self.assertEqual(Child.__required_keys__, frozenset({}))
```
JelleZijlstra, line 4188: Also test the value of `__closed__` and `__extra_items__`. Should also do this on most other classes below.

src/test_typing_extensions.py
```diff
@@ -4177,6 +4177,121 @@ class AllTheThings(TypedDict):
...
+        self.assertEqual(GrandChild.__optional_keys__, frozenset({}))
+        self.assertEqual(GrandChild.__readonly_keys__, frozenset({}))
+        self.assertEqual(GrandChild.__mutable_keys__, frozenset({'a', "__extra_items__"}))
+        self.assertEqual(GrandChild.__extra_items__, str)
+
+        self.assertEqual(Base.__annotations__, {})
+        self.assertEqual(Child.__annotations__, {"__extra_items__": int, "a": int})
+        self.assertEqual(GrandChild.__annotations__, {"__extra_items__": int, "a": int})
+
+        self.assertTrue(Base.__closed__)
```
JelleZijlstra, line 4276: Move all annotations on the same class together

src/typing_extensions.py
```diff
@@ -969,7 +987,7 @@ def __subclasscheck__(cls, other):
     _TypedDict = type.__new__(_TypedDictMeta, 'TypedDict', (), {})
 
     @_ensure_subclassable(lambda bases: (_TypedDict,))
-    def TypedDict(typename, fields=_marker, /, *, total=True, **kwargs):
+    def TypedDict(typename, fields=_marker, /, *, total=True, closed=False, **kwargs):
```
JelleZijlstra, line 990: For backward compatibility, if `closed` exists but is not True/False, add it to kwargs. Add a test case like this:

```
TD = TypedDict("TD", closed=int)
assert TD.__closed__ is False
assert TD.__annotations__ = {"closed": int}
```

doc/index.rst
PIG208: I realized that it would be `type(None)` if it is explicitly set to `None`, so that would be even better.

src/typing_extensions.py
```diff
@@ -929,6 +930,20 @@ def __new__(cls, name, bases, ns, *, total=True):
...
+            if closed and extra_items_type is None:
+                extra_items_type = Never
+            if closed and "__extra_items__" in own_annotations:
+                annotation_type = own_annotations.pop("__extra_items__")
+                qualifiers = set(_get_typeddict_qualifiers(annotation_type))
+                if Required in qualifiers or NotRequired in qualifiers:
+                    raise TypeError(
+                        "Special key __extra_items__ does not support "
+                        "Required and NotRequired"
+                    )
```
JelleZijlstra, line 945:
```suggestion
                if Required in qualifiers:
                    raise TypeError(
                        "Special key __extra_items__ does not support "
                        "Required"
                    )
                if NotRequired in qualifiers:
                    raise TypeError(
                        "Special key __extra_items__ does not support "
                        "NotRequired"
                    )
```
Nicer for users to give a specific error

src/test_typing_extensions.py
```diff
@@ -4177,6 +4195,134 @@ class AllTheThings(TypedDict):
...
+        self.assertEqual(ChildA.__extra_items__, Never)
+        self.assertTrue(ChildA.__closed__)
+
+        class ChildB(Base, closed=True):
+            __extra_items__: None
+
+        self.assertEqual(ChildB.__extra_items__, type(None))
+        self.assertTrue(ChildB.__closed__)
+
+    def test_backwards_compatibility(self):
```
JelleZijlstra, line 4321: This test needs to be gated so it only runs before 3.13, I think there's a decorator for that.

AlexWaygood: We have a `skip_if_early_py313_alpha` decorator, but it only skips the test if you're running 3.13a2 or an earlier alpha. You should be able to easily follow a similar approach, though

JelleZijlstra:
Also a few lint errors in CI

src/test_typing_extensions.py
PIG208: I just added `TYPING_3_13_0 = sys.version_info[:3] >= (3, 13, 0)` and use `skipIf` to deal with this issue.

PIG208:
Hm, fixing the errors now.

JelleZijlstra (approved):

PR #330, opened 2024-02-09
Add support for TypeIs (PEP 742)

JelleZijlstra:
Cross-linking python/peps#3649.


JelleZijlstra:
I picked the name TypeNarrower, but it's very possible that we'll eventually end up with a different name. That could raise problems for backwards compatibility in typing-extensions, but I think it's fine: we'd simply rename TypeNarrower to the new name, and keep around TypeNarrower as an alias indefinitely.

srittau (approved):

srittau (approved):
Good name.

PR #332, opened 2024-02-13
Update GitHub Actions versions

srittau:


AlexWaygood (approved):
Thanks! We need to do this over at typeshed as well -- we're getting deprecation warnings in CI about the node.js version of the actions we're using

PR #333, opened 2024-02-14
Drop runtime error in PEP 705 implementation

JelleZijlstra:
See https://discuss.python.org/t/pep-705-read-only-typeddict-items/37867/10

cc @alicederyn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Drop runtime error when a mutable `TypedDict` key overrides a read-only
+  one. Type checkers should still flag this as an error. Patch by Jelle
+  Zijlstra.
 - Speedup `issubclass()` checks against simple runtime-checkable protocols by
   around 6% (backporting https://github.com/python/cpython/pull/112717, by Alex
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Drop runtime error when a mutable `TypedDict` key overrides a read-only
```
alicederyn, line 3: This comment is back to front I think @JelleZijlstra ? May be confusing

JelleZijlstra: You're right, thanks

src/typing_extensions.py
```diff
@@ -942,11 +942,6 @@ def __new__(cls, name, bases, ns, *, total=True):
                 else:
                     optional_keys.add(annotation_key)
                 if ReadOnly in qualifiers:
-                    if annotation_key in mutable_keys:
```
alicederyn, line 945: Did you not want to replace this with a `mutable_keys.discard`? I think that would be symmetric with how this code handles other "illegal" overrides, e.g. with Required.

PR #334, opened 2024-02-14
Fix changelog entry for PEP 705 change

JelleZijlstra:
cc @alicederyn


JelleZijlstra:
I also need to apply https://github.com/python/typing_extensions/pull/333/files#r1489942541 (don't have time right now)

PR #348, opened 2024-03-04
Stabilise third party tests

hauntsaninja:
Use uv to test with the state of PyPI as of the commit we are testing

(except for pdm projects, leave those alone, hopefully they have some sort of lock file anyway)

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -103,12 +103,14 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install uv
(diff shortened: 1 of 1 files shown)


hauntsaninja:
Hmm, qcore! https://github.com/astral-sh/uv/issues/2145
Also would be a nicer diff with https://github.com/astral-sh/uv/issues/313
And https://github.com/astral-sh/uv/issues/2088 for making the feature public

JelleZijlstra:
I was hoping that this would get better because I released qcore 1.10.1 with 3.12 wheels, but I guess it doesn't get picked up because you limited us to packages released before the most recent commit.

I'm guessing I may need to apply the same change to asynq as I applied to qcore; I can work on that but don't have time right now.

PR #349, opened 2024-03-05
Make sure that `ReadOnly` is removed when using `get_type_hints(include_extra=False)`

sobolevn:
I've noticed that while working on https://github.com/python/cpython/issues/116127


JelleZijlstra (commented):
Thanks! Could you add a changelog entry and a versionchanged entry in the docs? Something like "When `include_extra=False`, `get_type_hints()` now strips :data:`ReadOnly` from the annotation."

sobolevn:
Done, I assumed that the next version is 4.11.0

JelleZijlstra:
Thank you!

PR #365, opened 2024-04-08
Add dependabot for GitHub Actions

srittau:


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: monthly
```
AlexWaygood, line 6: Maybe group them into one bundled PR, rather than having a separate PR for each action? No strong opinion, though

```suggestion
      interval: monthly
    groups:
      actions:
        patterns:
          - "*"
```

AlexWaygood (approved):

.github/dependabot.yml
srittau: I wondered why we grouped them in python/typing, makes sense.

PR #367, opened 2024-04-10
Make the daily test not fail tonight on Python 3.13

AlexWaygood:
it's about to fail because of https://github.com/python/cpython/pull/115913


JelleZijlstra:
This might be worth a changelog entry since redistributors may well complain about it.

AlexWaygood:
harumph

JelleZijlstra:
Thanks!

Issue #377, opened 2024-05-10
Python 3.13.0b1: test_typing_extensions: `AttributeError: attribute '__default__' of 'typing.TypeVar' objects is not writable`


befeleme:
When building typing_extensions in Fedora with Python 3.13.0b1, `test_typing_extensions` fails with the traceback below.
I looked at the recent commits but they don't seem to tackle that. It hasn't yet been seen in the project's CI, probably because it still runs on alpha 6.

```
/usr/bin/python3 -m unittest discover
E
======================================================================
ERROR: test_typing_extensions (unittest.loader._FailedTest.test_typing_extensions)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_typing_extensions
Traceback (most recent call last):
  File "/usr/lib64/python3.13/unittest/loader.py", line 396, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib64/python3.13/unittest/loader.py", line 339, in _get_module_from_name
    __import__(name)
    ~~~~~~~~~~^^^^^^
  File "/builddir/build/BUILD/typing_extensions-4.11.0/src/test_typing_extensions.py", line 1148, in <module>
    T_a = TypeVar('T_a')
  File "/builddir/build/BUILD/typing_extensions-4.11.0/src/typing_extensions.py", line 1406, in __new__
    _set_default(typevar, default)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/builddir/build/BUILD/typing_extensions-4.11.0/src/typing_extensions.py", line 1363, in _set_default
    type_param.__default__ = None
    ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: attribute '__default__' of 'typing.TypeVar' objects is not writable
```

AlexWaygood:
#378 fixed all but one test failure. The last remaining one is to do with `AsyncContextManager` having an optional second parameter on Python 3.13, but not in the backported version that `typing_extensions` provides.

JelleZijlstra:
@AlexWaygood the changes you made didn't look related to the test failure @befeleme pointed out here. Do you know what's going on here?

Also, there's another failure incoming due to changes I made for beta 2:

```
% ~/py/cpython/python.exe test_typing_extensions.py 
Traceback (most recent call last):
  File "/Users/jelle/py/typing_extensions/src/test_typing_extensions.py", line 42, in <module>
[...]

PR #378, opened 2024-05-10
Fix most tests on Python 3.13

AlexWaygood:
Helps with #377. Lots of changes to reflect changes in the PEP-696 implementation that were made upstream but had yet to be backported here.

There's still a couple of failing tests on Python 3.13 even with this PR, but this gets us a lot closer to a clean bill of health.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -38,7 +38,7 @@ from typing_extensions import assert_type, get_type_hints, get_origin, get_args,
 from typing_extensions import clear_overloads, get_overloads, overload
 from typing_extensions import NamedTuple, TypeIs
 from typing_extensions import override, deprecated, Buffer, TypeAliasType, TypeVar, get_protocol_members, is_protocol
-from typing_extensions import Doc
+from typing_extensions import Doc, NoDefault
 from _typed_dict_test_helper import Foo, FooGeneric, VeryAnnotated
 
 # Flags used to mark tests that only apply after a specific
@@ -6207,12 +6207,15 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
     def test_typevar_none(self):
         U = typing_extensions.TypeVar('U')
         U_None = typing_extensions.TypeVar('U_None', default=None)
-        self.assertEqual(U.__default__, None)
-        self.assertEqual(U_None.__default__, type(None))
+        self.assertIs(U.__default__, NoDefault)
+        self.assertFalse(U.has_default())
+        self.assertEqual(U_None.__default__, None)
+        self.assertTrue(U_None.has_default())
 
     def test_paramspec(self):
         P = ParamSpec('P', default=(str, int))
         self.assertEqual(P.__default__, (str, int))
+        self.assertTrue(P.has_default())
         self.assertIsInstance(P, ParamSpec)
         if hasattr(typing, "ParamSpec"):
             self.assertIsInstance(P, typing.ParamSpec)
@@ -6225,11 +6228,13 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
 
         P_default = ParamSpec('P_default', default=...)
         self.assertIs(P_default.__default__, ...)
+        self.assertTrue(P_default.has_default())
 
     def test_typevartuple(self):
         Ts = TypeVarTuple('Ts', default=Unpack[Tuple[str, int]])
         self.assertEqual(Ts.__default__, Unpack[Tuple[str, int]])
         self.assertIsInstance(Ts, TypeVarTuple)
+        self.assertTrue(Ts.has_default())
         if hasattr(typing, "TypeVarTuple"):
             self.assertIsInstance(Ts, typing.TypeVarTuple)
             typing_Ts = typing.TypeVarTuple('Ts')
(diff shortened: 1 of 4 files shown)


AlexWaygood:
@JelleZijlstra any idea what's up with the `pickle` test failing? It's only failing on Python 3.13, which suggests it might be an issue with the CPython implementation, and I can repro locally in my CPython clone:

```pycon
(main) % ./python.exe                                                                                                                ~/dev/cpython
Python 3.14.0a0 (heads/main:66f8bb76a1, May  8 2024, 17:00:15) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> from typing import NoDefault
>>> import pickle
>>> pickle.dumps(NoDefault)
  File "<python-input-2>", line 1, in <module>
    pickle.dumps(NoDefault)
    ~~~~~~~~~~~~^^^^^^^^^^^
_pickle.PicklingError: can't pickle 'NoDefaultType' object: typing.NoDefault
>>> import typing
>>> pickle.dumps(typing.NoDefault)
  File "<python-input-4>", line 1, in <module>
    pickle.dumps(typing.NoDefault)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
_pickle.PicklingError: can't pickle 'NoDefaultType' object: typing.NoDefault
```

What I don't understand is why the pickling tests are _passing_ in the CPython test suite: https://github.com/python/cpython/blob/7ac933e2609b2ef9b08ccf9c815b682b0e1ede2a/Lib/test/test_typing.py#L10230-L10234

src/test_typing_extensions.py
```diff
@@ -6276,6 +6281,34 @@ def test_pickle(self):
...
+        self.assertRegex(repr(NoDefault), r'typing(_extensions)?\.NoDefault')
+
+    def test_no_call(self):
+        with self.assertRaises(TypeError):
+            NoDefault()
+
+    @skipIf(
+        sys.version_info[:5] == (3, 13, 0, "beta", 1),
+        "incorrectly raises TypeError in the first 3.13 beta"
+    )
```
AlexWaygood, line 6306: I opened https://github.com/python/cpython/issues/118895

JelleZijlstra:
The pickle thing is weird; I can reproduce it on CPython when I comment out the `tests.addTests(doctest.DocTestSuite(typing))` call at the bottom of `test_typing.py`. Something must have a side effect but I'm not sure what.

AlexWaygood:
> The pickle thing is weird; I can reproduce it on CPython when I comment out the `tests.addTests(doctest.DocTestSuite(typing))` call at the bottom of `test_typing.py`. Something must have a side effect but I'm not sure what.

That _is_ weird. I guess I'll skip the test over here for now.

JelleZijlstra:
Found it:

```
% ./python.exe
Python 3.14.0a0 (heads/main:7ac933e260, May 10 2024, 05:30:33) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import typing, pickle
>>> pickle.dumps(typing.NoDefault)
  File "<python-input-1>", line 1, in <module>
    pickle.dumps(typing.NoDefault)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
_pickle.PicklingError: can't pickle 'NoDefaultType' object: typing.NoDefault
>>> typing.NoDefault.__doc__
  File "<python-input-2>", line 1, in <module>
    typing.NoDefault.__doc__
AttributeError: 'NoDefaultType' object has no attribute '__doc__'. Did you mean: '__dir__'?
>>> pickle.dumps(typing.NoDefault)
b'\x80\x04\x95\x19\x00\x00\x00\x00\x00\x00\x00\x8c\x07_typing\x94\x8c\tNoDefault\x94\x93\x94.'
>>> typing.NoDefault.__doc__
'The type of the NoDefault singleton.'
```

The good news is that I think python/cpython#118897 fixes this. (Nikita just commented as much.)

AlexWaygood:
Alright, so just one failing test left now:

```pytb
======================================================================
FAIL: test_async_contextmanager (test_typing_extensions.OtherABCTests.test_async_contextmanager)
----------------------------------------------------------------------
  File "/home/runner/work/typing_extensions/typing_extensions/src/test_typing_extensions.py", line 1652, in test_async_contextmanager
    self.assertEqual(typing_extensions.AsyncContextManager[int].__args__, (int,))
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Tuples differ: (<class 'int'>, bool | None) != (<class 'int'>,)

First tuple contains 1 additional elements.
First extra element 1:
bool | None

- (<class 'int'>, bool | None)
+ (<class 'int'>,)
```

The principled way of fixing that is to make `typing_extensions.AsyncContextManager` have an optional second type parameter, the same as `typing.AsyncContextManager` on py313+. But I think doing that might possibly involve backporting the whole `typing._GenericAlias` implementation from CPython, which I don't really wanna do in this PR.

JelleZijlstra (approved):

JelleZijlstra:
Let's merge this now to fix a number of the tests; we can fix the rest separately.

erlend-aasland:
Thanks for fixing these! Currently, the `canopen` test suite fails to run on Python 3.13; looks like this should do the trick. Is there a 4.11.1 release scheduled?

AlexWaygood:
> Thanks for fixing these! Currently, the `canopen` test suite fails to run on Python 3.13; looks like this should do the trick. Is there a 4.11.1 release scheduled?

If we cut a new release right now, it would break pydantic, which would be very bad (https://github.com/python/typing_extensions/issues/381) -- we've broken pydantic before, and the experience was extremely chaotic for us, we definitely don't want to do it again 🙃

I've opened a PR to fix pydantic's tests with the `typing_extensions` main branch here: https://github.com/pydantic/pydantic/pull/9426.

erlend-aasland:
Great! I'm watching https://github.com/pydantic/pydantic/pull/9426 and keeping an eye out for new releases of `typing-extensions` :)

AlexWaygood:
typing_extensions 4.12.0rc1 is now out @erlend-aasland! :D https://pypi.org/project/typing-extensions/4.12.0rc1/

JelleZijlstra:
The final release should happen in a week (next Thursday).

PR #379, opened 2024-05-10
Remove some unnecessary monkeypatching on Python 3.13+

AlexWaygood:
Even if https://github.com/python/cpython/pull/118900 is merged, I think we should still make this change: as a general principle, we should aim to monkeypatch the typing module as little as possible


AlexWaygood:
The 3.13 test failure is the pre-existing one we already have on `main`

PR #385, opened 2024-05-13
Add more tests for `(Async)Generator`

AlexWaygood:
Now that we reimplement these from CPython -- and now that they have a separate implementation from all the other `collections.abc` aliases in `typing_extensions` -- we should add more tests for them to make sure that all code paths are covered. Most of these tests are adapted from CPython's tests for the `collections.abc` aliases.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1590,12 +1590,92 @@ class CollectionsAbcTests(BaseTestCase):
         self.assertIsInstance(d, collections.Counter)
         self.assertIsInstance(d, typing_extensions.Counter)
 
-    def test_async_generator(self):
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

AlexWaygood:
> (We should do the same for `(Async)ContextManager`, but this PR felt big enough as it was; I'll do those separately.)

Actually, I think adding more tests for those would maybe just be duplicative. CPython doesn't include any more tests for the `contextlib` aliases specifically, and since we use the same `_SpecialGenericAlias` implementation for them, I think all the relevant code paths are already covered by the tests I've just added here for `(Async)Generator`.

PR #390, opened 2024-05-13
Backport `types.CapsuleType`

### LINKED ISSUE #374: Backport CapsuleType from 3.13
See https://github.com/python/cpython/issues/109562. `types.CapsuleType` was added in https://github.com/python/cpython/pull/109600 but cannot be used before 3.13.

AlexWaygood:
Fixes #374. I'm still not sure how useful this really is, but (as long as this all works with PyPy), it's pretty easy to backport -- so why not.


AlexWaygood:
Yeah, I thought that might happen. I suppose we could expose it on CPython only?

JelleZijlstra (approved):

PR #392, opened 2024-05-14
Allow type parameters without default values to follow those with default values in some situations

AlexWaygood:
This is a backport of https://github.com/python/cpython/pull/118774. The implementation feels unbelievably hacky (we're looking up the value of a local variable in frame two levels up the call stack), but I can't see another way of doing it. I initially tried an approach where we monkeypatched `_GenericAlias.__init__`, but this didn't work: see https://github.com/python/cpython/pull/118774#issuecomment-2108631342.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   at runtime rather than `types.NoneType`.
 - Fix most tests for `TypeVar`, `ParamSpec` and `TypeVarTuple` on Python
   3.13.0b1 and newer.
+- Backport CPython PR [#118774](https://github.com/python/cpython/pull/118774),
+  allowing type parameters without default values to follow those with
+  default values in some type parameter lists. Patch by Alex Waygood,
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
Working on the merge conflict. I think I wouldn't want to use a `_getframe`-based hack, but I'll have to think of an alternative.

AlexWaygood:
> I think I wouldn't want to use a `_getframe`-based hack, but I'll have to think of an alternative.

Yeah. I think we definitely need to make _some_ change to our current implementation at `typing_extensions`, since currently we're strictly more prohibitive than the CPython implementation, and that's not great. If we really don't like this hack, though, I think it would be okay (though not idea.) to be strictly more _lenient_ than the CPython implementation -- simply not check the order of the parameters under any circumstances.

JelleZijlstra (approved):

JelleZijlstra:
Thinking about it more, I think this is OK. It's well localized and only affects old versions of typing.

PR #394, opened 2024-05-16
Don't turn list defaults for ParamSpecs into tuples

AlexWaygood:
PEP 696 specifies that default values for ParamSpecs must be lists, and CPython on 3.13 keeps them as a list for the `__default__` attribute:

```pycon
Python 3.14.0a0 (heads/cleanup-typing:2e98d5dbf9, May 11 2024, 12:20:48) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> def foo[**P = [str, int]](): pass
... 
>>> foo.__type_params__[0].__default__
[<class 'str'>, <class 'int'>]
>>> 
```

Currently in the `typing_extensions` version, however, we turn any list into a tuple, which seems wrong:

```pycon
(main) % python                                                                                                          ~/dev/typing_extensions/src
Python 3.12.3 (main, Apr 30 2024, 10:12:02) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> from typing_extensions import ParamSpec
>>> ParamSpec("P", default=[int, str]).__default__
(<class 'int'>, <class 'str'>)
```

This PR fixes that.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -4997,7 +4997,7 @@ class ParamSpecTests(BaseTestCase):
         P = ParamSpec('P')
         P_co = ParamSpec('P_co', covariant=True)
         P_contra = ParamSpec('P_contra', contravariant=True)
-        P_default = ParamSpec('P_default', default=int)
+        P_default = ParamSpec('P_default', default=[int])
         for proto in range(pickle.HIGHEST_PROTOCOL):
             with self.subTest(f'Pickle protocol {proto}'):
                 for paramspec in (P, P_co, P_contra, P_default):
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -1468,8 +1468,9 @@ def __reduce__(self):
 def _set_default(type_param, default):
     type_param.has_default = lambda: default is not NoDefault
     if isinstance(default, (tuple, list)):
-        type_param.__default__ = tuple((typing._type_check(d, "Default must be a type")
-                                        for d in default))
+        type_param.__default__ = type(default)(
+            typing._type_check(d, "Default must be a type") for d in default
+        )
```
AlexWaygood, line 1473: The `_type_check()` calls here feel "morally correct" -- they're consistent with what a lot of the rest of the typing.py internals do. But on the other hand, CPython doesn't actually do any check at all in its 3.13 implementation; it just sets the `__default__` attribute to the value passed in, and moves on.

src/test_typing_extensions.py
```diff
@@ -6363,8 +6363,8 @@ def test_typevar_none(self):
         self.assertTrue(U_None.has_default())
 
     def test_paramspec(self):
-        P = ParamSpec('P', default=(str, int))
-        self.assertEqual(P.__default__, (str, int))
+        P = ParamSpec('P', default=[str, int])
```
JelleZijlstra, line 6366: Maybe add another test that uses a tuple.

src/typing_extensions.py
JelleZijlstra: So maybe we should just do that here too?

AlexWaygood: I wasn't sure whether the bug was that we were being too lenient or that CPython wasn't being restrictive enough. But having slept on it, I now agree that we should just accept any default value, and not do any type checking here. That's consistent with what CPython has always done for TypeVar bounds and TypeVar constraints, which should also conceptually be types:

```pycon
% pyenv local 3.11                                                                                                % python
Python 3.11.8 (main, Feb 15 2024, 19:32:18) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
>>> from typing import TypeVar
>>> def not_a_type(): pass
... 
>>> T = TypeVar("T", bound=not_a_type)
>>> U = TypeVar("U", not_a_type, not_a_type)
```

Issue #61, opened 2022-07-27
_typed_dict_test_helper.py is missing from the PyPI archive


kulikjak:
Hi, the archive downloaded from PyPI does include `test_typing_extensions.py`, but it omits `_typed_dict_test_helper.py` meaning that  running `pytest test_typing_extensions.py` results in:
```
ImportError while importing test module 'xxxx/typing_extensions-4.3.0/src/test_typing_extensions.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_typing_extensions.py:32: in <module>
    from _typed_dict_test_helper import FooGeneric
E   ModuleNotFoundError: No module named '_typed_dict_test_helper'
```

srittau:
Thank you for the report, I think I spotted the error in our `pyproject.toml` file.

srittau:
Actually when I run `flit build` locally I get totally different files than what's on PyPI. (And definitely too many, like the whole `.github` directory.) Cc @JelleZijlstra, who did the last release and might have more experience with flit.

hegjon:
Fixed in #97

AlexWaygood:
> Fixed in #97

Thanks!

PR #395, opened 2024-05-16
Add workflow for Trusted Publishing

JelleZijlstra:


src/test_typing_extensions.py
```diff
@@ -5718,8 +5717,7 @@ def test_typing_extensions_defers_when_possible(self):
                     getattr(typing, item))
 
     def test_typing_extensions_compiles_with_opt(self):
-        file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                                 'typing_extensions.py')
+        file_path = typing_extensions.__file__
```
hauntsaninja, line 5720: dumb question, why is this needed?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,106 @@
...
+          export path_to_file=$(find dist -type f -name "typing_extensions-*.tar.gz")
+          echo "::notice::Installing source distribution: $path_to_file"
+          python3 -m pip install --user $path_to_file
+      - name: Run typing_extensions tests against installed package
+        run: rm src/typing_extensions.py && python3 src/test_typing_extensions.py
+
+  publish-to-pypi:
+    name: >-
+      Publish Python distribution to PyPI
+    if: startsWith(github.ref, 'refs/tags/') # only publish to PyPI on tag pushes
```
hauntsaninja, line 87: should we validate tag name matches version?

hauntsaninja (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,106 @@
+# Based on
+# https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+
+name: Publish Python distribution to PyPI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
srittau, line 10: Should this workflow really run on every PR? It seems unlikely to break and basically duplicates the normal test workflow. Also there is a slight risk that an error in this workflow might accidentally publish a package, and running it less frequently reduces that risk.

To me, it would make more sense to just run this workflow on commits to main.

src/test_typing_extensions.py
srittau: Probably because in the distribution tests above, `typing_extensions.py` gets deleted (to test the installed version).

.github/workflows/publish.yml
JelleZijlstra: I wanted it to run on *this* PR so it gets tested :).

But agree we don't have to run it on every PR.

.github/workflows/publish.yml
JelleZijlstra: Good idea!

.github/workflows/publish.yml
AlexWaygood: Our existing workflow to test our packaging is run on every PR:

https://github.com/python/typing_extensions/blob/479ac33a0c97ea91c8177f6b6887ff7a9fb9970d/.github/workflows/package.yml#L1-L8

Relatedly, should that existing workflow be deleted as part of this PR? Or is it still relevant?

JelleZijlstra: Oh right, I saw that last night. I now removed package.yml and given that the typing-extensions test suite is extremely fast, it seems fine to keep running it on every PR.

.github/workflows/publish.yml
JelleZijlstra: Added a script that validates this plus a few other aspects of pyproject.toml.

.github/workflows/publish.yml
srittau: I still feel uncomfortable running this workflow on PRs for the reasons outlined above. I'd prefer to just run the test workflow and only run this workflow on the final merge to main.

scripts/check_package.py
AlexWaygood, line 1: We could also consider adding a separate workflow that runs https://github.com/abravalheri/validate-pyproject as a pre-commit hook. There's a risk it might pull in `typing_extensions` as a dependency somewhere, so I think it would need to be isolated as its own job, or we'd pollute the environment with an installed version of ourselves

.github/workflows/publish.yml
AlexWaygood: But then on the other hand, if we don't run this workflow on PRs, it increases the risk that packaging bugs like https://github.com/python/typing_extensions/issues/61 are introduced in PRs.

I still feel like I'd prefer to do the packaging _tests_ on every PR for that reason, given they only take 7s from start to finish. Maybe a compromise could be to put the `publish-to-pypi` job in a workflow by itself, and have that workflow only run on `push` events?

scripts/check_package.py
JelleZijlstra: Running that in a separate job sounds useful; we can do that separately though.

.github/workflows/publish.yml
srittau: > I still feel uncomfortable running this workflow on PRs for the reasons outlined above. I'd prefer to just run the test workflow and only run this workflow on the final merge to main.

To clarify: I don't mean this comment as a blocker! Overall I'm +1 on this PR. If you think it's better to do it this way, please do.

AlexWaygood (approved):
LGTM

.github/workflows/publish.yml
JelleZijlstra: The current tests in this PR wouldn't have caught #61, since they rely only on the installed version of typing_extensions.py and use the test helper files from the repo. I suppose we could add a step that manually unzips the sdist and runs tests on the files within it.

I think Sebastian's hesitations align with the recommendations in https://docs.pypi.org/trusted-publishers/security-model/ so I'll refactor with that in mind. I'm thinking:

- One workflow that runs on all PRs that builds the sdist and wheel, tests them, and uploads them as GH Actions artifacts.
- One workflow that runs only on releases that downloads the artifacts and uploads them to PyPI.

Once we have this merged, I plan to make an alpha release (I think it'd be 4.11a1) just to test the workflow. Once it works well and all of Alex's PRs are merged, we can release an RC for the next feature release.

hauntsaninja: For mypy I set up https://github.com/mgedmin/check-manifest that catches "this is or isn't in sdist" issues

JelleZijlstra: I think running in separate workflows is not practical, since actions/download-artifact doesn't provide a simple way to share the artifacts (https://github.com/actions/download-artifact?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories - where would I get the run id?). Other repos I checked (pip-audit, hatch) also use a single workflow file.

However, the upload step is run in a separate environment, which I configured to require approval from one of the people on this PR. I think that should address the concerns about running this on PRs.

AlexWaygood: > I think running in separate workflows is not practical, since actions/download-artifact doesn't provide a simple way to share the artifacts ([actions/download-artifact#download-artifacts-from-other-workflow-runs-or-repositories](https://github.com/actions/download-artifact?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories) - where would I get the run id?). Other repos I checked (pip-audit, hatch) also use a single workflow file.

Do we need to share the artifact -- couldn't we just do a fresh build in the separate workflow? I think it's pretty quick, no?

JelleZijlstra:
I think this is ready to merge now, if someone wants to approve again.

.github/workflows/publish.yml
JelleZijlstra: We could but then we wouldn't be testing the exact artifact we're releasing. Sure, they *should* be the same, but what if they're not?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,147 @@
...
+          echo "::notice::Installing source distribution: $path_to_file"
+          python -m pip install --user $path_to_file
+          python -m pip list
+      - name: Run typing_extensions tests against installed package
+        run: rm src/typing_extensions.py && python src/test_typing_extensions.py
+
+  publish-to-pypi:
+    name: >-
+      Publish Python distribution to PyPI
+    if: github.event_name == 'release' # only publish to PyPI on releases
```
AlexWaygood, line 127: I think this needs to be dollarised, or it might always evaluate to true:

```suggestion
    if: ${{ github.event_name == 'release' }} # only publish to PyPI on releases
```

AlexWaygood: hmm, maybe this isn't _necessary_. But I think better to be safe than sorry?

JelleZijlstra: Seems to work on Black undollarized: https://github.com/psf/black/blob/9c1fd463e1af2568c3465d8e93963208eb6b9201/.github/workflows/pypi_upload.yml#L102

.github/workflows/publish.yml
AlexWaygood: Oh I see your point. Right, so we'd need to duplicate all the test jobs in the separate workflow. Okay, this seems reasonable then :-)

.github/workflows/publish.yml
AlexWaygood: Yes, apparently it only needs to be dollarised if it starts with `!`, which feels very strange and fragile but hey. https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#overview

AlexWaygood (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,147 @@
...
+
+jobs:
+  build:
+    name: Build distribution
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
srittau, line 28:
```suggestion
        uses: actions/setup-python@v5
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,147 @@
...
+  test-wheel:
+    name: Test wheel
+    needs:
+      - build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
srittau, line 56:
```suggestion
        uses: actions/setup-python@v5
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,147 @@
...
+          name: python-package-distributions
+          path: dist/
+      - name: Install wheel
+        run: |
+          export path_to_file=$(find dist -type f -name "typing_extensions-*.whl")
+          echo "::notice::Installing wheel: $path_to_file"
+          python -m pip install --user $path_to_file
+          python -m pip list
+      - name: Run typing_extensions tests against installed package
+        run: rm src/typing_extensions.py && python src/test_typing_extensions.py
```
srittau, line 71: Just an idea: To make it more robust against future changes, maybe copy the `test_*` script to `/tmp` and delete all of `src`?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,147 @@
...
+  test-sdist:
+    name: Test source distribution
+    needs:
+      - build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
srittau, line 82:
```suggestion
        uses: actions/setup-python@v5
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,147 @@
...
+          name: python-package-distributions
+          path: dist/
+      - name: Install source distribution
+        run: |
+          export path_to_file=$(find dist -type f -name "typing_extensions-*.tar.gz")
+          echo "::notice::Installing source distribution: $path_to_file"
+          python -m pip install --user $path_to_file
+          python -m pip list
+      - name: Run typing_extensions tests against installed package
+        run: rm src/typing_extensions.py && python src/test_typing_extensions.py
```
srittau, line 122: See comment above.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,147 @@
...
+    permissions:
+      id-token: write # IMPORTANT: mandatory for trusted publishing
+
+    steps:
+      - name: Download all the dists
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution to PyPI
```
srittau, line 146: Possibly add a step before this to assert that exactly one wheel and one sdist have been built?

srittau (changes requested):

.github/workflows/publish.yml
srittau: Something like:

```
- name: Ensure exactly one sdist and one wheel have been downloaded
  run:  test $(ls *.tar.gz | wc -l) = 1 && test $(ls *.whl | wc -l) = 1

.github/workflows/publish.yml
JelleZijlstra: That's not enough because it relies on a few helper files. I think we'll have to change this either way if we ever change our file layout.

.github/workflows/publish.yml
JelleZijlstra: Good idea, done

srittau (approved):

Issue #396, opened 2024-05-16
PEP-696 specialisation tests from CPython fail


AlexWaygood:
We've fixed a number of issues with our PEP-696 backport in recent days, and we're now a lot closer to CPython's implementation on Python 3.13. We're still missing 5 specialisation-related tests from CPython's test suite, however, and they all fail using the `typing_extensions` implementation on Python 3.12:

<details>
<summary>Missing tests</summary>

```diff
diff --git a/src/test_typing_extensions.py b/src/test_typing_extensions.py
index fa04e59..0ab1e7f 100644
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -6402,6 +6402,26 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         class A(Generic[Unpack[Ts]]): ...
         Alias = Optional[Unpack[Ts]]
 
+    def test_typevartuple_specialization(self):
+        T = TypeVar("T")
+        Ts = TypeVarTuple('Ts', default=Unpack[Tuple[str, int]])
+        self.assertEqual(Ts.__default__, Unpack[Tuple[str, int]])
+        class A(Generic[T, Unpack[Ts]]): ...
+        self.assertEqual(A[float].__args__, (float, str, int))
+        self.assertEqual(A[float, range].__args__, (float, range))
+        self.assertEqual(A[float, *tuple[int, ...]].__args__, (float, *tuple[int, ...]))
+
+    def test_typevar_and_typevartuple_specialization(self):
+        T = TypeVar("T")
+        U = TypeVar("U", default=float)
+        Ts = TypeVarTuple('Ts', default=Unpack[Tuple[str, int]])
+        self.assertEqual(Ts.__default__, Unpack[Tuple[str, int]])
+        class A(Generic[T, U, Unpack[Ts]]): ...
+        self.assertEqual(A[int].__args__, (int, float, str, int))
+        self.assertEqual(A[int, str].__args__, (int, str, str, int))
+        self.assertEqual(A[int, str, range].__args__, (int, str, range))
+        self.assertEqual(A[int, str, *tuple[int, ...]].__args__, (int, str, *tuple[int, ...]))
+
     def test_no_default_after_typevar_tuple(self):
         T = TypeVar("T", default=int)
         Ts = TypeVarTuple("Ts")
@@ -6487,6 +6507,34 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
[...]

PR #397, opened 2024-05-16
Backport PEP-696 specialisation on Python >=3.11.1

AlexWaygood:
This is a fix for #396 on Python 3.11 and Python 3.12. Unfortunately, the `__typing_prepare_subst__` hook doesn't exist on Python 3.10 and lower, so I'm not sure it's possible to backport it on older versions of Python without monkeypatching `typing._GenericAlias.__getitem__`, which seems like a bad idea. So maybe we should just document that not all features of PEP-696 are guaranteed to work on Python <=3.10?

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -6402,6 +6402,10 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         class A(Generic[Unpack[Ts]]): ...
         Alias = Optional[Unpack[Ts]]
 
+    @skipIf(
+        sys.version_info < (3, 11),
+        "Not yet backported for older versions of Python"
+    )
     def test_typevartuple_specialization(self):
         T = TypeVar("T")
         Ts = TypeVarTuple('Ts', default=Unpack[Tuple[str, int]])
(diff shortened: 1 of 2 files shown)


src/test_typing_extensions.py
```diff
@@ -6402,6 +6402,34 @@ def test_typevartuple(self):
         class A(Generic[Unpack[Ts]]): ...
         Alias = Optional[Unpack[Ts]]
 
+    @skipIf(
+        sys.version_info < (3, 11, 1),
+        "Not yet backported for older versions of Python"
+    )
```
AlexWaygood, line 6408: I verified locally that these tests pass with Python 3.11.1 but not with Python 3.11.0...

JelleZijlstra: It seems like the failures are just that we're missing the defaults from `__args__`, which doesn't seem too bad. (It may cause problems for introspection but shouldn't usually break creation of types that are normally in annotations.) I'm fine leaving those out for now.

JelleZijlstra (approved):

src/test_typing_extensions.py
AlexWaygood: Yes, the test failures "just" indicate incorrect behaviour rather than crashes

PR #398, opened 2024-05-16
Prepare releaes 4.12.0a1

JelleZijlstra:


hauntsaninja (approved):

srittau:
I think the test I suggested needs to look into the dist directory, not the current one.

PR #402, opened 2024-05-16
Prepare release 4.12.0rc1

JelleZijlstra:


CHANGELOG.md
```diff
@@ -1,29 +1,37 @@
...
+    3.13.0b1 and newer.
+  - Backport CPython PR [#118774](https://github.com/python/cpython/pull/118774),
+    allowing type parameters without default values to follow those with
+    default values in some type parameter lists. Patch by Alex Waygood,
+    backporting a CPython PR by Jelle Zijlstra.
+  - It is now disallowed to use a `TypeVar` with a default value after a
+    `TypeVarTuple` in a type parameter list. This matches the CPython
+    implementation of PEP 696 on Python 3.13+.
+  - Fix bug in PEP-696 implementation where default values for `ParamSpec`s
+    would be cast to tuples if a list was provided as the default value.
```
AlexWaygood, line 33: Re-reading my own changelog entry, this is possibly slightly more grammatical:

```suggestion
  - Fix bug in PEP-696 implementation where a default value for a `ParamSpec`
    would be cast to a tuple if a list was provided.
```

CHANGELOG.md
```diff
@@ -46,6 +54,12 @@ well, release 4.12.0rc1 will follow soon.
   `typing.ContextManager` and `typing.AsyncContextManager` on Python
   3.13+.
 - Backport `types.CapsuleType` from Python 3.13.
+- Releases are now made using [Trusted Publishers](https://docs.pypi.org/trusted-publishers/), improving the security of the release process. Patch by Jelle Zijlstra.
```
AlexWaygood, line 57: Long line:

```suggestion
- Releases are now made using [Trusted Publishers](https://docs.pypi.org/trusted-publishers/)
  improving the security of the release process. Patch by Jelle Zijlstra.
```

AlexWaygood (commented):
Thanks! I didn't add a changelog for https://github.com/python/typing_extensions/pull/397 -- not sure if we should or not, given that it only works on Python 3.11.1+? It might be better to leave it as an undocumented feature, not sure

AlexWaygood (approved):

hauntsaninja (approved):

PR #403, opened 2024-05-16
Add security documentation

### LINKED ISSUE #393: Improving the security of typing-extensions
typing-extensions is among the most widely used Python packages, and we should make sure we keep it secure.

I'd like to propose improvements in a few areas:

### Better release process

The current release process involves the release manager (in practice, me) building an sdist and wheel on their laptop and uploading it to PyPI. This opens up a few possible issues: the release manager's laptop could be compromised; they could make a mistake in what files to upload; they could sneakily manipulate the contents of the files before they are uploaded.

We should instead use https://docs.pypi.org/trusted-publishers/ to publish directly from a GitHub Action to PyPI. This removes the release manager's machine from the loop and ensures the release matches what is in the repo. (Recall that the recent xz backdoor involved a tarball with contents different from the repo.)

### Manage who has access

To reduce the risk of account compromise, we should limit access to the repo. I would like to propose:

* Remove commit (and PyPI maintainer) access from those who have not used it recently
* Ensure that every permission level (e.g., admin) has at least two people in it, so we don't rely on a single person who might disappear or lose interest

### Ensure code is reviewed

We should ensure all code in the repo is reviewed. I would like to propose:

* Require reviews for all pull requests
* Require pull requests for all changes (since PRs have more visibility than direct commits to the repo)

### Keep the repo simple

The less surface area, the less the chance of issues. We should commit to:

* Never add any non-stdlib dependencies to this repo
* Never add any non-Python code (e.g., C extensions)

---

I am happy to implement all of these if other maintainers agree.

JelleZijlstra:
Closes #393.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1234,4 +1234,5 @@ future:
 
 If you have any feedback on our security process, please `open an issue
 <https://github.com/python/typing_extensions/issues/new>`__. To report
-an issue privately, email security@python.org.
+an issue privately, use `GitHub's private reporting feature
+<https://github.com/python/typing_extensions/security>`__.
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1229,7 +1229,7 @@ future:
 
 * ``typing_extensions`` will never include any native extensions, only
   pure Python code.
-* ``typing_extensions`` will not depend on any third-party dependencies.
+* ``typing_extensions`` will not have any third-party dependencies.
 * We will follow best practices for a secure release process.
 
 If you have any feedback on our security process, please `open an issue
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1216,3 +1216,22 @@ versions of Python, but all are listed here for completeness.
    See :py:func:`typing.no_type_check_decorator`.
 
    .. versionadded:: 4.7.0
+
+Security
+--------
+
+``typing_extensions`` is among the most widely used packages in the
+Python ecosystem. Therefore, we take security seriously and strive
+to use a transparent, secure release process.
+
+We commit to the following in order to keep the package secure in the
+future:
+
+* ``typing_extensions`` will never include any native extensions, only
+  pure Python code.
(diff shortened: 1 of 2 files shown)


AlexWaygood (commented):
Should this be in `SECURITY.md`? (The guy in the talk we're in just now said that's where security researchers expect to find this kind of stuff)

hauntsaninja (approved):

JelleZijlstra:
@AlexWaygood I think we should also add SECURITY.md but I want to talk to someone (Seth?) to confirm the PSRT is OK with dealing with hypothetical security vulnerabilities in typing-extensions.

AlexWaygood (approved):
I think it would probably be fine to add an incomplete `SECURITY.md` file (with this information in it) and extend it later. But this is also fine!

hauntsaninja:
Let's merge this. We should also think about doing https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability

JelleZijlstra:
I enabled private vulnerability reporting and added a SECURITY.md pointing to it.

hauntsaninja (approved):

doc/index.rst
```diff
@@ -1216,3 +1216,23 @@ versions of Python, but all are listed here for completeness.
...
+``typing_extensions`` is among the most widely used packages in the
+Python ecosystem. Therefore, we take security seriously and strive
+to use a transparent, secure release process.
+
+We commit to the following in order to keep the package secure in the
+future:
+
+* ``typing_extensions`` will never include any native extensions, only
+  pure Python code.
+* ``typing_extensions`` will not depend on any third-party dependencies.
```
AlexWaygood, line 1232: "depend" used twice in the same sentence:

```suggestion
* ``typing_extensions`` will not have any third-party dependencies.
```

AlexWaygood (approved):

PR #407, opened 2024-05-22
Make sure `isinstance(typing_extensions.ParamSpec("P"), typing.TypeVar)` is unaffected by `sys.setprofile()`

### LINKED ISSUE #318: isinstance return different result with sys.setprofile
python: 3.8.5

```python
import sys
from typing import TypeVar


def run():
    sys.modules.pop("typing_extensions", None)
    return isinstance(ParamSpec("P"), TypeVar)


def trace_call(frame, event, arg):
   return trace_call


if __name__ == "__main__":
    print(run()) # True

    sys.setprofile(trace_call)
    print(run()) # False??

    sys.setprofile(None)
    print(run()) #  True
```

AlexWaygood:
Fixes #318. The workaround here is to avoid the `super()` call -- the `super()` call accesses the `__class__` variable from the `__init__` method, which due to a CPython bug on Python <=3.10 means that `__class__` is removed from `ParamSpec.__dict__` if a profiling function has been set.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -92,7 +92,9 @@ jobs:
           export path_to_file=$(find dist -type f -name "typing_extensions-*.tar.gz")
           echo "::notice::Unpacking source distribution: $path_to_file"
           tar xzf $path_to_file -C dist/
+          ls -a
           cd ${path_to_file%.tar.gz}
+          ls -a
           python src/test_typing_extensions.py
 
   test-sdist-installed:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -92,8 +92,8 @@ jobs:
           export path_to_file=$(find dist -type f -name "typing_extensions-*.tar.gz")
           echo "::notice::Unpacking source distribution: $path_to_file"
           tar xzf $path_to_file -C dist/
-          cd ${path_to_file%.tar.gz}
-          python src/test_typing_extensions.py
+          cd ${path_to_file%.tar.gz}/src
+          python test_typing_extensions.py
 
   test-sdist-installed:
     name: Test installed source distribution
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -92,8 +92,8 @@ jobs:
           export path_to_file=$(find dist -type f -name "typing_extensions-*.tar.gz")
           echo "::notice::Unpacking source distribution: $path_to_file"
           tar xzf $path_to_file -C dist/
-          cd ${path_to_file%.tar.gz}/src
-          python test_typing_extensions.py
+          cd ${path_to_file%.tar.gz}
+          python src/test_typing_extensions.py
 
   test-sdist-installed:
     name: Test installed source distribution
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -92,8 +92,8 @@ jobs:
           export path_to_file=$(find dist -type f -name "typing_extensions-*.tar.gz")
           echo "::notice::Unpacking source distribution: $path_to_file"
           tar xzf $path_to_file -C dist/
-          cd ${path_to_file%.tar.gz}
-          python src/test_typing_extensions.py
+          cd ${path_to_file%.tar.gz}/src
+          python test_typing_extensions.py
```
AlexWaygood, line 96: I had to make this change or the subprocess wasn't able to `import typing_extensions`. It was raising `ModuleNotFoundError`

src/test_typing_extensions.py
```diff
@@ -5020,6 +5020,48 @@ def test_eq(self):
...
+        # Run this in an isolated process or it pollutes the environment
+        # and makes other tests fail:
+        try:
+            proc = subprocess.run(
+                [sys.executable, "-c", code], check=True, capture_output=True, text=True,
+            )
+        except subprocess.CalledProcessError as exc:
+            print("stdout", exc.stdout, sep="\n")
+            print("stderr", exc.stderr, sep="\n")
+            raise
```
AlexWaygood, line 5054: I tried running this in an isolated scope using `exec()`, but it was still causing some pollution of the global environment that was causing other, unrelated tests to fail. This seems to be the only way of doing a test with total interpreter isolation.

.github/workflows/publish.yml
JelleZijlstra: With your change it uses the copy of `typing_extensions` in the repo instead of the installed one; that is incorrect. We shouldn't do this.

AlexWaygood: I see. Do you have any idea why the subprocess wasn't able to import `typing_extensions` prior to this change, but only for this CI job? I don't fully understand that (see https://github.com/python/typing_extensions/actions/runs/9197903212/job/25299411282?pr=407)

JelleZijlstra: Oh actually, this might be fine, since in this job we're not attempting to test an installed version of typing_extensions. I'm not actually sure why this doesn't work in CI though. Possibly the test itself is picking up the wrong copy of `typing_extensions`.

AlexWaygood: I'm actually not sure how the pre-existing tests currently pass _without_ this? I think it's correct? Otherwise we unpack the version of `typing_extensions` we want into an `src/` directory, but never `cd` into that `src` directory, right?

JelleZijlstra: Yes, that's what it looks like.

JelleZijlstra (approved):

JelleZijlstra:
The 4.12.0 release should go out today, do you think it's better to merge this in now or wait for the next release? The change seems pretty safe so I think it's fine to merge now in the RC phase.

AlexWaygood:
> The 4.12.0 release should go out today, do you think it's better to merge this in now or wait for the next release? The change seems pretty safe so I think it's fine to merge now in the RC phase.

Oh shoot, sorry -- I merged before seeing this.

I think it's okay to put it into the next release; I agree that this change seems pretty safe. But I also don't think that this issue affects many people, so I would also be fine with cutting the release based on the commit prior to this.

PR #412, opened 2024-05-25
Support my PEP 649 branch

JelleZijlstra:
Tests now pass with python/cpython#119361.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -64,13 +64,14 @@ skip_if_py313_beta_1 = skipIf(
 )
 
 ANN_MODULE_SOURCE = '''\
+import sys
 from typing import List, Optional
 from functools import wraps
 
 try:
     __annotations__[1] = 2
 except NameError:
-    pass  # 3.14+
+    assert sys.version_info >= (3, 14)
 
 class C:
 
@@ -83,7 +84,7 @@ class M(type):
     try:
         __annotations__['123'] = 123
     except NameError:
-        pass  # 3.14+
+        assert sys.version_info >= (3, 14)
     o: type = object
 
 (pars): bool = True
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -67,7 +67,10 @@
 from typing import List, Optional
 from functools import wraps
 
-__annotations__[1] = 2
+try:
+    __annotations__[1] = 2
+except NameError:
+    pass  # 3.14+
```
srittau, line 73: Maybe add an `assert system.version >= (3, 14)` inside the `except` block?

srittau (approved):

src/typing_extensions.py
```diff
@@ -942,7 +942,13 @@ def __new__(cls, name, bases, ns, *, total=True, closed=False):
...
 
             annotations = {}
-            own_annotations = ns.get('__annotations__', {})
+            if "__annotations__" in ns:
+                own_annotations = ns["__annotations__"]
+            elif "__annotate__" in ns:
+                # TODO: Use inspect.VALUE here, and make the annotations lazily evaluated
+                own_annotations = ns["__annotate__"](1)
+            else:
+                own_annotations = {}
```
AlexWaygood, line 951: You mentioned in https://github.com/python/cpython/pull/119361#discussion_r1615277717 that you'd be adding a helper function for this kind of thing at some point -- would it be worth waiting for that, since the first 3.14 alpha is some way off?

I can also see the advantages of getting `typing_extensions` working with your branch, though, so that you can start testing with the rest of the ecosystem that depends on `typing_extensions`. So I definitely don't want to block you merging this.

AlexWaygood (approved):

src/typing_extensions.py
JelleZijlstra: I do expect we'll need more changes later, but it's helpful to get this merged so I can test the rest of the ecosystem too.

PR #414, opened 2024-05-28
Switch from flake8 to ruff

AlexWaygood:
Obviously I work at Astral, so I'm biased -- but I do think using linting tools written in another language is a particularly good fit for us at `typing_extensions`. Most linters (including many flake8 plugins) can't really be installed into the same environment you'd want to use locally for testing `typing_extensions`, as it's pretty hard to find a linter written in Python that doesn't depend (directly or indirectly) on `typing_extensions`. If you then use the same environment to run tests as you used for running the linter, you're never sure whether you're actually testing your local copy of `typing_extensions` or the installed version in `site-packages`.

~~As well as switching from flake8 to ruff, this PR also deletes the `test-requirements.txt`; development dependencies are now listed in a `dev` extra in `pyproject.toml`.~~


pyproject.toml
```diff
@@ -60,3 +60,40 @@ email = "levkivskyi@gmail.com"
 [tool.flit.sdist]
 include = ["CHANGELOG.md", "README.md", "tox.ini", "*/*test*.py"]
 exclude = []
+
+[tool.ruff]
+line-length = 90
+target-version = "py38"
+fix = true
```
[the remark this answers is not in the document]
AlexWaygood:
```suggestion
```

srittau (approved):
LGTM, although I'd like @JelleZijlstra's opinion.

src/typing_extensions.py
```diff
@@ -413,12 +413,12 @@ def clear_overloads():
 OrderedDict = typing.OrderedDict
 Counter = typing.Counter
 ChainMap = typing.ChainMap
-Text = typing.Text
+Text = typing.Text  # noqa: UP019
```
JelleZijlstra, line 416: Can we just turn off the UP* rules that tell you about deprecated parts of typing? I don't think they're helpful for this project.

AlexWaygood: Fair point :) I'll make the change later

JelleZijlstra (approved):

PR #422, opened 2024-06-03
Avoid error if origin has a buggy __eq__

JelleZijlstra:
#419

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -6617,6 +6617,7 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         a4 = Callable[[Unpack[Ts]], T]
         self.assertEqual(a4.__args__, (Unpack[Ts], T))
 
+    @skip_if_py313_beta_1
     def test_generic_with_broken_eq(self):
         # See https://github.com/python/typing_extensions/pull/422 for context
         class BrokenEq(abc.ABCMeta):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -6620,7 +6620,7 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
     @skip_if_py313_beta_1
     def test_generic_with_broken_eq(self):
         # See https://github.com/python/typing_extensions/pull/422 for context
-        class BrokenEq(abc.ABCMeta):
+        class BrokenEq(type):
             def __eq__(self, other):
                 if other is typing_extensions.Protocol:
                     raise TypeError("I'm broken")
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -6617,6 +6617,21 @@ class TypeVarLikeDefaultsTests(BaseTestCase):
         a4 = Callable[[Unpack[Ts]], T]
         self.assertEqual(a4.__args__, (Unpack[Ts], T))
 
+    def test_generic_with_broken_eq(self):
+        # See https://github.com/python/typing_extensions/pull/422 for context
+        class BrokenEq(abc.ABCMeta):
+            def __eq__(self, other):
+                if other is typing_extensions.Protocol:
+                    raise TypeError("I'm broken")
+                return False
+
+        class G(Generic[T], metaclass=BrokenEq):
+            pass
+
+        alias = G[int]
+        self.assertIs(get_origin(alias), G)
+        self.assertEqual(get_args(alias), (int,))
+
     @skipIf(
         sys.version_info < (3, 11, 1),
         "Not yet backported for older versions of Python"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+- Fix regression in v4.12.0 where specialization of certain
+  generics with an overridden `__eq__` method would raise errors.
+  Patch by Jelle Zijlstra.
+
 # Release 4.12.1 (June 1, 2024)
 
 - Preliminary changes for compatibility with the draft implementation
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2958,7 +2958,8 @@ def _has_generic_or_protocol_as_origin() -> bool:
     except AttributeError:
         return False  # err on the side of leniency
     else:
-        # If we somehow get invoked from outside typing.py, also err on the side of leniency
+        # If we somehow get invoked from outside typing.py,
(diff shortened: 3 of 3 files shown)


AlexWaygood:
This is a good solution. Something we could also do is to check that `frame.f_globals.get("__name__") == "typing"` before we check the value of `origin`. We only specifically care about an `origin` local variable if it's being set in a frame from `typing.py`, not if it's being set in a frame from `site-packages/dirty_equals/_base.py`

JelleZijlstra:
Not sure that applies here, since in the pydantic stacktrace we're still being invoked from `typing.py`, it's just that `origin` is some funny object that has a broken `__eq__`. However, it's definitely *possible* that someone would invoke `typing_extensions. _collect_type_vars` directly without going through typing, so I'll add that check.

AlexWaygood:
Yes, I don't think it applies here, I'm just inclined at this point to err on the side of absolute caution when it comes to this hack :-)

Have you been able to figure out exactly why the `NameError` is occurring in the pydantic tests? `_length_repr` looks like it's defined [here](https://github.com/samuelcolvin/dirty-equals/blob/2c85486b29492977f82c850eadf1afc0e2a9434e/dirty_equals/_sequence.py#L257). My guess would be that something to do with the `_get_frame` hack means that the `__eq__` method can't find it in the globals for some reason, but I haven't been able to quickly find a minimised repro (and probably shouldn't spend too much time on it right now :/ )

AlexWaygood:
(To be clear I don't think we need to necessarily have a test for this one, since it's pretty obvious that this would fix it. But I'd still be curious to know exactly what the cause of this is, if we _can_ figure it out.)

JelleZijlstra:
Yeah, I don't understand the NameError either. I was able to reproduce the failure in a test:
```
+    def test_generic_with_broken_eq(self):
+        # See https://github.com/python/typing_extensions/pull/422 for context
+        class BrokenEq(abc.ABCMeta):
+            def __eq__(self, other):
+                if other is typing_extensions.Protocol:
+                    raise TypeError("I'm broken")
+                return False
+
+        class G(Generic[T], metaclass=BrokenEq):
+            pass
+
+        alias = G[int]
+        self.assertIs(get_origin(alias), G)
+        self.assertEqual(get_args(alias), (int,))
+
```

```
  File "test_typing_extensions.py", line 6631, in test_generic_with_broken_eq
    alias = G[int]
  File "/Users/jelle/.pyenv/versions/3.8.16/lib/python3.8/typing.py", line 261, in inner
    return func(*args, **kwds)
  File "/Users/jelle/.pyenv/versions/3.8.16/lib/python3.8/typing.py", line 898, in __class_getitem__
    return _GenericAlias(cls, params)
  File "/Users/jelle/.pyenv/versions/3.8.16/lib/python3.8/typing.py", line 672, in __init__
    self.__parameters__ = _collect_type_vars(params)
  File "/Users/jelle/py/typing_extensions/src/typing_extensions.py", line 2994, in _collect_type_vars
    enforce_default_ordering = _has_generic_or_protocol_as_origin()
  File "/Users/jelle/py/typing_extensions/src/typing_extensions.py", line 2961, in _has_generic_or_protocol_as_origin
    return frame.f_locals.get("origin") in (
  File "test_typing_extensions.py", line 6625, in __eq__
    raise TypeError("I'm broken")
TypeError: I'm broken
```

Though it only reproduces on 3.8 for some reason.

AlexWaygood:
That test looks good enough to me; it's raising the `TypeError` in the right place in the stack trace

> Though it only reproduces on 3.8 for some reason.

Yeah, and the same with the pydantic test failure, right? Maybe something about frames where it wouldn't find globals properly got fixed in Python 3.9. Who knows.

JelleZijlstra:
Actually the NameError appears to be because the error happens while `dirty_equals` is being imported. The code that triggers the error is on line 215 of `_sequence.py`, and `_length_repr` is on line 257, so it doesn't exist yet.

JelleZijlstra:
I guess I found a CPython bug too.

src/test_typing_extensions.py
```diff
@@ -6617,6 +6617,21 @@ def test_allow_default_after_non_default_in_alias(self):
         a4 = Callable[[Unpack[Ts]], T]
         self.assertEqual(a4.__args__, (Unpack[Ts], T))
 
+    def test_generic_with_broken_eq(self):
```
AlexWaygood, line 6620: as a short-term fix until we fix the CPython bug:

```suggestion
    @skip_if_py313_beta_1
    def test_generic_with_broken_eq(self):
```

AlexWaygood (approved):

AlexWaygood:
If we want to be maximally paranoid, we could also make this change:

```diff
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2954,8 +2954,10 @@ if not _PEP_696_IMPLEMENTED:
 def _has_generic_or_protocol_as_origin() -> bool:
     try:
         frame = sys._getframe(2)
-    # not all platforms have sys._getframe()
-    except AttributeError:
+    # - Catch AttributeError: not all Python implementations have sys._getframe()
+    # - Catch ValueError: maybe we're called from an unexpected module
+    #   and the call stack isn't deep enough
+    except (AttributeError, ValueError):
```

JelleZijlstra:
```
% python -c 'from typing_extensions import _has_generic_or_protocol_as_origin; _has_generic_or_protocol_as_origin()'
  File "<string>", line 1, in <module>
  File "/Users/jelle/py/typing_extensions/src/typing_extensions.py", line 2956, in _has_generic_or_protocol_as_origin
    frame = sys._getframe(2)
            ^^^^^^^^^^^^^^^^
ValueError: call stack is not deep enough
```

I guess I get to fix another critical bug.

src/test_typing_extensions.py
```diff
@@ -6617,6 +6617,22 @@ def test_allow_default_after_non_default_in_alias(self):
         a4 = Callable[[Unpack[Ts]], T]
         self.assertEqual(a4.__args__, (Unpack[Ts], T))
 
+    @skip_if_py313_beta_1
+    def test_generic_with_broken_eq(self):
+        # See https://github.com/python/typing_extensions/pull/422 for context
+        class BrokenEq(abc.ABCMeta):
```
AlexWaygood, line 6623: not specific to the `abc` module; any metaclass will do:

```suggestion
        class BrokenEq(type):
```

AlexWaygood (approved):
LGTM

PR #423, opened 2024-06-05
Add typing_extensions.get_annotations

### LINKED ISSUE #409: Add `inspect.get_annotations` backport
See https://discuss.python.org/t/pep-649-deferred-evaluation-of-annotations-tentatively-accepted/21331/62

JelleZijlstra:
Fixes #409

This is based on the implementation in python/cpython#119891. The version that eventually lands in CPython may look a bit different, but should be compatible with what's in this PR.


src/_inspect_stock_annotations.py
AlexWaygood, line 1: In #151, the consensus seemed to be that we shouldn't add new "data files" to the src directory for usage in tests, because these will end up in the site-packages directory for people who've installed typing_extensions. Should we just store these as strings in test_typing_extensions.py, like we did for the data files in that PR?

srittau: It might also be time to split the test file, as it's approaching 10k lines.

AlexWaygood: Maybe we could have an `src/test_typing_extensons` directory: then it's only a single importable module in `site-packages`, but we'd be able to split it up so that it doesn't get too huge.

src/typing_extensions.py
```diff
@@ -3599,6 +3602,147 @@ def __eq__(self, other: object) -> bool:
...
+# some other variation: in any case, inspect.get_annotations
+# will continue to exist and will gain a `format` parameter.
+_PEP_649_OR_749_IMPLEMENTED = (
+    hasattr(inspect, 'get_annotations')
+    and inspect.get_annotations.__kwdefaults__ is not None
+    and "format" in inspect.get_annotations.__kwdefaults__
+)
+
+
+class Format(int, enum.Enum):
```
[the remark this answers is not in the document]
JelleZijlstra: I thought it was only added in 3.11, but that's StrEnum; we've always had IntEnum.
```suggestion
class Format(enum.IntEnum):
```

src/_inspect_stock_annotations.py
JelleZijlstra: I like the idea of making a test directory. I'll send a separate PR for that.

src/typing_extensions.py
```diff
@@ -3599,6 +3602,145 @@ def __eq__(self, other: object) -> bool:
...
+        The purpose of this backport is to allow users who would like to use
+        FORWARDREF or SOURCE semantics once PEP 649 is implemented, but who also
+        want to support earlier Python versions, to simply write:
+
+            typing_extensions.get_annotations(obj, format=Format.FORWARDREF)
+
+        """
+        format = Format(format)
+
+        if eval_str and format != Format.VALUE:
```
AlexWaygood, line 3664:
```suggestion
        if eval_str and format is not Format.VALUE:
```

src/typing_extensions.py
```diff
@@ -3599,6 +3602,145 @@ def __eq__(self, other: object) -> bool:
...
+            return {}
+
+        if not isinstance(ann, dict):
+            raise ValueError(f"{obj!r}.__annotations__ is neither a dict nor None")
+
+        if not ann:
+            return {}
+
+        if not eval_str:
+            if format == Format.SOURCE:
```
AlexWaygood, line 3715:
```suggestion
            if format is Format.SOURCE:
```

AlexWaygood (commented):
(https://docs.python.org/3/howto/enum.html#comparisons)

src/_inspect_stock_annotations.py
JelleZijlstra: I tried this and I think it's not worth it. I like to simply run `python test_typing_extensions.py` to run the test suite; that would no longer work with a package. I'll move the test files from this PR into `exec()` calls instead.

src/typing_extensions.py
```diff
@@ -3599,6 +3602,147 @@ def __eq__(self, other: object) -> bool:
...
+
+        if unwrap is not None:
+            while True:
+                if hasattr(unwrap, '__wrapped__'):
+                    unwrap = unwrap.__wrapped__
+                    continue
+                if isinstance(unwrap, functools.partial):
+                    unwrap = unwrap.func
+                    continue
+                break
```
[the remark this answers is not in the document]
JelleZijlstra: Not sure. Looks like unwrap() doesn't handle functools.partial specially, though unlike this code it deals with cycles.

src/typing_extensions.py
```diff
@@ -3599,6 +3602,147 @@ def __eq__(self, other: object) -> bool:
...
+                obj_globals = unwrap.__globals__
+
+        if globals is None:
+            globals = obj_globals
+        if locals is None:
+            locals = obj_locals
+
+        return_value = {key:
+            value if not isinstance(value, str) else eval(value, globals, locals)
+            for key, value in ann.items() }
```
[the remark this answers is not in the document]
AlexWaygood: Filed https://github.com/python/cpython/pull/120270

src/typing_extensions.py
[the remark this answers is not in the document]
AlexWaygood: It's pretty odd; no CPython tests fail if I make this change:

```diff
diff --git a/Lib/inspect.py b/Lib/inspect.py
index 2b7f8bec482..4f9c43d9fc7 100644
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -235,13 +235,11 @@ def get_annotations(obj, *, globals=None, locals=None, eval_str=False):
             if module:
                 obj_globals = getattr(module, '__dict__', None)
         obj_locals = dict(vars(obj))
-        unwrap = obj
     elif isinstance(obj, types.ModuleType):
         # module
         ann = getattr(obj, '__annotations__', None)
         obj_globals = getattr(obj, '__dict__')
         obj_locals = None
-        unwrap = None
     elif callable(obj):
         # this includes types.Function, types.BuiltinFunctionType,
         # types.BuiltinMethodType, functools.partial, functools.singledispatch,
@@ -249,7 +247,6 @@ def get_annotations(obj, *, globals=None, locals=None, eval_str=False):
         ann = getattr(obj, '__annotations__', None)
         obj_globals = getattr(obj, '__globals__', None)
         obj_locals = None
-        unwrap = obj
     else:
         raise TypeError(f"{obj!r} is not a module, class, or callable.")
 
@@ -265,17 +262,8 @@ def get_annotations(obj, *, globals=None, locals=None, eval_str=False):
     if not eval_str:
         return dict(ann)
 
-    if unwrap is not None:
-        while True:
-            if hasattr(unwrap, '__wrapped__'):
-                unwrap = unwrap.__wrapped__
-                continue
-            if isinstance(unwrap, functools.partial):
-                unwrap = unwrap.func
-                continue
-            break
-        if hasattr(unwrap, "__globals__"):
-            obj_globals = unwrap.__globals__
+    if not isinstance(obj, types.ModuleType):
+        obj_globals = getattr(unwrap(obj), "__globals__", None) or obj_globals
```

You could consider that simplification here since we already import `inspect` in `typing_extensions`, though it obviously can't be applied to CPython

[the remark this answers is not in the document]
JelleZijlstra: That sounds like we're missing test coverage for `functools.partial`. I think we should mirror the CPython implementation as much as possible.

AlexWaygood (approved):
LGTM. https://github.com/python/typing_extensions/pull/423#discussion_r1627416573 can be fixed in a followup once we've sorted it out at cpython

PR #430, opened 2024-06-20
Add TypeExpr

### LINKED ISSUE #429: Add `typing_extensions.TypeExpr`
We should provide TypeExpr from PEP-747 in typing_extensions. @davidfstr are you interested in doing this? The runtime implementation is likely going to be very simple.

Once it's in typing-extensions, I may work on implementing TypeExpr support in pyanalyze since we have some internal use cases.

JelleZijlstra:
Fixes #429. cc @davidfstr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add `typing_extensions.TypeExpr` from PEP 747. Patch by
+  Jelle Zijlstra.
 - Add `typing_extensions.get_annotations`, a backport of
   `inspect.get_annotations` that adds features specified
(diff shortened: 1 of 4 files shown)


davidfstr:
Implementation and tests look good.

One capability of TypeExpr that appears to be missing from this implementation is support for the [explicit syntax](https://peps.python.org/pep-0747/#explicit-typeexpr-values) `TypeExpr(typx)` that just returns `typx` unchanged. I'm not sure if you wanted to add that capability in this initial PR or in a followup PR.

JelleZijlstra:
Thanks, I forgot about that; I'll add it in this PR.

davidfstr:
LGTM

srittau (approved):

PR #437, opened 2024-07-22
feat: add attrspec

suned:


JelleZijlstra:
Please first introduce your proposal on discuss.python.org, then write a PEP, then add it to typing-extensions.

Also, don't make unrelated formatting changes in the same PR; it makes it very hard to find your actual changes.

PR #442, opened 2024-08-08
Added support for Concatenate with Ellipsis

### LINKED ISSUE #110: Backport allowing passing an ellipsis as the last argument of Concatenate to 3.10

Daraan:
Fixes #110.

This PR adds support for Ellipsis in Concatenate for the Python 3.8-3.10 backports.

Does not fix:
- #48 : Using Ellipsis as argument for Callable with a generic..
  - <= 3.9: because of `typing.Callable` limitations
  - For 3.10 could go back to `typing_extentions._ConcatenateGenericAlias` but also tricky.

Changes:

- Changed `_concatenate_getitem` to reflect the Python3.11 typing module that allows the Ellipsis parameter.
- For Python 3.10 uses `typing._ConcatenateGenericAlias` and `_concatenate_getitem`
- For <=Python 3.9 the `typing_extensions._ConcatenateGenericAlias` and `_concatenate_getitem` like before
- For <3.9.2 uses a Dummy class that is temporarily swapped with an Ellipsis input
- Added tests

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -70,7 +70,7 @@ from typing_extensions import (
     TypeAlias,
     TypeAliasType,
     TypedDict,
-    TypeExpr,
+    TypeForm,
     TypeGuard,
     TypeIs,
     TypeVar,
@@ -5508,33 +5508,33 @@ class TypeIsTests(BaseTestCase):
             issubclass(int, TypeIs)
 
 
-class TypeExprTests(BaseTestCase):
+class TypeFormTests(BaseTestCase):
     def test_basics(self):
-        TypeExpr[int]  # OK
-        self.assertEqual(TypeExpr[int], TypeExpr[int])
+        TypeForm[int]  # OK
+        self.assertEqual(TypeForm[int], TypeForm[int])
 
-        def foo(arg) -> TypeExpr[int]: ...
-        self.assertEqual(gth(foo), {'return': TypeExpr[int]})
+        def foo(arg) -> TypeForm[int]: ...
+        self.assertEqual(gth(foo), {'return': TypeForm[int]})
 
     def test_repr(self):
-        if hasattr(typing, 'TypeExpr'):
+        if hasattr(typing, 'TypeForm'):
             mod_name = 'typing'
         else:
             mod_name = 'typing_extensions'
-        self.assertEqual(repr(TypeExpr), f'{mod_name}.TypeExpr')
-        cv = TypeExpr[int]
-        self.assertEqual(repr(cv), f'{mod_name}.TypeExpr[int]')
-        cv = TypeExpr[Employee]
-        self.assertEqual(repr(cv), f'{mod_name}.TypeExpr[{__name__}.Employee]')
-        cv = TypeExpr[Tuple[int]]
-        self.assertEqual(repr(cv), f'{mod_name}.TypeExpr[typing.Tuple[int]]')
+        self.assertEqual(repr(TypeForm), f'{mod_name}.TypeForm')
+        cv = TypeForm[int]
+        self.assertEqual(repr(cv), f'{mod_name}.TypeForm[int]')
+        cv = TypeForm[Employee]
+        self.assertEqual(repr(cv), f'{mod_name}.TypeForm[{__name__}.Employee]')
+        cv = TypeForm[Tuple[int]]
+        self.assertEqual(repr(cv), f'{mod_name}.TypeForm[typing.Tuple[int]]')
 
     def test_cannot_subclass(self):
         with self.assertRaises(TypeError):
-            class C(type(TypeExpr)):
+            class C(type(TypeForm)):
                 pass
         with self.assertRaises(TypeError):
-            class D(type(TypeExpr[int])):
+            class D(type(TypeForm[int])):
                 pass
 
     def test_call(self):
@@ -5546,24 +5546,24 @@ class TypeExprTests(BaseTestCase):
         ]
         for obj in objs:
             with self.subTest(obj=obj):
-                self.assertIs(TypeExpr(obj), obj)
+                self.assertIs(TypeForm(obj), obj)
 
         with self.assertRaises(TypeError):
-            TypeExpr()
+            TypeForm()
         with self.assertRaises(TypeError):
-            TypeExpr("too", "many")
+            TypeForm("too", "many")
 
     def test_cannot_init_type(self):
         with self.assertRaises(TypeError):
-            type(TypeExpr)()
+            type(TypeForm)()
         with self.assertRaises(TypeError):
-            type(TypeExpr[Optional[int]])()
+            type(TypeForm[Optional[int]])()
 
     def test_no_isinstance(self):
         with self.assertRaises(TypeError):
-            isinstance(1, TypeExpr[int])
+            isinstance(1, TypeForm[int])
         with self.assertRaises(TypeError):
-            issubclass(int, TypeExpr)
+            issubclass(int, TypeForm)
 
 
 class LiteralStringTests(BaseTestCase):
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -86,7 +86,7 @@ __all__ = [
     'Text',
     'TypeAlias',
     'TypeAliasType',
-    'TypeExpr',
+    'TypeForm',
     'TypeGuard',
     'TypeIs',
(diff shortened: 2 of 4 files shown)


src/typing_extensions.py
```diff
@@ -1795,28 +1795,62 @@ def __parameters__(self):
             return tuple(
                 tp for tp in self.__args__ if isinstance(tp, (typing.TypeVar, ParamSpec))
             )
+# 3.10+
+else:
+    _ConcatenateGenericAlias = typing._ConcatenateGenericAlias
 
+# 3.10.2+
+if sys.version_info[:3] >= (3, 10, 2):
```
JelleZijlstra, line 1803:
```suggestion
if sys.version_info >= (3, 10, 2):
```

src/typing_extensions.py
```diff
@@ -1795,28 +1795,62 @@ def __parameters__(self):
...
+        if not (parameters[-1] is ... or isinstance(parameters[-1], ParamSpec)):
+            raise TypeError("The last parameter to Concatenate should be a "
+                            "ParamSpec variable or ellipsis.")
+        msg = "Concatenate[arg, ...]: each arg must be a type."
+        parameters = (*(typing._type_check(p, msg) for p in parameters[:-1]),
+                      parameters[-1])
+        return _ConcatenateGenericAlias(self, parameters)
+
+# 3.11+; Concatenate does not accept ellipsis in 3.10
+if hasattr(typing, 'Concatenate') and sys.version_info[:2] >= (3, 11):
```
JelleZijlstra, line 1851:
```suggestion
if hasattr(typing, 'Concatenate') and sys.version_info >= (3, 11):
```

src/test_typing_extensions.py
```diff
@@ -5351,25 +5379,56 @@ def test_invalid_uses(self):
...
+            ):
+                Concatenate[1, ..., P]
+
+    @skipUnless(TYPING_3_11_0, "Cannot be backported to <=3.9"
+                               "Cannot use ... with 3.10 typing._ConcatenateGenericAlias")
+    def test_alias(self):
+        P = ParamSpec("P")
+        C1 = Callable[Concatenate[int, P], Any]
+        # Python <= 3.9 fails because parameters to generic types must be types.
+        # For Python 3.10 & typing._ConcatenateGenericAlias will
```
JelleZijlstra, line 5413: This sentence seems garbled, I'm not sure what you're trying to say.

src/test_typing_extensions.py
```diff
@@ -5351,25 +5379,56 @@ def test_invalid_uses(self):
 
         with self.assertRaisesRegex(
             TypeError,
-            'The last parameter to Concatenate should be a ParamSpec variable',
+            'The last parameter to Concatenate should be a ParamSpec variable or ellipsis',
         ):
             Concatenate[P, T]
 
+        # Cannot construct a Callable with Ellipsis in 3.8 as args must be types
+        if sys.version_info[:2] >= (3, 9, 2):
```
JelleZijlstra, line 5387:
```suggestion
        if sys.version_info >= (3, 9, 2):
```

JelleZijlstra (commented):
Thanks for these fixes!

Please avoid force pushing for future changes; it makes it harder to track the history of the code. Instead, just push new commits to your branch, and we'll squash them when we merge the PR.

src/test_typing_extensions.py
Daraan: Removed this block.

My idea was to note about these three cases

1. Cannot use with <= 3.9 because types need to be passed
2. Cannot use `typing_extensions.Concatenate`, `_concatenate_getitem`  with the original `typing._ConcatenateGenericAlias`
3. A full backport to `typing_extensions._ConcatenateGenericAlias` might be possible but is not implemented.

---

Btw. should the function be renamed? I was not sure how to name it and now realized test_alias is used elsewhere with different semantics.

src/test_typing_extensions.py
Daraan: Tried to keep it consistent with the other locations as I've seen is quite often. In this case good catch.

Daraan: Ah I see the behavior of sys.version_info was changed and hat to be sliced before.

src/test_typing_extensions.py
```diff
@@ -5377,6 +5390,21 @@ def test_valid_uses(self):
             self.assertEqual(C3.__origin__, C4.__origin__)
             self.assertNotEqual(C3, C4)
 
+        # Callable with Ellipsis cannot be constructed in 3.8 and below 3.9.2
+        if sys.version_info[:2] <= (3, 9, 2):
+            return
+
```
AlexWaygood, line 5396: There are obviously lots of situations in this file where we already disobey this rule (which arguably we should fix, IMO) -- but in general, it's good to put assertions that only pass on certain versions in separate test methods which are skipped on certain Python versions. If you do so, the test explicitly shows up as being "skipped" in the summary report when you run the tests, rather than the test silently behaving differently on different Python versions.

```suggestion
    @skipUnless(sys.version_info >= (3, 9, 3))
    def test_valid_uses_py39_plus(self):
        # Callable with Ellipsis cannot be constructed in 3.8 and below 3.9.2

```

src/test_typing_extensions.py
Daraan: I've now renamed the function I named `test_alias` before.

Daraan:
I've made a rework (moved the parts to ` _create_concatenate_alias`) that is now simpler and also works for Python 3.8 and fully solves #110 on the current version range.

src/test_typing_extensions.py
```diff
@@ -5267,6 +5270,10 @@ class Y(Protocol[T, P]):
                 self.assertEqual(G2.__args__, (int, Concatenate[int, P_2]))
```
Daraan, line 5270: Should the `__args__` here be replaced by `get_args`?

AlexWaygood: fine to keep it as it is, realistically this attribute is never going away :)

src/test_typing_extensions.py
```diff
@@ -5390,25 +5416,56 @@ def test_invalid_uses(self):
...
+            TypeError,
+            'each arg must be a type',
+        ):
+            Concatenate[1, P]
+
+        with self.assertRaisesRegex(
+            TypeError,
+            'each arg must be a type.',
+        ):
+            Concatenate[1, ..., P]
```
AlexWaygood, line 5442: I wonder if it's really worth adding this test? It seems like in later Python versions we decided that it was better to not raise exceptions in these cases, so that seems like the "desirable" behaviour, since we aim to reflect the behaviour in the latest version of the typing module as much as possible. That means that therefore the behaviour of raising exceptions on these in earlier Python versions is sort-of _undesirable_ behaviour. I wouldn't see it as a big priority for us to fix it, but I'm also not sure it's necessary to have these assertions, if they're arguably asserting undesirable behaviour?

Daraan: Thank you for the reasoning and I do agree with you. The reasons why `_type_check` was made more lax was to prevent bugs and have less restrictions to implement new features.

The only reason I see to keep at least the original test would be to test for a regression like forgetting to call `_type_check` at all.

I would roll back the new test, it does not test something that is desirable to not pass. However, the reason to keep the other one around is not that strong. Would you prefer to keep it? In that case I'll add some comment to it.

AlexWaygood:
Looks like you might have closed this accidentally. Probably this is something to do with the fact that it was a PR from your fork's `main` branch. I think it's unfortunately impossible to repair PRs when they get to this state — do you want to recreate the PR using a different branch on your fork? :-)

Daraan:
Thanks for your kind reply. Yep I've just butchered this 🙈, I am sorry. It's as you assume related to the fork from `main` ...

AlexWaygood:
We've all been there, pretty sure I've done it at least once :-)

src/test_typing_extensions.py
Daraan: I've changed the test to assure the `_type_check` but with the *currently* *undesired behavior*.

PR #449, opened 2024-08-21
Support ParamSpec for TypeAliasType

### LINKED ISSUE #448: `TypeAliasType` when a ParamSpec should be replaced raised type error
`TypeAliasType` currently has some bug/limitation when using it together with a `ParamSpec` in Python <3.10.

Root of the problem is the changed `typing._type_check` that was changed in Python 3.11 which makes `typing_extensions.TypeAliasType.__getitem__` run into the TypeError in older Python versions:

```diff
arg: Ellipsis | list[type] # should be valid
-     if not callable(arg):
-        raise TypeError(f"{msg} Got {arg!r:.100}.")
+  if type(arg) is tuple:
+        raise TypeError(f"{msg} Got {arg!r:.100}.")
```

---

The following cases work in Python 3.11+ and fail in <= Python3.10

```py
from typing_extensions import TypeAliasType, Callable, ParamSpec, Any

P = ParamSpec('P')
Foo = TypeAliasType("Foo", Callable[P, Any], type_params=(P,))

# Ok
FooInt = Foo[int] # (int) -> Any

# Error
FooInt2 = Foo[[int]] # (int) -> Any

# Error
FooIntInt = Foo[[int, int]] # (int, int) -> Any

# Error
FooWhatever = Foo[...] # (...) -> Any

# Error; should be invalid context
FooWhatever2 = Foo[[...]] # (...) -> Any

# Error; ( but also invalid parameter expression -> should use Concatenate)
FooIntWhatever = Foo[[int, ...]] # (int, ...) -> Any

# OK, correct parameter expression
FooConcatIntWhatever = Foo[Concatenate[int, ...]]
```

Daraan:
Fixes #448 that usage of ParamSpec with TypeAliasType was limited for < Python 3.11

This PR enables `TypeAliasType` to use Ellipsis and list arguments for `ParamSpec` type params.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7207,7 +7207,7 @@ class TypeAliasTypeTests(BaseTestCase):
         fully_subscripted = still_generic[float]
         self.assertEqual(get_args(fully_subscripted), (Iterable[float],))
         self.assertIs(get_origin(fully_subscripted), ListOrSetT)
-        
+
         # Test ParamSpec and Ellipsis
         P = ParamSpec('P')
         CallableP = TypeAliasType("CallableP", Callable[P, Any], type_params=(P,))
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7285,6 +7285,7 @@ class TypeAliasTypeTests(BaseTestCase):
             # arguments : expected parameters
             int : (),
             ... : (),
+            None : (),
             T2 : (T2,),
             Union[int, List[T2]] : (T2,),
             Tuple[int, str] : (),
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7227,7 +7227,6 @@ class TypeAliasTypeTests(BaseTestCase):
             ... : (),
             T2 : (T2,),
             Union[int, List[T2]] : (T2,),
-            Ts : (Ts,),
             Tuple[int, str] : (),
             Tuple[T, T_default, T2] : (T, T_default, T2),
             Tuple[Unpack[Ts]] : (Ts,),
@@ -7237,11 +7236,16 @@ class TypeAliasTypeTests(BaseTestCase):
             TypeAliasType("NestedAlias", List[T], type_params=(T,))[T2] : (T2,),
         }
         # currently a limitation, these args are no longer unpacked in 3.11
-        test_argument_cases_311_plus = {
-            Unpack[Ts] : (Ts,),
+        # also valid on 310 if GenericAlias is used
+        test_argument_cases_310_plus = {
             Unpack[Tuple[int, T2]] : (T2,),
             Concatenate[int,  P] : (P,),
+            Unpack[Ts] : (Ts,),
         }
+        test_argument_cases_311_plus = {
+            Ts : (Ts,),
+        }
+        test_argument_cases.update(test_argument_cases_310_plus)
         test_argument_cases.update(test_argument_cases_311_plus)
 
         test_alias_cases = [
@@ -7280,8 +7284,10 @@ class TypeAliasTypeTests(BaseTestCase):
                 self.assertEqual(subscripted.__parameters__, ())
             for expected_args, expected_parameters in test_argument_cases.items():
                 with self.subTest(alias=alias, args=expected_args):
+                    if expected_args in test_argument_cases_310_plus and sys.version_info < (3, 10):
+                        self.skipTest("args are unpacked before 3.11 or need GenericAlias")
                     if expected_args in test_argument_cases_311_plus and sys.version_info < (3, 11):
-                        self.skipTest("args are unpacked before 3.11")
+                        self.skipTest("Case is not valid before 3.11")
                     self.assertEqual(get_args(alias[expected_args]), (expected_args,))
                     self.assertEqual(alias[expected_args].__parameters__, expected_parameters)
 
@@ -7397,20 +7403,15 @@ class TypeAliasTypeTests(BaseTestCase):
 
         callable_concat = CallableP[Concatenate[int, P]]
         self.assertEqual(callable_concat.__parameters__, (P,))
-        if TYPING_3_11_0:
-            concat_usage = callable_concat[str]
+        concat_usage = callable_concat[str]
+        with self.subTest("get_args of Concatenate in TypeAliasType"):
+            if sys.version_info < (3, 10, 2):
+                self.skipTest("GenericAlias keeps Concatenate in __args__ prior to 3.10.2")
             self.assertEqual(get_args(concat_usage), ((int, str),))
+        with self.subTest("Equality of parameter_expression without []"):
+            if not TYPING_3_10_0:
+                self.skipTest("Nested list is invalid type form")
             self.assertEqual(concat_usage, callable_concat[[str]])
-        elif TYPING_3_10_0:
-            with self.assertRaises(TypeError, msg="Parameters to generic types must be types"):
-                callable_concat[str]
-            concat_usage = callable_concat[[str]]
-            self.assertEqual(get_args(concat_usage), (int, [str]))
-        else:
-            with self.assertRaises(TypeError, msg="Parameters to generic types must be types"):
-                callable_concat[[str]]
-            concat_usage = callable_concat[str]
-            self.assertEqual(get_args(concat_usage), (int, str))
 
     @skipUnless(TYPING_3_11_0, "__args__ behaves differently")
     def test_substitution_311_plus(self):
@@ -7506,7 +7507,8 @@ class TypeAliasTypeTests(BaseTestCase):
         with self.assertRaises(TypeError, msg="is not a generic class"):
             invalid_tupleT[str]
 
-    @skipIf(TYPING_3_11_0, "Most cases are allowed in 3.11+")
+    # The condition should align with the version of GeneriAlias usage in __getitem__
+    @skipIf(TYPING_3_9_0, "Most cases are allowed in 3.11+ or with GenericAlias")
     def test_invalid_cases_before_3_11(self):
         T = TypeVar('T')
         ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7342,26 +7342,10 @@ class TypeAliasTypeTests(BaseTestCase):
 
         # More complex cases
         Ts = TypeVarTuple("Ts")
-        CallableTs = TypeAliasType("CallableTs", Callable[[Unpack[Ts]], Any], type_params=(Ts, ))
-        unpack_callable = CallableTs[Unpack[Tuple[int, T]]]
-        if TYPING_3_11_0:
-            self.assertEqual(get_args(unpack_callable), (Unpack[Tuple[int, T]],))
-        else:
-            self.assertEqual(get_args(unpack_callable), (Tuple[int, T],))
-        self.assertEqual(unpack_callable.__parameters__, (T,))
-
         Variadic = TypeAliasType("Variadic", Tuple[int, Unpack[Ts]], type_params=(Ts,))
         mixed_subscripedPT = Variadic[Callable[Concatenate[int,  P], T]]
-        self.assertEqual(mixed_subscripedPT.__parameters__, (P, T))
         self.assertEqual(get_args(mixed_subscripedPT), (Callable[Concatenate[int,  P], T],))
 
-        done_subscripted_no_list = mixed_subscripedPT[T, Any] # Expected ParamSpec, ellipsis, or list of types
-        if TYPING_3_10_0:
-            done_subscripted_list = mixed_subscripedPT[[T], Any]
-            self.assertEqual(done_subscripted_list, done_subscripted_no_list)
-        else:
-            with self.assertRaises(TypeError, msg="Parameters to generic types must be types."):
-                mixed_subscripedPT[[T], Any]
 
     @skipUnless(TYPING_3_11_0, "__args__ behaves differently")
     def test_311_substitution(self):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7332,6 +7332,10 @@ class TypeAliasTypeTests(BaseTestCase):
                 subscripted = alias[int, float]
                 self.assertEqual(get_args(subscripted), (int, float))
                 self.assertEqual(subscripted.__parameters__, ())
+            with self.subTest(alias=alias, args=[int, float]):
+                subscripted = alias[[int, float]]
+                self.assertEqual(get_args(subscripted), ([int, float],))
+                self.assertEqual(subscripted.__parameters__, ())
             for expected_args, expected_parameters in test_argument_cases.items():
                 with self.subTest(alias=alias, args=expected_args):
                     self.assertEqual(get_args(alias[expected_args]), (expected_args,))
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7236,14 +7236,14 @@ class TypeAliasTypeTests(BaseTestCase):
             TypeAliasType("NestedAlias", List[T], type_params=(T,))[T2] : (T2,),
         }
         # currently a limitation, these args are no longer unpacked in 3.11
-        # also valid on 310 if GenericAlias is used
+        # OK if GenericAlias in __getitem__ is used
         test_argument_cases_310_plus = {
+            Unpack[Ts] : (Ts,),
             Unpack[Tuple[int, T2]] : (T2,),
             Concatenate[int,  P] : (P,),
-            Unpack[Ts] : (Ts,),
         }
         test_argument_cases_311_plus = {
-            Ts : (Ts,),
+            Ts : (Ts,),  # invalid case
         }
         test_argument_cases.update(test_argument_cases_310_plus)
         test_argument_cases.update(test_argument_cases_311_plus)
@@ -7413,8 +7413,7 @@ class TypeAliasTypeTests(BaseTestCase):
                 self.skipTest("Nested list is invalid type form")
             self.assertEqual(concat_usage, callable_concat[[str]])
 
-    @skipUnless(TYPING_3_11_0, "__args__ behaves differently")
-    def test_substitution_311_plus(self):
+    def test_substitution(self):
         # To pass these tests alias.__args__ in TypeAliasType.__getitem__ needs adjustment
         # Unpack and Concatenate are unpacked in versions before
         T = TypeVar('T')
@@ -7428,6 +7427,7 @@ class TypeAliasTypeTests(BaseTestCase):
         CallableP = TypeAliasType("CallableP", Callable[P, T], type_params=(P, T))
         callable_concat = CallableP[Concatenate[int, P], Any]
         self.assertEqual(get_args(callable_concat), (Concatenate[int, P], Any))
+        self.assertEqual(callable_concat.__parameters__, (P,))
 
     @skipUnless(TYPING_3_12_0, "__args__ behaves differently")
     def test_substitution_312_plus(self):
@@ -7508,7 +7508,7 @@ class TypeAliasTypeTests(BaseTestCase):
             invalid_tupleT[str]
 
     # The condition should align with the version of GeneriAlias usage in __getitem__
-    @skipIf(TYPING_3_9_0, "Most cases are allowed in 3.11+ or with GenericAlias")
+    @skipIf(TYPING_3_10_0, "Most cases are allowed in 3.11+ or with GenericAlias")
     def test_invalid_cases_before_3_11(self):
         T = TypeVar('T')
         ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7318,10 +7318,6 @@ class TypeAliasTypeTests(BaseTestCase):
 
     def test_getitem(self):
         T = TypeVar('T')
-        ValueWithoutT = TypeAliasType("ValueWithoutT", int, type_params=(T,))
-        still_subscripted = ValueWithoutT[str]
-        self.assertEqual(get_args(still_subscripted), (str,))
-
         ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
         subscripted = ListOrSetT[int]
         self.assertEqual(get_args(subscripted), (int,))
@@ -7335,6 +7331,10 @@ class TypeAliasTypeTests(BaseTestCase):
         fully_subscripted = still_generic[float]
         self.assertEqual(get_args(fully_subscripted), (Iterable[float],))
         self.assertIs(get_origin(fully_subscripted), ListOrSetT)
+
+        ValueWithoutTypeVar = TypeAliasType("ValueWithoutTypeParam", int, type_params=(T,))
+        still_subscripted = ValueWithoutTypeVar[str]
+        self.assertEqual(get_args(still_subscripted), (str,))
 
     def test_callable_without_concatenate(self):
         P = ParamSpec('P')
@@ -7396,7 +7396,6 @@ class TypeAliasTypeTests(BaseTestCase):
         CallableP = TypeAliasType("CallableP", Callable[P, T], type_params=(P, T))
         callable_concat = CallableP[Concatenate[int, P], Any]
         self.assertEqual(get_args(callable_concat), (Concatenate[int, P], Any))
-        self.assertEqual(callable_concat.__parameters__, (P,))
 
     @skipUnless(TYPING_3_12_0, "__args__ behaves differently")
     def test_substitution_312_plus(self):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7390,8 +7390,8 @@ class TypeAliasTypeTests(BaseTestCase):
         T = TypeVar("T")
         T2 = TypeVar("T2")
         ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
-        too_many = ListOrSetT[int, bool]
 
+        too_many = ListOrSetT[int, bool]
         self.assertEqual(get_args(too_many), (int, bool))
         self.assertEqual(too_many.__parameters__, ())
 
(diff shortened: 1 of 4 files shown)


Daraan:
Currently pyright is not happy with the synatx and does not infer the types correctly anymore.
Waiting for https://github.com/microsoft/pyright/issues/9027

src/typing_extensions.py
```diff
@@ -3524,16 +3525,41 @@ def _raise_attribute_error(self, name: str) -> Never:
...
-                    item, f'Subscripting {self.__name__} requires a type.'
-                )
-                for item in parameters
-            ]
-            return typing._GenericAlias(self, tuple(parameters))
+        if sys.version_info >= (3, 11):
+            def __getitem__(self, parameters):
+                if not isinstance(parameters, tuple):
+                    parameters = (parameters,)
+                    parameters = [
```
JelleZijlstra, line 3532: Why was this indented? This will work unexpectedly when passing multiple parameters.

src/typing_extensions.py
```diff
@@ -3524,16 +3525,41 @@ def _raise_attribute_error(self, name: str) -> Never:
...
+                    return typing._type_check(
+                        item, f'Subscripting {self.__name__} requires a type.'
+                    )
+
+            def __getitem__(self, parameters):
+                if not isinstance(parameters, tuple):
+                    parameters = (parameters,)
+                    parameters = [
+                        self._check_parameter(item, typ)
+                        for item, typ in zip(parameters, self.__type_params__)
```
JelleZijlstra, line 3560: What if the two sequences are of different length? Do we test this case?

src/typing_extensions.py
```diff
@@ -1,3 +1,4 @@
+# pyright: reportShadowedImports=false
```
Daraan, line 1: remove again

src/typing_extensions.py
Daraan: Thanks for the catch. This is a typo. It should be identical and to the variant before.

JelleZijlstra: This suggests we need more test cases, since this wasn't caught by the tests.

src/typing_extensions.py
Daraan: Just thinking aloud:

You think of a case like this?

```
ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
WrongSubscripted = ListOrSetT[int, bool]
```

JelleZijlstra: Those are useful (they should throw errors I believe). But in the case where there are multiple parameters and they match, we need to make sure `_check_parameter` is called.

src/typing_extensions.py
Daraan: My suggestion would then also be to raise a TypeError like the normal typing 3.12 variant.

src/typing_extensions.py
Daraan: I am not sure about python 3.13. At least in Python 3.12.5 both cases pass without errors using `typing.TypeAliasType`, and for both cases produce no error. Would you suggest making it stricter? Imo the Python 3.12 variant seems to be very lenient.

src/typing_extensions.py
Daraan: It seems hard to actually break this. But found a test that would have caught this.

src/typing_extensions.py
```diff
@@ -3524,16 +3526,47 @@ def _raise_attribute_error(self, name: str) -> Never:
...
+            def __getitem__(self, parameters):
+                if not isinstance(parameters, tuple):
+                    parameters = (parameters,)
+                parameters = [
+                        typing._type_check(
+                            item, f'Subscripting {self.__name__} requires a type.'
+                        )
+                        for item in parameters
+                    ]
+                alias = typing._GenericAlias(self, tuple(parameters))
```
Daraan, line 3539: The typing variant sets the `__value__` and `__type_params__` on the Generic. Also adding this here. Possibly move this to another PR.

src/typing_extensions.py
JelleZijlstra: Ah sorry, I should have checked what I did in CPython. We should match the CPython behavior as precisely as possible.

Daraan: I've now handled this case and added test cases.

src/test_typing_extensions.py
```diff
@@ -3591,7 +3591,7 @@ class CP(P[int]):
             self.assertEqual(x.bar, 'abc')
             self.assertEqual(x.x, 1)
             self.assertEqual(x.__dict__, {'foo': 42, 'bar': 'abc'})
-            s = pickle.dumps(P)
+            s = pickle.dumps(GlobalProto)
             D = pickle.loads(s)
```
Daraan, line 3595: I've renamed this to not have global `P` that is not a ParamSpec. This also caused the wrong inference in the current pyright version mentioned above.

Is it okay to include this in this PR or should it be its own PR?

src/typing_extensions.py
```diff
@@ -3524,16 +3526,63 @@ def _raise_attribute_error(self, name: str) -> Never:
...
+                else:
+                    type_params = self.__type_params__
+                parameters = [
+                        checked
+                        for item, typ in zip(parameters, type_params)
+                        for checked in self._check_parameter(item, typ)
+                ]
+                if sys.version_info[:2] == (3, 10):
+                    alias = typing._GenericAlias(self, tuple(parameters),
+                                                 _typevar_types=(TypeVar, ParamSpec)
```
Daraan, line 3578: TypeVarTuple needs to be probably allowed too. Needs tests and likely update

src/test_typing_extensions.py
AlexWaygood: This PR should only contain the changes that are needed to fix #448. There's a couple of reasons for that:
- It makes the PR easier for us to review :-) The diff on GitHub will only show the specific changes that are required to fix the bug in question
- When looking through git history in the future, it will be much easier to see at-a-glance which specific lines needed to be changed in order to fix this bug, and which lines of code preexisted this change

Having said that, I'm not sure pyright complaining about this assignment would be the best reason for changing the name of this variable, so I'm not sure I'd open another PR either. We don't run pyright in CI, and it would be somewhat silly for us to do so in this repository. We have no hope of getting our code to type-check with any type checker, since we are to some extent defining the typing system itself in the code in this repository. If pyright is complaining about this for you locally and it's bugging you, I think the best solution there might be to change your IDE settings so that pyright doesn't run on this repository ;)

src/test_typing_extensions.py
```diff
@@ -7165,13 +7165,58 @@ def test_attributes(self):
         self.assertEqual(ListOrSetT.__type_params__, (T,))
         self.assertEqual(ListOrSetT.__parameters__, (T,))
 
+        subscripted = ListOrSetT[int]
```
AlexWaygood, line 7168: To some extent this issue predated your PR, but this PR makes it worse, so I think this would be a good opportunity to fix it: this `test_attributes()` method is huge! I think it would be worthwhile to split it into three methods:
- `test_attributes`
- `test_substitution_with_alias_generic_over_typevar`
- `test_substitution_with_alias_generic_over_typevartuple`

This would also clarify that for everything except the first few assertions in this test, we're really testing the substitution behaviour rather than testing which attributes exist on the `TypeAliasType` instances.

src/test_typing_extensions.py
```diff
@@ -7246,6 +7292,171 @@ def test_getitem(self):
...
+        self.assertEqual(get_args(callable_arg2_list), ([int, int],))
+        # (...) -> Any
+        callable_ellipsis = CallableP[...]
+        self.assertEqual(get_args(callable_ellipsis), (...,))
+
+        callable_ellipsis2 = CallableP[(...,)]
+        self.assertEqual(callable_ellipsis, callable_ellipsis2)
+        # (int, ...) -> Any
+        callable_arg_more = CallableP[[int, ...]]
+        self.assertEqual(get_args(callable_arg_more), ([int, ...],))
```
AlexWaygood, line 7322: This is really the same test being done multiple times, but with different inputs to the test. This can be made more readable by using `subTest`, which has the added advantage that one specific test case can fail without failing all the other test cases at the same time: it will report a specific failure for one subtest, but continue to run the test for all the other cases:

```suggestion
        get_args_test_cases = {
            # () -> Any
            CallableP[[]]: ([],),
            # (int) -> Any
            CallableP[int]: (int,),
            CallableP[[int]]: ([int],),
            # (int, int) -> Any
            CallableP[int, int]:  ([int],),
            CallableP[[int, int]]: ([int, int],),
            # (...) -> Any
            CallableP[...]: (...,),
            # (int, ...) -> Any
            CallableP[[int, ...]]: ([int, ...],)
        }
        
        for index, (expression, expected_args) in enumerate(get_args_test_cases.items()) {
            with self.subTest(index=index, expression=expression):
                self.assertEqual(get_args(expression), expected_args))

        self.assertEqual(CallableP[...], CallableP[(...,)])
```

src/test_typing_extensions.py
```diff
@@ -7246,6 +7292,171 @@ def test_getitem(self):
...
+        variadic_tvt_callableA2 = subcriped_callable_tvt[Unpack[Tuple[str]], object]
+        self.assertEqual(variadic_tvt_callableA, variadic_tvt_callableA2)
+
+        variadic_tvt_callableB = subcriped_callable_tvt[[str, int], object]
+        variadic_tvt_callableB2 = subcriped_callable_tvt[Unpack[Tuple[str, int]], object]
+        variadic_tvt_callableB3 = subcriped_callable_tvt[str, int, object]
+        self.assertNotEqual(variadic_tvt_callableB, variadic_tvt_callableB2)
+        self.assertEqual(variadic_tvt_callableB2, variadic_tvt_callableB3)
+
+    def test_invalid_cases(self):
```
AlexWaygood, line 7388: This test is also very large; I think it could be usefully split up into several smaller test methods :-)

src/typing_extensions.py
```diff
@@ -3524,16 +3526,57 @@ def _raise_attribute_error(self, name: str) -> Never:
...
+        if sys.version_info >= (3, 11):
+            def __getitem__(self, parameters):
+                if not isinstance(parameters, tuple):
+                    parameters = (parameters,)
+                parameters = [
+                        typing._type_check(
+                            item, f'Subscripting {self.__name__} requires a type.'
+                        )
+                        for item in parameters
+                    ]
```
AlexWaygood, line 3538: formatting nitpick:

```suggestion
                    typing._type_check(
                        item, f'Subscripting {self.__name__} requires a type.'
                    )
                    for item in parameters
                ]
```

src/typing_extensions.py
```diff
@@ -3524,16 +3526,57 @@ def _raise_attribute_error(self, name: str) -> Never:
...
+                        item, f'Subscripting {self.__name__} requires a type.'
+                    )
+
+            def __getitem__(self, parameters):
+                if not isinstance(parameters, tuple):
+                    parameters = (parameters,)
+                parameters = [
+                        checked
+                        for item in parameters
+                        for checked in self._check_parameter(item)
```
AlexWaygood, line 3568: formatting nitpick:

```suggestion
                    checked
                    for item in parameters
                    for checked in self._check_parameter(item)
```

AlexWaygood (commented):
I haven't looked deeply at the changes in this PR yet (will do so either today or tomorrow), but here's some comments from skimming through:

src/test_typing_extensions.py
```diff
@@ -7165,13 +7165,58 @@ def test_attributes(self):
...
+        self.assertEqual(subcriped_callable_tvt.__parameters__, (Ts, T))
+
+        # Use with Callable+Unpack
+        CallableTs = TypeAliasType("CallableTs", Callable[[Unpack[Ts]], Any], type_params=(Ts, ))
+        self.assertEqual(CallableTs.__type_params__, (Ts,))
+        self.assertEqual(CallableTs.__parameters__, (*Ts,))
+
+        unpack_callable = CallableTs[Unpack[Tuple[int, T]]]
+        self.assertEqual(unpack_callable.__parameters__, (T,))
+
```
Daraan: After restructuring I thought why not make a comprehensive test to test alias-arg combinations.
It is compact and kind of complete, on the other hand it increases the total amount of tests and skipped tests. I am not sure if this is wanted.

src/test_typing_extensions.py
```diff
@@ -7173,6 +7173,118 @@ def test_attributes(self):
...
+                subscripted = alias[int, float]
+                self.assertEqual(get_args(subscripted), (int, float))
+                self.assertEqual(subscripted.__parameters__, ())
+            for expected_args, expected_parameters in test_argument_cases.items():
+                with self.subTest(alias=alias, args=expected_args):
+                    if expected_args in test_argument_cases_311_plus and sys.version_info < (3, 11):
+                        self.skipTest("args are unpacked before 3.11")
+                    self.assertEqual(get_args(alias[expected_args]), (expected_args,))
+                    self.assertEqual(alias[expected_args].__parameters__, expected_parameters)
+
```
Daraan, line 7287: After restructuring I thought why not make a comprehensive test to test alias-arg combinations.
It is compact and kind of complete, on the other hand it increases the total amount of tests and skipped tests. I am not sure if this is wanted.

src/test_typing_extensions.py
Daraan: Follow up question. Some additional artifacts in TypeAliasType.__getitem__, like the dunder attributes or the cases covered by `test_subscription_without_type_params` are not required to fix #448. Should I fork them for another Issue & PR? And if yes, can I base them base them on this PR? I think otherwise it will make things harder (for me).
Possibly I'll try to unify `__getitem__` again and branch based on the version elsewhere.

src/test_typing_extensions.py
Daraan: Thank you great idea. I was not familiar with this yet. Added the subtests. However, if the argument is a list, e.g. CallableP[[int, ...]], the alias is not hashable and a dict cannot be used here. A list of lists looked worse to me.

src/test_typing_extensions.py
```diff
@@ -7246,6 +7286,174 @@ def test_getitem(self):
...
+        self.assertEqual(callable_not_enough.__parameters__, ())
+        self.assertEqual(get_args(callable_not_enough), (int, ))
+
+        callable_too_many = CallableP[str, float, T2, int]
+        self.assertEqual(callable_too_many.__parameters__, (T2, ))
+        self.assertEqual(get_args(callable_too_many), (str, float, T2, int, ))
+
+        # Cases that result in parameterless variable
+
+        # Callable
```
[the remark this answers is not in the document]
Daraan: Tests moved to [`test_list_argument`](https://github.com/Daraan/typing_extensions/blob/f2aa35c38cfded86df6f2d81bfe2315be7f262dd/src/test_typing_extensions.py#L7494)

src/test_typing_extensions.py
```diff
@@ -7233,19 +7345,207 @@ def test_or(self):
...
+        self.assertEqual(too_manyT.__parameters__, (T, ))
+
+        # With and without list for ParamSpec
+        callable_too_manyT = CallablePT[str, float, T]
+        self.assertEqual(get_args(callable_too_manyT), (str, float, T))
+        self.assertEqual(callable_too_manyT.__parameters__, (T, ))
+
+        callable_too_manyT2 = CallablePT[[str], float, int, T2]
+        self.assertEqual(get_args(callable_too_manyT2), ([str], float, int, T2))
+        self.assertEqual(callable_too_manyT2.__parameters__, (T2, ))
```
AlexWaygood, line 7492: I'd use subtests for these as well:

```suggestion
        test_cases = [
            # not_enough
            (TwoT[int],                              [(int,), ()]),
            (TwoT[T],                                [(T,), (T,)]),
            # callable and not enough
            (CallablePT[int],                        [(T,), (T,)]),
            # too many
            (ListOrSetT[int, bool],                  [(int, bool), ()]),
            # callable and too many
            (CallablePT[str, float, int],            [(str, float, int), ()]),
            # Check if TypeVar is still present even if over substituted
            (ListOrSetT[int, T],                     [(int, T), (T,)]),
            # With and without list for ParamSpec
            (CallablePT[str, float, T],              [(str, float, T), (T,)]),
            (CallablePT[[str], float, int, T2], []), [([str], float, int, T2), (T2,)]),
        ]
        
        for index, (alias, [expected_args, expected_params]) in enumerate(test_cases):
            with self.subTest(index=index, alias=alias):
                self.assertEqual(get_args(alias), expected_args)
                self.assertEqual(alias.__parameters__, expected_params)
```

src/test_typing_extensions.py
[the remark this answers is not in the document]
JelleZijlstra: Good catch, I think this is a bug in CPython. Reported python/cpython#124445.

src/typing_extensions.py
JelleZijlstra: > The typing variant sets the __value__ and __type_params__ on the Generic

Not exactly; instead it creates a `types.GenericAlias`, which proxies these attributes. I believe we could also use `types.GenericAlias` here (though not on 3.8 where it doesn't exist).

JelleZijlstra: However, this PR has a better chance of being merged if you limit it to fixing one issue at a time. Feel free to open additional PRs or issues for other improvements.

Daraan: > > The typing variant sets the **value** and **type_params** on the Generic
> 
> Not exactly; instead it creates a `types.GenericAlias`, which proxies these attributes. I believe we could also use `types.GenericAlias` here (though not on 3.8 where it doesn't exist).

Thanks I'll try to strip out the points where its easily possible.

I've now created the appropriate issue for this.
- #469

src/test_typing_extensions.py
```diff
@@ -7233,19 +7345,194 @@ def test_or(self):
...
+                callable_concat[str]
+            concat_usage = callable_concat[[str]]
+            self.assertEqual(get_args(concat_usage), (int, [str]))
+        else:
+            with self.assertRaises(TypeError, msg="Parameters to generic types must be types"):
+                callable_concat[[str]]
+            concat_usage = callable_concat[str]
+            self.assertEqual(get_args(concat_usage), (int, str))
+
+    @skipUnless(TYPING_3_11_0, "__args__ behaves differently")
```
Daraan, line 7415: This is a regression, these cases work on `main`! Needs again change how Unpack is addressed.

src/typing_extensions.py
Daraan: > > The typing variant sets the **value** and **type_params** on the Generic
> 
> Not exactly; instead it creates a `types.GenericAlias`, which proxies these attributes. I believe we could also use `types.GenericAlias` here (though not on 3.8 where it doesn't exist).

I think this works great, I applied this now to this PR as well and it cleans up some tests. Needed some rework in __getitem__ again.

src/test_typing_extensions.py
Daraan: This has been addressed. test is not skipped anymore

AlexWaygood (commented):
This has merge conflicts now; could you take a look?

Daraan:
> This has merge conflicts now; could you take a look?

Thank you. ;y IDE is buggy today and I merged it incorrectly. on it

src/test_typing_extensions.py
```diff
@@ -7173,6 +7173,82 @@ def test_attributes(self):
...
+            TypeAliasType("CallableP", Callable[P, Any], type_params=(P, )),
+            TypeAliasType("CallableT", Callable[..., T], type_params=(T, )),
+            TypeAliasType("CallableTs", Callable[[Unpack[Ts]], Any], type_params=(Ts, )),
+            # TypeVarTuple
+            TypeAliasType("Variadic", Tuple[int, Unpack[Ts]], type_params=(Ts,)),
+            # TypeVar with default
+            TypeAliasType("TupleT_default", Tuple[T_default, T], type_params=(T, T_default)),
+            TypeAliasType("CallableT_default", Callable[[T], T_default], type_params=(T, T_default)),
+            # default order reversed
+            TypeAliasType("TupleT_default_reversed", Tuple[T_default, T], type_params=(T_default, T)),
```
JelleZijlstra, line 7228: Hm, this should be invalid. It's a SyntaxError with the native version:

```
>>> type X[T=int, U] = (T, U)
  File "<python-input-0>", line 1
    type X[T=int, U] = (T, U)
                  ^
SyntaxError: non-default type parameter 'U' follows default type parameter
```

We should also check against this in TypeAliasType in CPython, and then backport to typing-extensions.

src/test_typing_extensions.py
```diff
@@ -7233,12 +7309,13 @@ def test_or(self):
...
 
     def test_getitem(self):
+        T = TypeVar('T')
         ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
         subscripted = ListOrSetT[int]
         self.assertEqual(get_args(subscripted), (int,))
         self.assertIs(get_origin(subscripted), ListOrSetT)
-        with self.assertRaises(TypeError):
-            subscripted[str]
+        with self.assertRaises(TypeError, msg="not a generic class"):
```
JelleZijlstra, line 7317:
```suggestion
        with self.assertRaisesRegex(TypeError, "not a generic class"):
```
`msg=` is the message that gets raised if the test fails, not the expected message.

src/test_typing_extensions.py
```diff
@@ -7247,6 +7324,159 @@ def test_getitem(self):
...
+        callable_concat = CallableP[Concatenate[int, P], Any]
+        self.assertEqual(get_args(callable_concat), (Concatenate[int, P], Any))
+
+    @skipUnless(TYPING_3_12_0, "__args__ behaves differently")
+    def test_substitution_312_plus(self):
+        # To pass these tests alias.__args__ in TypeAliasType.__getitem__ needs adjustment
+        # Would raise: TypeError: Substitution of bare TypeVarTuple is not supported
+        T = TypeVar('T')
+        Ts = TypeVarTuple("Ts")
+        Variadic = TypeAliasType("Variadic", Tuple[int, Unpack[Ts]], type_params=(Ts,))
```
JelleZijlstra, line 7398: This is `typing.TypeAliasType` in 3.12+, so I'm not sure it makes sense to test this in typing_extensions at all.

Daraan: Yes you're right thank you. It makes not much sense. I've kept this as a reminder to see see if I could solve another problem side by side with minimal effort during `__getitem__` but it does not seem to be the case to do this easily. I've moved this to a separate issue as it is not related to this PR and I will remove these tests.

src/test_typing_extensions.py
Daraan: I kind of expected it it be invalid. There is ... a thought... of an issue in my head about default values, but I am not yet sure if it is based on a mistake of mine, if it should be a dicussion first or if it should be reported as a bug (and to which repo).

I think I should move this test to another region (or out of this PR) to keep better track of it. Any suggestions on this?

JelleZijlstra: I would:

- Report a bug to CPython about this. The behavior when using the TypeAliasType constructor should match that of the type statement, so it should be an error to have a type parameter without a default following one with a default.
- Address it in CPython and then backport to typing-extensions
- Remove any tests from this PR that rely on behavior that should be an error because of this bug.

Daraan: Removed the tests and created https://github.com/python/cpython/issues/124787

src/test_typing_extensions.py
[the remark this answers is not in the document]
Daraan: Removed these tests related to this issue.

src/test_typing_extensions.py
Daraan: This currently displays a weired behavior, if I do thie minimal example:
```python
from typing import Tuple, TypeVar, Union, List, Set, Tuple
from typing_extensions import TypeVarTuple, Unpack, get_args, TypeAliasType
T = TypeVar('T')
ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
subscripted = ListOrSetT[int]
subscripted[int]
TypeError: ListOrSetT[int] is not a generic class
```

However, If I run the test code here in 3.10.0/14 this raises "TypeError: There are no type variables left in ListOrSetT[int]".
So far I have not been able to track down where this change comes from.

src/test_typing_extensions.py
```diff
@@ -7246,6 +7286,196 @@ def test_getitem(self):
...
+        self.assertEqual(callable_generic_raw.__parameters__, (T,))
+
+        # Usage with Concatenate
+        callable_concat = CallableP[Concatenate[int, P]]
+        self.assertEqual(callable_concat.__parameters__, (P,))
+        if TYPING_3_11_0:
+            self.assertEqual(get_args(callable_concat), (Concatenate[int, P],))
+            concat_usage = callable_concat[str]
+            self.assertEqual(get_args(concat_usage), ((int, str),))
+            self.assertEqual(concat_usage, callable_concat[[str]])
```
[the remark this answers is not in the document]
Daraan: Since using `types.GenericAlias` these tests became much clearer. Given that `Concatenate` might need some special treatmeant in `TypeAliasType.__getitem__`  (currently done by `isinstance(param, list)`) and this is the only test where this is looked into with a bit more detail, I think it should probaly stay.

src/typing_extensions.py
```diff
@@ -3572,6 +3572,34 @@ def _raise_attribute_error(self, name: str) -> Never:
         def __repr__(self) -> str:
             return self.__name__
 
+        if sys.version_info < (3, 11):
+            def _check_single_param(self, param, recursion=0):
```
JelleZijlstra, line 3576: Doesn't seem like this needs to be a generator, it always yields a single value. Let's make it return that value instead and simplify `_check_parameters`.

Daraan: Looking at it again. As the current cpython implementation does not not seems to use _type_convert or any restrictions, should we just drop all checks here?

src/test_typing_extensions.py
```diff
@@ -7341,6 +7417,114 @@ def test_getitem(self):
...
+        P2 = ParamSpec('P2')
+        CallableP = TypeAliasType("CallableP", Callable[P, Any], type_params=(P,))
+
+        callable_concat = CallableP[Concatenate[int, P2]]
+        self.assertEqual(callable_concat.__parameters__, (P2,))
+        concat_usage = callable_concat[str]
+        with self.subTest("get_args of Concatenate in TypeAliasType"):
+            if not TYPING_3_9_0:
+                # args are: ([<class 'int'>, ~P2],)
+                self.skipTest("Nested ParamSpec is not substituted")
```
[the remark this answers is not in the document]
Daraan: Changes in #489 are necessary to remove `_check_parameters` / `_type_convert`. More specifically, `_ConcatenateGenericAlias` needs `_copy_with` and `__getitem__` if it is not converted to a list via the above functions to be handled correctly in 3.8 & 3.9.

src/typing_extensions.py
Daraan: Changes in #489 are necessary to remove `_check_parameters` / `_type_convert`. More specifically, `_ConcatenateGenericAlias` needs `_copy_with` and `__getitem__` if it is not converted to a list via the above functions to be handled correctly in 3.8 & 3.9.

JelleZijlstra (approved):

Issue #62, opened 2022-08-02
License issue


davidmarek:
Hi, in my company we use the https://clearlydefined.io project to automatically review licensing for any 3rd party libraries. Unfortunately, typing-extensions is labeled as OTHER there. See https://github.com/clearlydefined/curated-data/blob/master/curations/pypi/pypi/-/typing-extensions.yaml

Past tries to correct the license information in clearlydefined failed, because the LICENSE file in this repository doesn't match PSF-2.0 or Python-2.0 license as defined in SPDX. See for example https://github.com/clearlydefined/curated-data/pull/20420

Could you clearly state what license this project uses and update the LICENSE to match the declared license?

srittau:
The `LICENSE` file is basically a direct copy of the `LICENSE` file from the CPython repository, with one exception: The zero-clause BSD license for documentation is absent.

Apart from that, I don't think anyone here can give you a definitive answer, since all information should be present in the LICENSE file. If in doubt, please contact a lawyer.

davidmarek:
When comparing the `LICENSE` with the CPython `LICENSE` it's clear it's a copy paste. Is there a special reason why this package needs to do that? It's confusing because the license talks about "this software (Python)". I checked some other projects (mypy, typeshed, pythondotorg) under python/ org and they usually have their own license.
Thanks for the info, it seems to me that this can't be easily resolved without lawyers :-(

srittau:
Maybe @gvanrossum or @JukkaL have more insight into this.

gvanrossum:
I dunno -- maybe @JelleZijlstra remembers what happened when typing_extensions was separated out from the typing repo?

The linked issues on clearlydefined are highly technical, I don't have time to look into that.

srittau:
I believe we just copied over the LICENSE file from the `typing` repository. And that in turn seems to be copied from CPython.

gvanrossum:
And the cpython license file occasionally gets updated. Perhaps re-copying the license file from cpython will resolve the OP's issue? @davidmarek What do you think?

davidmarek:
I think that would help. CPython is marked as OTHER in the clearlydefined.io dataset but that's because the CPython license doesn't match what's in https://spdx.org/licenses/Python-2.0.html. 
There's an open issue in SPDX to update the license text and it looks like that might be resolved soon.
The issue: https://github.com/spdx/license-list-XML/issues/1200
[...]

PR #451, opened 2024-08-23
Change license identifier in project metadata to text

mschoettle:
I'd like to propose changing the license metadata to text with the license identifier "Python-2.0" reflecting the content of the `LICENSE` file.

License scanning tools such as the one GitLab uses rely on the `license` field returned by the PyPI API which returns `license: null` for this package ([JSON](https://pypi.org/pypi/typing-extensions/json)).

Related to #62

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ version = "4.12.2"
 description = "Backported and Experimental Type Hints for Python 3.8+"
 readme = "README.md"
 requires-python = ">=3.8"
-license = { file = "LICENSE" }
+license = { text = "Python-2.0" }
 keywords = [
     "annotations",
     "backport",


hauntsaninja:
That license doesn't match our existing license

mschoettle:
Unfortunately, the classifiers are ambiguous (or not complete) for some licenses (BSD, Python, Apache). See also here: https://github.com/pypa/trove-classifiers/issues/17. It seems that the classifiers are not taken into account for the `license` field that the PyPI API returns. With the license being specified as proposed by this PR the API returns the license field.

There is also now [PEP 639](https://peps.python.org/pep-0639/) (provisional) which aims to improve this.

srittau:
I don't think we should use classifiers that contradict the official documentation. It's great to see that there's some progress to improve the license situation, but I'd prefer to wait until there's an officially accepted solution. But I'll defer the decision to @JelleZijlstra and @AlexWaygood here, as core developers and most active regarding typing\_extensions.

JelleZijlstra:
I am not very familiar with packaging arcana, but I'd prefer to avoid relying on any unspecified behavior, and instead use PEP 639 now that it's been provisionally accepted.

srittau:
> I am not very familiar with packaging arcana, but I'd prefer to avoid relying on any unspecified behavior, and instead use PEP 639 now that it's been provisionally accepted.

That's the most sensible approach, but I don't think flit supports PEP 639 yet.

mschoettle:
Waiting for PEP 639 makes sense to me.

There is now a feature request for flit: https://github.com/pypa/flit/issues/692

AlexWaygood:
Has this been superseded by #507?

mschoettle:
Looks like it. Although it uses the old syntax, not the new one introduced by PEP 639 I think.

cdce8p:
> Looks like it. Although it uses the old syntax, not the new one introduced by PEP 639 I think.

`flit_core` doesn't yet support the new syntax.

JelleZijlstra:
Looks like flit just merged support for PEP 639. Hopefully that'll be released soon and we can upgrade.

cdce8p:
> Looks like flit just merged support for PEP 639. Hopefully that'll be released soon and we can upgrade.

Flit `3.11` was released earlier today. Opened #530 to update the project metadata.

mschoettle:
Thanks @cdce8p. Closing this as it is superseded by #530.

Issue #469, opened 2024-09-25
GenericAlias from TypeAliasType subscription lacks dunder attributes


Daraan:
Depending on the python version some attributes for subcripted `TypeAliasType` are not present after the `__getitem__` call.

`__name__` is present in Python 3.11+; `"__name__", "__type_params__", "__value__"` are missing in earlier versions.

```python
from typing_extensions import TypeAliasType, List, TypeVar

T = TypeVar("T")
X = TypeAliasType("X", List[T], type_params=(T,)

XT = X[T]

for attr in ("__name__", "__type_params__", "__value__", "__module__"):
    print("XT has", attr, hasattr(XT, attr)
```

### Tests to pass:

The created aliases should pass the following tests:

```python
    def test_alias_attributes(self):
        T = TypeVar('T')
        T2 = TypeVar('T2')
        ListOrSetT = TypeAliasType("ListOrSetT", Union[List[T], Set[T]], type_params=(T,))
        subscripted = ListOrSetT[int]
        still_generic = ListOrSetT[Iterable[T2]]
        fully_subscripted = still_generic[float]
        
        self.assertEqual(subscripted.__name__, "ListOrSetT")
        self.assertEqual(subscripted.__value__, Union[List[T], Set[T]])
        self.assertEqual(subscripted.__type_params__, (T,))
        self.assertEqual(subscripted.__module__, ListOrSetT.__module__)
        self.assertEqual(still_generic.__name__, "ListOrSetT")
        self.assertEqual(still_generic.__value__, Union[List[T], Set[T]])
        self.assertEqual(still_generic.__type_params__, (T,))
        self.assertEqual(still_generic.__module__, ListOrSetT.__module__)
        self.assertEqual(fully_subscripted.__name__, "ListOrSetT")
        self.assertEqual(fully_subscripted.__value__, Union[List[T], Set[T]])
        self.assertEqual(fully_subscripted.__type_params__, (T,))
        self.assertEqual(fully_subscripted.__module__, ListOrSetT.__module__)
```

PR #470, opened 2024-09-25
Add missing dunder attributes for TypeAliasType instances

Daraan:
Adresses: #449
- https://github.com/python/typing_extensions/issues/469

As @JelleZijlstra [mentioned](https://github.com/python/typing_extensions/pull/449#discussion_r1773747048) that GenericAlias can be used to address #469. `GenericAlias` seems promising, however I am not sure what greater underlying changes this swap could maybe cause.

Currently I am still tracing some errors and I want to make sure I am not regressing some cases that currently are not covered by tests. This is also a reason this is currently only used for 3.10+.
See below I think using it in 3.9 will cause more harm than benefits.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -3533,6 +3533,8 @@ else:
                 )
                 for item in parameters
             ]
+            if sys.version_info >= (3, 10):
+                return typing.GenericAlias(self, tuple(parameters))
             alias = typing._GenericAlias(self, tuple(parameters))
             alias.__value__ = self.__value__
             alias.__type_params__ = self.__type_params__
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -3534,7 +3534,7 @@ else:
                 for item in parameters
             ]
             if sys.version_info >= (3, 10):
-                return typing.GenericAlias(self, tuple(parameters))
+                return _types.GenericAlias(self, tuple(parameters))
             alias = typing._GenericAlias(self, tuple(parameters))
             alias.__value__ = self.__value__
             alias.__type_params__ = self.__type_params__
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -3533,7 +3533,14 @@ def __getitem__(self, parameters):
                 )
                 for item in parameters
             ]
-            return typing._GenericAlias(self, tuple(parameters))
+            if sys.version_info >= (3, 10):
+                return typing.GenericAlias(self, tuple(parameters))
```
AlexWaygood, line 3537: It is re-exported in the `typing` module, but that's really an accident. You're meant to import it from the `types` module :)

```suggestion
                return _types.GenericAlias(self, tuple(parameters))
```

Daraan:
From what I see `GenericAlias` can only be safely used here for Python3.10+

In 3.9 it will raise an TypeError in belows test as it tries to subscript the `_ConcatenateGenericAlias` which inherits from list. The location where this is causes is located here https://github.com/python/cpython/blob/8b9a8e0e082f849f6db03b2ed9074a86813b3b77/Lib/typing.py#L779

With `_GenericAlias` I can get the below test to pass in #449 (commit not yet uploaded) however with `GenericAlias` this seems tricky.

```python
    def test_callable_with_concatenate(self):
        P = ParamSpec('P')
        CallableP = TypeAliasType("CallableP", Callable[P, Any], type_params=(P,))

        callable_concat = CallableP[Concatenate[int, P]]
        self.assertEqual(callable_concat.__parameters__, (P,))
        concat_usage = callable_concat[str]  # <-- will fail for 3.9 with GenericAlias
```

AlexWaygood (approved):
Thanks! I pushed a slightly different solution to your PR branch (using a `_GenericAlias` subclass) so that all the tests would pass on py38 and py39 as well. Your tests were fantastic; they really helped to see all aspects of the problem!

JelleZijlstra (approved):

Issue #471, opened 2024-09-26
Backport of "_collect_type_vars" behaves wrong if Unpack is used on generic non-TypeVarTuple types


Daraan:
Consider the following two cases, which only should yield a single type_var `T`.

```python
# Python < 3.11
from typing_extensions import Unpack, TypeVarTuple, _collect_type_vars, Tuple, TypeVar, TypedDict, Generic
Ts = TypeVarTuple("Ts")
T= TypeVar("T")

class Movie(Generic[T], TypedDict):; 
    genre : T

print(_collect_type_vars([Unpack[Movie[T]]]))
# (typing_extensions.Unpack[typing_extensions.Movie[~T]], ~T)

print(_collect_type_vars([Unpack[Tuple[T]]]))
# (typing_extensions.Unpack[typing.Tuple[~T]], ~T)
```

The problem is that the second condition `isinstance(Unpack_instance, TypeVar)` is True, which appends the unpack variable itself which makes no sense.

`typing_extensions` monkey patches this function in the `typing` module

PR #472, opened 2024-09-26
_collect_type_vars should not collect Unpack objects itself

Daraan:
Fixes:  #471

- #471

This PR fixes that these two cases should only yield `T` and not the unpack variable itself:

```py
print(_collect_type_vars([Unpack[Movie[T]]]))
# (typing_extensions.Unpack[typing_extensions.Movie[~T]], ~T)

print(_collect_type_vars([Unpack[Tuple[T]]]))
# (typing_extensions.Unpack[typing.Tuple[~T]], ~T)
```

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5780,7 +5780,8 @@ class UnpackTests(BaseTestCase):
                         klass[int]
 
     def test_with_non_typevartuple(self):
-        # The two functions below are used by typing._GenericAlias and typing.Generic via monkey patching.
+        # These functions are, among others, used to determine typing._GenericAlias.__parameters__
+        # and are monkey patched.
         T = TypeVar('T')
         class MyTypedDict(TypedDict):
             foo: int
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -5779,6 +5779,33 @@ class D(Protocol[T1, T2, Unpack[Ts]]): pass
                     with self.assertRaises(TypeError):
                         klass[int]
 
+    def test_with_non_typevartuple(self):
+        # The two functions below are used by typing._GenericAlias and typing.Generic via monkey patching.
```
JelleZijlstra, line 5783: Could you add a test that exercises public functionality? `_collect_type_vars` is a private function and it doesn't matter what it does; we're interested in something like `MyTypedDictT[int]` misbehaving.

It's fine to keep the tests you have here as more focused regression tests.

src/test_typing_extensions.py
```diff
@@ -5805,6 +5806,11 @@ class MyTypedDictT(Generic[T], TypedDict):
             self.assertEqual(typing_extensions._collect_parameters((Unpack[MyTypedDictT[T]], )), (T,))
             self.assertEqual(typing_extensions._collect_parameters((Unpack[Tuple[int]], )), ())
             self.assertEqual(typing_extensions._collect_parameters((Unpack[Tuple[int, T]], )), (T,))
+        self.assertEqual(typing._GenericAlias(Any, Unpack[MyTypedDict]).__parameters__, ())
```
JelleZijlstra, line 5809: This is still a private object that shouldn't be used directly. Can you focus the tests on use of public functionality that currently doesn't work and is fixed by your change?

Daraan: I think I then misunderstood what you meant by "keep the tests here" 🙃 . Can you clarify that for me?

I interpreted "here" as in the code, but it should have some other case added. Do I then interpret it correctly that you mean "here" with here in this PR and I should remove this whole block for the private function?

JelleZijlstra: Thinking about it more: Let's have this PR add only tests that exercise public functionality, and remove any tests that test only private functions.

JelleZijlstra:
This is looking good now, could you also add a changelog entry like Alex added on a few other PRs today?

AlexWaygood (approved):
Thanks!

JelleZijlstra (approved):

Daraan:
> Thanks!

Thanks to you two as well!

Issue #468, opened 2024-09-25
TypeAliasType not raising TypeError when it is has no parameters


Daraan:
When using a `TypeAliasType` without type_params it should raise an error which it currently doesn't:

```python
from typing import TypeAliasType as typing_TA
from typing_extensions import TypeAliasType

# this should raise an error
Simple = TypeAliasType("Simple", int)
Simple_typing = typing_TA("Simple", int)
try:
   Simple[str]  # No TypeError
except TypeError:
   print("TypeError raised like expected")
else:
   print("No type error was raised.")  # <-- will be printed

Simple_typing[str] # will raise TypeError
```

However, currently in 3.12+ when using `type_params=()` the code unexpectedly allows subscription.
See also  [cpython issue #124498](https://github.com/python/cpython/issues/124498) and its [PR](https://github.com/python/cpython/pull/124499). This issue likely needs an upstream decision first if the following case is intended or a bug:

```python
from typing import TypeAliasType as typing_TA
Simple2 = typing_TA("Simple", int, type_params=())
Simple2[str]  # no type error for typing.TypeAliasType
```

Also related:
- #239
- https://github.com/pydantic/pydantic/issues/6158

Daraan:
Currently the 3.13.0 `typing.TypeAliasType` variant passes the following tests:

```python
    def test_subscription_without_type_params(self):
        Simple = TypeAliasType("Simple", int)
        with self.assertRaises(TypeError, msg="Only generic type aliases are subscriptable"):
            Simple[int]
        with self.assertRaises(TypeError, msg="Only generic type aliases are subscriptable"):
            Simple[[]]
        with self.assertRaises(TypeError, msg="Only generic type aliases are subscriptable"):
            Simple[()]
       
        # A TypeVar in the value does not allow subscription
        T = TypeVar('T')
        MissingTypeParamsErr = TypeAliasType("MissingTypeParamsErr", List[T])
        self.assertEqual(MissingTypeParamsErr.__type_params__, ())
        self.assertEqual(MissingTypeParamsErr.__parameters__, ())
        with self.assertRaises(TypeError, msg="Only generic type aliases are subscriptable"):
            MissingTypeParamsErr[int]
        with self.assertRaises(TypeError, msg="Only generic type aliases are subscriptable"):
            MissingTypeParamsErr[[]]
        with self.assertRaises(TypeError, msg="Only generic type aliases are subscriptable"):
            MissingTypeParamsErr[()]
            
[...]

PR #473, opened 2024-09-26
Raise TypeError when TypeAliasType is subscripted without having type_params

Daraan:
- Fixes: #468

I was not sure if it should check against `__parameters__` or `__type_params__` the [source](https://github.com/sobolevn/cpython/blob/0b647141d587065c5b82bd658485adca8823a943/Objects/typevarobject.c#L1408) checks against the later.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -3524,21 +3524,8 @@ else:
         def __repr__(self) -> str:
             return self.__name__
 
-        def _is_subscriptable(self):
-            if len(self.__parameters__) > 0:
-                return True
-            if _should_collect_from_parameters(self.__value__):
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -3525,6 +3525,8 @@ def __repr__(self) -> str:
             return self.__name__
 
         def __getitem__(self, parameters):
+            if len(self.__type_params__) == 0:
```
AlexWaygood, line 3528: nit:

```suggestion
            if not self.__type_params__:
```

AlexWaygood (approved):
Thanks!

JelleZijlstra (approved):

PR #475, opened 2024-09-28
Rename TypeExpr to TypeForm

JelleZijlstra:
No backwards compatibility required because we never released
TypeExpr.

Also took the opportunity to expand the docstring.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2059,12 +2059,14 @@ elif sys.version_info[:2] >= (3, 9):
 
     @_TypeFormForm
     def TypeForm(self, parameters):
-        """A special typing construct to represent valid type expressions.
+        """A special form representing the value that results from the evaluation
+        of a type expression. This value encodes the information supplied in the
+        type expression, and it represents the type described by that type expression.
 
-        Expressions are assignable to TypeForm if they represent valid type
-        expressions. For example, ``int`` is assignable to ``TypeForm`` (because
-        ``int``, as a type, is a valid type expression), but ``1`` is not (because
-        integers are not valid type expressions).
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
cc @davidfstr @erictraut

src/typing_extensions.py
```diff
@@ -2047,23 +2047,28 @@ def f(val: Union[int, Awaitable[int]]) -> int:
...
+        # that the object is a TypeForm.
         def __call__(self, obj, /):
             return obj
 
-    @_TypeExprForm
-    def TypeExpr(self, parameters):
-        """Special typing form used to represent a type expression.
+    @_TypeFormForm
+    def TypeForm(self, parameters):
+        """A special typing construct to represent valid type expressions.
```
erictraut, line 2062: This phrasing is subtly confusing in the same way that earlier drafts of the PEP were. A `TypeForm` doesn't represent a *type expression*. It represents the set of runtime objects that encode the information in a type expression. This is the reason we moved away from the `TypeExpr` terminology. I recommend using the phrasing from the latest draft of the PEP. I tried hard to make that phrasing as clear as possible to avoid such confusion.

JelleZijlstra: Thanks, I rewrote the docstring to align with the Specification section of PEP 747.

srittau (approved):

erictraut (approved):

davidfstr:
Thanks @JelleZijlstra ! Getting `typing_extensions.TypeExpr` renamed to `TypeForm` was bullet ♯1 on my mypy implementation for this evening but I see you beat me to it!

PR #477, opened 2024-10-01
[3.14] Address invalid inputs of TypeAliasType

Daraan:
I've prepared this draft  which contains two changes:
- `typing_extensions.TypeAliasType` not raising TypeError if a non-tuple is passed compared to the `typing` variant.
- backport of 
https://github.com/python/cpython/pull/124795 for <=3.11 **(needs to be accepted first)**

- [x] Await merge and change if necessary
- [x] Afterwards fix correct python versions. I currently assume the backport is valid in 3.14. 
- [x] Should backport for 3.12-13 be made? Yes.

It should maybe noted that for 3.12.0-3.12.7 the `typing.TypeAliasType` variant does not contain this stricter backport.
I do not think that there is a necessity to do a backport for 3.12 and it will likely do more harm (I did not test it).

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -7312,16 +7312,24 @@ class TypeAliasTypeTests(BaseTestCase):
             class MyAlias(TypeAliasType):
                 pass
 
+    def test_type_params_compatibility(self):
+        # Regression test to assure compatibility with typing variants
+        with self.subTest(type_params="typing.TypeVar"):
+            TypeAliasType("TypingTypeParams", ..., type_params=(typing.TypeVar('T'),))
+        with self.subTest(type_params="typing.TypeAliasType"):
+            if not hasattr(typing, "TypeAliasType"):
+                self.skipTest("typing.TypeAliasType is not available before 3.12")
+            TypeAliasType("TypingTypeParams", ..., type_params=(typing.TypeVarTuple("Ts"),))
+        with self.subTest(type_params="typing.TypeAliasType"):
+            if not hasattr(typing, "ParamSpec"):
+                self.skipTest("typing.ParamSpec is not available before 3.10")
+            TypeAliasType("TypingTypeParams", ..., type_params=(typing.ParamSpec("P"),))
+
     def test_type_params_possibilities(self):
         T = TypeVar('T')
         # Test not a tuple
         with self.assertRaisesRegex(TypeError, "type_params must be a tuple"):
             TypeAliasType("InvalidTypeParams", List[T], type_params=[T])
-
-        # Regression test to assure compatibility with typing.TypeVar
-        typing_T = typing.TypeVar('T')
-        with self.subTest(type_params="typing.TypeVar"):
-            TypeAliasType("TypingTypeParams", List[typing_T], type_params=(typing_T,))
 
         # Test default order and other invalid inputs
         T_default = TypeVar('T_default', default=int)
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
I merged the upstream PR but did not backport it to 3.12 and 3.13, so this should now be ready to go. I'll push some buttons to do so and then review the code.

src/test_typing_extensions.py
```diff
@@ -7312,6 +7312,62 @@ def test_no_instance_subclassing(self):
...
 
+    def test_type_params_compatibility(self):
+        # Regression test to assure compatibility with typing variants
+        with self.subTest(type_params="typing.TypeVar"):
+            TypeAliasType("TypingTypeParams", ..., type_params=(typing.TypeVar('T'),))
+        with self.subTest(type_params="typing.TypeAliasType"):
+            if not hasattr(typing, "TypeAliasType"):
+                self.skipTest("typing.TypeAliasType is not available before 3.12")
+            TypeAliasType("TypingTypeParams", ..., type_params=(typing.TypeVarTuple("Ts"),))
+        with self.subTest(type_params="typing.TypeAliasType"):
```
JelleZijlstra, line 7323: This is the same `type_params` as before. As on some other PRs, seems better to use separate test methods.

src/test_typing_extensions.py
```diff
@@ -7312,6 +7312,62 @@ def test_no_instance_subclassing(self):
             class MyAlias(TypeAliasType):
                 pass
 
+    def test_type_params_compatibility(self):
+        # Regression test to assure compatibility with typing variants
+        with self.subTest(type_params="typing.TypeVar"):
+            TypeAliasType("TypingTypeParams", ..., type_params=(typing.TypeVar('T'),))
+        with self.subTest(type_params="typing.TypeAliasType"):
+            if not hasattr(typing, "TypeAliasType"):
+                self.skipTest("typing.TypeAliasType is not available before 3.12")
```
JelleZijlstra, line 7321: Shouldn't this one look at whether typing.TypeVarTuple exists instead?

src/test_typing_extensions.py
```diff
@@ -7312,6 +7312,62 @@ def test_no_instance_subclassing(self):
...
+            ((str,), f"Expected a type param, got {str!r}"),
+        ]
+
+        for case in valid_cases:
+            with self.subTest(type_params=case):
+                TypeAliasType("OkCase", List[T], type_params=case)
+        for case, msg in invalid_cases:
+            with self.subTest(type_params=case):
+                if TYPING_3_12_0 and sys.version_info < (3, 14):
+                    self.skipTest("No backport for 3.12 and 3.13 requires cpython PR #124795")
```
JelleZijlstra, line 7368: We decided not to backport in CPython. This test should be changed so that typing_extensions does provide the backport.

src/typing_extensions.py
```diff
@@ -3531,11 +3531,24 @@ class TypeAliasType:
...
                 raise TypeError("TypeAliasType name must be a string")
+            if not isinstance(type_params, tuple):
+                raise TypeError("type_params must be a tuple")
             self.__value__ = value
             self.__type_params__ = type_params
 
+            default_value_encountered = False
             parameters = []
             for type_param in type_params:
+                if not isinstance(type_param, (TypeVar, TypeVarTuple, ParamSpec)):
```
JelleZijlstra, line 3542: This needs to accept both the typing and typing_extensions versions of these classes, which are different on some versions. Test this.

src/test_typing_extensions.py
Daraan: As `TypeAliasType` cannot be subclassed, I tried to do a minimal invasive workaround with `__new__` to return the `typing` variant but this fails the pickling check.

Do you know a workaround or is this `__new__` method not something to pursue further?

---

At the moment the <3.12 backport would pass the test and would work as a full backport for 3.12 & 3.13.

JelleZijlstra: I think we can just use the pre-3.12 variant on 3.12 and 3.13.

Daraan: Guess with annotationlib (likely?) comming it needs a backport anyway.

src/test_typing_extensions.py
```diff
@@ -7466,11 +7466,11 @@ def test_type_params_possibilities(self):
             ((T_default, T), f"non-default type parameter {T!r} follows default"),
             ((P_default, P), f"non-default type parameter {P!r} follows default"),
             ((Ts_default, T), f"non-default type parameter {T!r} follows default"),
-
-            # Potentially add invalid inputs, e.g. literals or classes
-            # depends on upstream
+            # Only type params are accepted
             ((1,), "Expected a type param, got 1"),
             ((str,), f"Expected a type param, got {str!r}"),
+            # Unpack backport behaves like TypeVar in some cases
```
JelleZijlstra, line 7472: What "some cases" exactly? Ideally this should be accepted.

JelleZijlstra (approved):

PR #478, opened 2024-10-03
Allow subscription of `Callable[Concatenate[P], Any]` with ellipsis in Python 3.10

### LINKED ISSUE #48: `Concatenate` implementation generates runtime exception if parameterized by `...`

### LINKED ISSUE #110: Backport allowing passing an ellipsis as the last argument of Concatenate to 3.10

Daraan:
Fixes #48 for python 3.10.3+

The first versions 3.10.0 - 3.10.2 did not enforce restrictions on the last parameter of `_ConcatenateGenericAlias`, 3.10.3-3.10.* only supports `ParamSpec`, i.e. no Ellipsis, hence #48 also holds for these versions.

This PR monkey patches the `copy_with` method of `typing._ConcatenateGenericAlias` with the Python 3.11 equivalent that accepts an ellipsis variant.

---

- This does not solve [110](https://github.com/python/typing_extensions/issues/110) , which requires an earlier check during `Concatenate`
- A backport to `3.8, 3.9` would need a monkey patch of `typing._type_check`

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5401,6 +5401,18 @@ class ConcatenateTests(BaseTestCase):
             ):
                 Concatenate[1, P]
 
+    @skipUnless(TYPING_3_10_0, "Missing backported to <=3.9. See issue #48")
+    def test_alias_subscription_with_ellipsis(self):
+        P = ParamSpec('P')
+        X = Callable[Concatenate[int, P], Any]
+
+        C1 = X[...]
+        self.assertEqual(C1.__parameters__, ())
+        with self.subTest("Compare Concatenate[int, ...]"):
+            if sys.version_info[:2] == (3, 10):
+                self.skipTest("Needs Issue #110 | PR # 442: construct Concatenate with ...")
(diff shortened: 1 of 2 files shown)


Daraan:
Welp it fails the "C[...]" cases in `test_typing.py` this tries to address 😥 This goes back to ~ https://github.com/python/cpython/pull/27518

PR #479, opened 2024-10-03
Support subscription of `Callable[Concatenate[P], Any]` with `...` in Python 3.10

Daraan:
Fixes https://github.com/python/typing_extensions/issues/48 for python 3.10.3+

The first versions 3.10.0 - 3.10.2 did not enforce restrictions on the last parameter of _ConcatenateGenericAlias. 3.10.3-3.10.* only support ParamSpec, i.e. no Ellipsis, hence https://github.com/python/typing_extensions/issues/48 also holds for these versions.

This PR adds a backport of `_ConcatenateGenericAlias` which overrrides the problematic `copy_with` method.

---

- This partialy addresses #110 on 3.10, which requires an earlier check during `Concatenate` but does not fix it fully (see the skipped test)

PR #481, opened 2024-10-05
Concatenate to support Ellipsis argument

Daraan:
Fixes https://github.com/python/typing_extensions/issues/110.

This PR adds support for Ellipsis in Concatenate for the Python 3.8-3.10 backports.

Does not fix:
   - https://github.com/python/typing_extensions/issues/48 : Using Ellipsis as argument for Callable with a generic..
      -  <= 3.9: because of typing.Callable limitations
      -  For 3.10 needs modification of `_ConcatenateGenericAlias`

Changes:
    - Changed _concatenate_getitem to reflect the Python3.11 typing module that allows the Ellipsis parameter.
    - For Python 3.10 uses typing._ConcatenateGenericAlias and _concatenate_getitem
    - For <=Python 3.9 the typing_extensions._ConcatenateGenericAlias and _concatenate_getitem like before
    - For <3.9.2 uses a Dummy class that is temporarily swapped with an Ellipsis input
    - added tests and updated tests

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5426,20 +5426,14 @@ class ConcatenateTests(BaseTestCase):
         ):
             Callable[Concatenate[int, ...], Any][Any]
 
-    @skipIf(TYPING_3_11_0, "Args can be non-types in 3.11+")
-    def test_invalid_uses_before_3_11(self):
+    def test_invalid_use(self):
+        # Assure that `_type_check` is called.
         P = ParamSpec('P')
         with self.assertRaisesRegex(
             TypeError,
-            'each arg must be a type',
+            "each arg must be a type",
         ):
-            Concatenate[1, P]
-
-        with self.assertRaisesRegex(
-            TypeError,
-            'each arg must be a type.',
-        ):
-            Concatenate[1, ..., P]
+            Concatenate[(str,), P]
 
     @skipUnless(TYPING_3_11_0 or (3, 10, 0) <= sys.version_info < (3, 10, 2),
                 "Cannot be backported to <=3.9. See issue #48"
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1726,8 +1726,9 @@ class GetUtilitiesTestCase(TestCase):
         self.assertEqual(get_args(Unpack), ())
         self.assertEqual(get_args(Callable[Concatenate[int, P], int]),
                          (Concatenate[int, P], int))
-        if sys.version_info >= (3, 9, 2):
-            # Cannot construct Callable[Concatenate[int, ...] with non-types
+        with self.subTest("Concatenate[int, ...]"):
+            if sys.version_info < (3, 9, 2):
+                self.skipTest("arguments must be types before 3.9.2, i.e. no ...")
             self.assertEqual(get_args(Callable[Concatenate[int, ...], int]),
                         (Concatenate[int, ...], int))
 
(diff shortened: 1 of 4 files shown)


Daraan:
I am sorry I had to recreate #442.

Latest change since then:

```diff
-        if not TYPING_3_11_0:
-            with self.assertRaisesRegex(
-                TypeError,
-                'each arg must be a type',
-           ):
-              Concatenate[1, P]
# different location
+    def test_invalid_use(self):
+       # Assure that `_type_check` is called.
+        P = ParamSpec('P')
+        with self.assertRaisesRegex(
+            TypeError,
+            "each arg must be a type",
+        ):
+            Concatenate[(str,), P]
```

src/test_typing_extensions.py
```diff
@@ -1720,12 +1720,15 @@ class C(Generic[T]): pass
...
-                         (Concatenate[int, P], int))
         self.assertEqual(get_args(Required[int]), (int,))
         self.assertEqual(get_args(NotRequired[int]), (int,))
         self.assertEqual(get_args(Unpack[Ts]), (Ts,))
         self.assertEqual(get_args(Unpack), ())
+        self.assertEqual(get_args(Callable[Concatenate[int, P], int]),
+                         (Concatenate[int, P], int))
+        with self.subTest("Concatenate[int, ...]"):
+            self.assertEqual(get_args(Callable[Concatenate[int, ...], int]),
+                        (Concatenate[int, ...], int))
```
JelleZijlstra, line 1731:
```suggestion
        self.assertEqual(get_args(Callable[Concatenate[int, ...], int]),
                        (Concatenate[int, ...], int))
```
No need for the subtest

src/test_typing_extensions.py
```diff
@@ -5371,13 +5383,27 @@ def test_valid_uses(self):
         self.assertEqual(C1.__origin__, C2.__origin__)
         self.assertNotEqual(C1, C2)
 
-        # Test collections.abc.Callable too.
-        if sys.version_info[:2] >= (3, 9):
-            C3 = collections.abc.Callable[Concatenate[int, P], int]
-            C4 = collections.abc.Callable[Concatenate[int, T, P], T]
+        with self.subTest("typing.Callable with Ellipsis"):
```
JelleZijlstra, line 5386: Perhaps better to make a new test method.

src/test_typing_extensions.py
```diff
@@ -5390,25 +5416,50 @@ def test_invalid_uses(self):
...
+    def test_invalid_use(self):
+        # Assure that `_type_check` is called.
+        P = ParamSpec('P')
+        with self.assertRaisesRegex(
+            TypeError,
+            "each arg must be a type",
+        ):
+            Concatenate[(str,), P]
+
+    @skipUnless(TYPING_3_11_0 or (3, 10, 0) <= sys.version_info < (3, 10, 2),
```
JelleZijlstra, line 5438: In this case I think I'd want to block it on all 3.10 versions. It's a bad experience if something works on early 3.10 and breaks if people upgrade to a later bugfix release. Better to allow it only on 3.11+.

JelleZijlstra: Also, the version check seems wrong here, the change was made in 3.10.3 not 3.10.2.

src/typing_extensions.py
```diff
@@ -1795,28 +1795,56 @@ def __parameters__(self):
             return tuple(
                 tp for tp in self.__args__ if isinstance(tp, (typing.TypeVar, ParamSpec))
             )
+# 3.10+
+else:
+    _ConcatenateGenericAlias = typing._ConcatenateGenericAlias
 
+# 3.8-3.9.2
+class _EllipsisDummy: ...
```
JelleZijlstra, line 1803: If you make this inherit from ParamSpec, can it fix 3.10.3+?

Daraan: I might be missing something here. To what case are you referring here?
This backport of `Concatenate[...]` also works for all 3.10, as the critical part is in `_concatenate_getitem`

Is it possibly related to the following test I jointly introduced also in #479?

```python
def test_alias_subscription_with_ellipsis(self):
    P = ParamSpec('P')
    X = Callable[Concatenate[int, P], Any]
    C1 = X[...]  # <-- Needs PR #479 for 3.10.3+
    self.assertEqual(get_args(C1), (Concatenate[int, ...], Any))  # <-- needs this PR
```

---

That aside; It is possible to do `_EllipsisDummy = ParamSpec("_EllipsisDummy")` if needed as subclassing `ParamSpec` is (currently) not possible. I thought using a new class might have less side-effects than using a ParamSpec instance.

JelleZijlstra: I'm referring to the comment right above here, about the test that works only in 3.10.0-3.10.2 but not 3.10.3+. Making `_EllipsisDummy` an instance of ParamSpec might fix that test for later versions of 3.10.

src/test_typing_extensions.py
JelleZijlstra: Sure!

src/typing_extensions.py
Daraan: I am sorry for the confusion. I do think we talk about the same location :), if not correct me please. I made the necessary edits above.

#479 allows to backport that test to all versions of 3.10, however for #479 the test `self.assertEqual(get_args(C1), (Concatenate[int, ...], Any))` would need this PR to work without an error on 3.10 as well.  
Both PRs are needed make this test work for the 3.10 version range.

JelleZijlstra: I see, thanks. #479 is merged now, so would you mind updating this PR? There's a merge conflict right now.

Daraan: Sure. Unified the code and the test we talked about, aside from that added one more test for invalid usage.

JelleZijlstra (approved):

PR #482, opened 2024-10-09
Fix `Union[..., NoneType]` injection by `get_type_hints` if a `None` default value is used.

### LINKED ISSUE #310: Annotated/get_type_hints interaction in python <3.11
I'm not sure if you're aiming for exact compatibility among python versions for all of the functions you support or not.

But due to https://github.com/python/cpython/issues/90353, `get_type_hints` with `include_extras=True` has inconsistent behavior among python versions below 3.11.

* `def foo(bar: Annotated[Union[int, None]])` -> `Annotated[Union[int, None]]`
* `def foo(bar: Annotated[Union[int, None]] = None)` -> `Optional[Annotated[Union[int, None]]]`

I see https://github.com/python/typing_extensions/pull/94, which mentions not wanting to do more rigorous porting due to it cascading into more code needing to be pulled in, so I'm not sure what the reaction to this will be...

Daraan:
Fixes #310

This PR reverts injection of `Union[..., NoneType]` by `typing.get_type_hints` in Python <3.11 if a function uses a `None` default value.


src/typing_extensions.py
```diff
@@ -1236,10 +1236,75 @@ def get_type_hints(obj, globalns=None, localns=None, include_extras=False):
...
+                else:
+                    ref = ForwardRef(
+                        original_value,
+                        is_argument=not isinstance(obj, _types.ModuleType)
+                    )
+                original_value = typing._eval_type(ref, globalns, localns)
+            # Values was not modified or original is already Optional
+            if original_value == value or _could_be_inserted_optional(original_value):
+                continue
+            # NoneType was added to value
```
Daraan, line 1300: Alternatively `hints[name] = original_value` which should be equivalent. I wonder which would be the safer alternative.

src/test_typing_extensions.py
```diff
@@ -1645,6 +1645,37 @@ def test_final_forward_ref(self):
         self.assertNotEqual(gth(Loop, globals())['attr'], Final[int])
         self.assertNotEqual(gth(Loop, globals())['attr'], Final)
 
+    def test_annotation_and_optional_default(self):
```
Daraan, line 1648: Do you think that more tests are necessary? Do you have ideas where something could go wrong?

src/typing_extensions.py
```diff
@@ -1236,10 +1236,75 @@ def get_type_hints(obj, globalns=None, localns=None, include_extras=False):
...
+        if sys.version_info < (3, 11) and hint:
+            hint = _clean_optional(obj, hint, globalns, localns)
         if include_extras:
             return hint
         return {k: _strip_extras(t) for k, t in hint.items()}
 
+    _NoneType = type(None)
+
+    def _could_be_inserted_optional(t):
+        """detects Union[..., None] pattern"""
```
Daraan, line 1248: `_UnionGenericAlias` does not exist in the whole version range. Is this sufficient to assure that it's a Union? I am not 100% sure.

src/test_typing_extensions.py
```diff
@@ -1645,6 +1645,37 @@ def test_final_forward_ref(self):
...
+        def wanted_optional(bar: optional_annotation): ...
+        def wanted_optional_default(bar: optional_annotation = None): ...
+        def wanted_optional_ref(bar: 'Optional[Annotated[Union[int, None], "data"]]'): ...
+
+        def no_optional(bar: annotation): ...
+        def no_optional_default(bar: annotation = None): ...
+        def no_optional_defaultT(bar: Union[annotation, T] = None): ...
+        def no_optional_defaultT_ref(bar: "Union[annotation, T]" = None): ...
+
+        for func in(wanted_optional, wanted_optional_default, wanted_optional_ref):
```
JelleZijlstra, line 1661:
```suggestion
        for func in (wanted_optional, wanted_optional_default, wanted_optional_ref):
```

src/typing_extensions.py
```diff
@@ -1236,10 +1236,75 @@ def get_type_hints(obj, globalns=None, localns=None, include_extras=False):
...
+            return False
+        return True
+
+    # < 3.11
+    def _clean_optional(obj, hints, globalns=None, localns=None):
+        # reverts injected Union[..., None] cases from typing.get_type_hints
+        # when a None default value is used.
+        # see https://github.com/python/typing_extensions/issues/310
+        original_hints = getattr(obj, '__annotations__', None)
+        defaults = typing._get_defaults(obj)
```
JelleZijlstra, line 1263: Should exit early if there are no defaults. The current code risks accessing `.__annotations__` directly on classes, which could have unwanted effects.

src/typing_extensions.py
```diff
@@ -1236,10 +1236,75 @@ def get_type_hints(obj, globalns=None, localns=None, include_extras=False):
...
+                        original_value,
+                        is_argument=not isinstance(obj, _types.ModuleType)
+                    )
+                original_value = typing._eval_type(ref, globalns, localns)
+            # Values was not modified or original is already Optional
+            if original_value == value or _could_be_inserted_optional(original_value):
+                continue
+            # NoneType was added to value
+            if len(value.__args__) == 2:
+                hints[name] = value.__args__[0]  # not a Union
```
JelleZijlstra, line 1302: It's bad form to both modify your arguments and return them. Either return a new dictionary or modify the dictionary in place and return None.

src/typing_extensions.py
JelleZijlstra: Using `original_value` is incorrect as we may have modified the internals of the hint. For example, `get_type_hints()` turns `List["int"]` into `List[int]`.

JelleZijlstra (changes requested):
This PR is incorrect for this example:

```
>>> def f(x: Union[str, None, "str"] = None): pass
... 
>>> typing_extensions.get_type_hints(f)
{'x': <class 'str'>}
```

I am not sure this approach is viable.

Daraan (commented):
What are your thoughts on the "viability"? For that one example or in general?

I tried to improve the recreation of the `typing.get_type_hints` path that is taken before the injection.

JelleZijlstra:
I haven't thought too hard about examples that might break things, but I'm concerned about using `==` to check whether the two annotations are the same; I don't know if we can rely on equality of annotation objects working reliably for this.

Daraan:
> I haven't thought too hard about examples that might break things, but I'm concerned about using `==` to check whether the two annotations are the same; I don't know if we can rely on equality of annotation objects working reliably for this.

Do you refer to `assertEqual` here, or also for the one in `typing_extensions.py`?

- I'll also added `hash` checks, which are fine. `str` and `repr` are fine, too.
- `assertIs` fails in some(to-many) cases depending on the python version.
In essence the problem can be seen here:

```python
import typing
from typing import List
a = List["str"]

print(typing._eval_type(a, None, None) is typing._eval_type(a, None, None))  # False
```

`typing._eval_type` uses `_GenericAlias.copy_with` or in some cases creates a new `GenericAlias` which do not make use of the caches.

src/typing_extensions.py
```diff
@@ -1236,10 +1236,91 @@ def get_type_hints(obj, globalns=None, localns=None, include_extras=False):
...
+                    t
+                    if get_origin(t) not in (Union, Optional)
+                    else (
+                        Optional[t.__args__[0]]
+                        if get_origin(t) == Optional
+                        else Union[t.__args__]
+                    )
+                )
+                for k, t in hint.items()
+            }
```
Daraan, line 1256: Using this for other Python versions would return cached versions of `Union`. I am not 100% sure if something could go wrong here.

src/test_typing_extensions.py
```diff
@@ -1645,6 +1645,93 @@ def test_final_forward_ref(self):
...
+                # Equality
+                self.assertEqual(type_hints, expected)
+                # Hash
+                for k in type_hints.keys():
+                    self.assertEqual(hash(type_hints[k]), hash(expected[k]))
+                # Repr
+                with self.subTest("Check str and repr"):
+                    if skip_reason == "UnionType not preserved in 3.10":
+                        self.skipTest(skip_reason)
+                    self.assertEqual(str(type_hints) + repr(type_hints), str(expected) + repr(expected))
```
JelleZijlstra, line 1733: Concatenating these seems a bit odd. In any case, `type_hints` is a dictionary, so the str() and repr() are the same.

src/test_typing_extensions.py
```diff
@@ -1645,6 +1645,93 @@ def test_final_forward_ref(self):
...
+            Optional[List[str]]     : Optional[List[str]],
+            Optional[annotation]     : Optional[annotation],
+            Union[str, None, str]   : Optional[str],
+            Unpack[Tuple[int, None]]: Unpack[Tuple[int, None]],
+            # Note: A starred *Ts will use typing.Unpack in 3.11+ see Issue #485
+            Unpack[Ts]              : Unpack[Ts],
+        }
+        # contains a ForwardRef, TypeVar(~prefix) or no expression
+        do_not_stringify_cases = {
+            ()          : {},
```
JelleZijlstra, line 1669:
```suggestion
            ()          : {},  # Special-cased below to create an unannotated parameter
```

src/typing_extensions.py
JelleZijlstra: `get_origin()` is the right choice here.

src/typing_extensions.py
```diff
@@ -1236,10 +1236,83 @@ def get_type_hints(obj, globalns=None, localns=None, include_extras=False):
...
+        original_hints = obj.__annotations__
+        for name, value in hints.items():
+            # Not a Union[..., None] or replacement conditions not fullfilled
+            if (not _could_be_inserted_optional(value)
+                or name not in defaults
+                or defaults[name] is not None
+            ):
+                continue
+            original_value = original_hints[name]
+            if original_value is None:  # should not happen
```
JelleZijlstra, line 1285: Why shouldn't this happen? You can put `None` in annotations.

src/typing_extensions.py
```diff
@@ -1236,10 +1236,83 @@ def get_type_hints(obj, globalns=None, localns=None, include_extras=False):
...
+                        original_value,
+                        is_argument=not isinstance(obj, _types.ModuleType)
+                    )
+            original_evaluated = typing._eval_type(original_value, globalns, localns)
+            if sys.version_info < (3, 9) and get_origin(original_evaluated) is Union:
+                # Union[str, None, "str"] is not reduced to Union[str, None]
+                original_evaluated = Union[original_evaluated.__args__]
+            # Compare if values differ
+            if original_evaluated != value:
+                hints[name] = original_evaluated
```
JelleZijlstra, line 1315:
```suggestion
            hints[name] = original_evaluated
```
Would this also work?

Daraan: Yes it would. However there are minor differences involving identities and caching.

> typing._eval_type uses _GenericAlias.copy_with or in some cases creates a new GenericAlias which do not make use of the caches.

If a ForwardRef is involved `original_evaluated` can be a new object as `_tp_cache` is not queried. Not changing `hints["x"]` will return an object that matches `expected["x"]` by identity in more cases.

That's a minor feature and I am fine with dropping it if you want; for now added a comment to clarify this.

src/test_typing_extensions.py
```diff
@@ -1726,11 +1726,13 @@ def func(x: annot): pass
                 # Hash
                 for k in type_hints.keys():
                     self.assertEqual(hash(type_hints[k]), hash(expected[k]))
+                    # Test if UnionTypes are preserved
+                    self.assertEqual(isinstance(type_hints[k], type(expected[k])), True)
```
JelleZijlstra, line 1730:
```suggestion
                    self.assertIs(type(type_hints[k]), type(expected[k]))
```

src/typing_extensions.py
```diff
@@ -1282,7 +1282,7 @@ def _clean_optional(obj, hints, globalns=None, localns=None):
             ):
                 continue
             original_value = original_hints[name]
-            if original_value is None:  # should not happen
+            if original_value is None:  # should be NoneType already; check just in case
```
JelleZijlstra, line 1285: It shouldn't be NoneType already, since we get this directly from `__annotations__`.

Daraan: Excuse me I have argued wrongly. You are right that `None` would be from `__annotations__`. In such a case `_could_be_inserted_optional(NoneType)` would have already skipped this iteration (`get_origin(NoneType) is not Union`). That's why `original_value` should be never None here.

Daraan:
I am not sure why its still flagged as changes requested from me. I think this should be again waiting for review :)

JelleZijlstra (approved):

PR #489, opened 2024-10-23
Fix Concatenate and Generic with ParamSpec substitution

### LINKED ISSUE #126: Generic with ParamSpec raise TypeError
I create generic class with specifying `__init__` arguments through ParamSpec:
```
from typing import Generic

P = ParamSpec('P')

class CLS(Generic[P]):
    def __init__(self, *args: P.args, **kwargs: P.kwargs):
        ...

class ONE(CLS[int]):
    ...

class TWO(CLS[int, int]):
    ...

# should pass type checks:
ONE(1)
TWO(1, 2)
# should fail type checks:
ONE(1, 2)
TWO(1)
```
mypy type checking works as expected, but during runtime I got next error:
```
  File "temp/issue.py", line 13, in <module>
    class TWO(CLS[int, int]):
  File "/usr/local/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/typing.py", line 277, in inner
    return func(*args, **kwds)
  File "/usr/local/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/typing.py", line 1004, in __class_getitem__
  File "/usr/local/lib/python3.9/site-packages/typing_extensions.py", line 109, in _check_generic
    raise TypeError(f"Too {'many' if alen > elen else 'few'} parameters for {cls};"
TypeError: Too many parameters for <class '__main__.CLS'>; actual 2, expected 1
```
Is it a bug? It works fine in 3.10 and 3.11

### LINKED ISSUE #492: Generic with ParamSpec and Concatenate substitutions behave incorrectly
The following code displays some incorrect behavior, but different parts do not work in certain ranges of 3.8-3.11.3

- 3.8 & 3.9 substitution of `Concatenate`
- 3.10 does not correctly collect type parameters
- < 3.11.3 wrong substitution behavior; type parameter not removed

```python
from typing import Generic, TypeVar
from typing_extensions import ParamSpec, Concatenate
P = ParamSpec('P')
T = TypeVar("T")

class CLS(Generic[P]): ...

Y = CLS[T] # or with a ParamSpec
# 3.10
Y[str]  # raises TypeError is not a generic class

# 3.8 & 3.9
G9 = CLS[Concatenate[T, P]]
H9 = G9[int, [T]] # should be ok; but TypeError
H9[str]  # should also work

# 3.11.0
G8 = CLS[Concatenate[T, ...]]
H1 = G8[int]
H1[str]  # for python 3.11.0-3 this still has a parameter and does not raise a TypeError
```

This is a variant of #126 but with generics which require different fixes. There are multiple flavors of this issue depending on the version parameters are not recognized or not substituted correctly.

For a more comprehensive list check the skipped tests of #491 [here](https://github.com/python/typing_extensions/pull/491/files#diff-5548e20fd249808373a3a5176f402603ac2230e02e681c172dd528fe44e60638)

---

It's possible that some substitutions with `Concatenate` worked prior to #479 and caused a regression here, but I am not yet certain.

---

Fixes come with

- #489

Daraan:
- Fixes #492

this PR is is a follow up on #491 and addresses multiple locations that case problems for different situations and versions; see comments below.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1720,12 +1720,14 @@ class GetUtilitiesTestCase(TestCase):
         # In 3.9 and lower we use typing_extensions's hacky implementation
         # of ParamSpec, which gets incorrectly wrapped in a list
         self.assertIn(get_args(Callable[P, int]), [(P, int), ([P], int)])
-        self.assertEqual(get_args(Callable[Concatenate[int, P], int]),
-                         (Concatenate[int, P], int))
         self.assertEqual(get_args(Required[int]), (int,))
         self.assertEqual(get_args(NotRequired[int]), (int,))
         self.assertEqual(get_args(Unpack[Ts]), (Ts,))
         self.assertEqual(get_args(Unpack), ())
+        self.assertEqual(get_args(Callable[Concatenate[int, P], int]),
+                         (Concatenate[int, P], int))
+        self.assertEqual(get_args(Callable[Concatenate[int, ...], int]),
+                        (Concatenate[int, ...], int))
 
 
 class CollectionsAbcTests(BaseTestCase):
@@ -5267,6 +5269,10 @@ class ParamSpecTests(BaseTestCase):
                 self.assertEqual(G2.__args__, (int, Concatenate[int, P_2]))
                 self.assertEqual(G2.__parameters__, (P_2,))
 
+                G3 = klass[int, Concatenate[int, ...]]
+                self.assertEqual(G3.__args__, (int, Concatenate[int, ...]))
+                self.assertEqual(G3.__parameters__, ())
+
         # The following are some valid uses cases in PEP 612 that don't work:
         # These do not work in 3.9, _type_check blocks the list and ellipsis.
         # G3 = X[int, [int, bool]]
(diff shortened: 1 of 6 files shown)


Daraan:
The code could be slimmed down in some cases if we introduce `__typing_subst__` and `__typing_prepare_subst__`, was there a decision against introducing them?

AlexWaygood:
> The code could be slimmed down in some cases if we introduce `__typing_subst__` and `__typing_prepare_subst__`, was there a decision against introducing them?

Not that I'm aware of. But I think functions in typing.py and elsewhere don't pay attention to these hooks on Python less than 3.11(?), so it might be less effective to add them than you might think. (I'm not confident in this opinion, definitely try it out if you think it might work!)

src/test_typing_extensions.py
```diff
@@ -5273,20 +5278,118 @@ class Y(Protocol[T, P]):
...
+                    G5 = klass[[int, str, T]]
+                    self.assertEqual(G5.__parameters__, (T,))
+                    self.assertEqual(G5.__args__, ((int, str, T),))
+
+                    H9 = G9[int, [T]]
+                    self.assertEqual(H9.__parameters__, (T,))
+
+                # This is an invalid parameter expression but useful for testing correct subsitution
+                G10 = klass[int, Concatenate[str, P]]
+                with self.subTest("Check invalid form substitution"):
```
Daraan, line 5382: This is not a valid parameter expression, it is nice for debugging though, should keep or remove it?

src/typing_extensions.py
```diff
@@ -1795,6 +1812,96 @@ def __parameters__(self):
             return tuple(
                 tp for tp in self.__args__ if isinstance(tp, (typing.TypeVar, ParamSpec))
             )
+
+        # 3.8; needed for typing._subs_tvars
```
JelleZijlstra, line 1816:
```suggestion
        # 3.8; needed for typing._subst_tvars
```

src/typing_extensions.py
```diff
@@ -2496,6 +2609,21 @@ def _is_unpack(obj):
...
+            assert len(self.__args__) == 1
+            arg, = self.__args__
+            if isinstance(arg, typing._GenericAlias):
+                if arg.__origin__ is not tuple:
+                    raise TypeError("Unpack[...] must be used with a tuple type")
+                return arg.__args__
+            return None
+
+        def __getattr__(self, attr):
+            if attr == '__typing_unpacked_tuple_args__':
```
JelleZijlstra, line 2623: Why does this need to go through `__getattr__`? Why can't we add a property named `__typing_unpacked_tuple_args__`?

src/typing_extensions.py
```diff
@@ -3007,6 +3136,12 @@ def wrapper(*args, **kwargs):
                 )
 
 
+def _is_param_expr(arg):
+    return arg is ... or isinstance(
+        arg, (tuple, list, ParamSpec, _ConcatenateGenericAlias)
```
JelleZijlstra, line 3141: Do we need to cover the typing as well as the typing-extensions versions of the last two?

Daraan: For `_ConcatenateGenericAlias` we indeed need it. Fixed and added also tests for both.

src/typing_extensions.py
Daraan: I had a reason once, but I do not see it anymore. Is now a property.

JelleZijlstra (approved):

PR #491, opened 2024-10-23
Fix Generic with ParamSpec not subscriptable with multiple arguments [non-generic-variant]

### LINKED ISSUE #126: Generic with ParamSpec raise TypeError

Daraan:
- Fixes #126

This PR assures that no TypeError is raised, however there are still multiple limitations for Python <3.11.3

I think some of the 3.10 cases could be a regression; I am still looking into that and will create a separate issue. #489 will fix most of the cases as a follow up of this PR.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -3677,6 +3677,33 @@ else:
         def __repr__(self) -> str:
             return self.__name__
 
+        if sys.version_info < (3, 11):
+            def _check_single_param(self, param, recursion=0):
+                # Allow [], [int], [int, str], [int, ...], [int, T]
+                if param is ...:
+                    return ...
+                if param is None:
+                    return None
+                # Note in <= 3.9 _ConcatenateGenericAlias inherits from list
+                if isinstance(param, list) and recursion == 0:
+                    return [self._check_single_param(arg, recursion+1)
+                            for arg in param]
+                return typing._type_check(
+                        param, f'Subscripting {self.__name__} requires a type.'
+                    )
+
+        def _check_parameters(self, parameters):
+            if sys.version_info < (3, 11):
+                return tuple(
+                    self._check_single_param(item)
+                    for item in parameters
+                )
+            return tuple(typing._type_check(
+                        item, f'Subscripting {self.__name__} requires a type.'
+                    )
+                    for item in parameters
+            )
+
         def __getitem__(self, parameters):
             if not self.__type_params__:
                 raise TypeError("Only generic type aliases are subscriptable")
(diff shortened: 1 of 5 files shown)


Daraan:
Note changes here are the combined checks from [`_prepare_paramspec_params(cls, params)`](https://github.com/python/cpython/blob/6a2f12a754d6be016c8abda5f8d432795574b594/Lib/typing.py#L234) and https://github.com/python/cpython/blob/6a2f12a754d6be016c8abda5f8d432795574b594/Lib/typing.py#L1342 that would normally not execute `_check_generic` in 3.10+

src/typing_extensions.py
```diff
@@ -3020,6 +3024,17 @@ def _check_generic(cls, parameters, elen=_marker):
 
         This gives a nice error message in case of count mismatch.
         """
+        # If substituting a single ParamSpec with multiple arguments
+        # we do not check the count
+        if (inspect.isclass(cls) and issubclass(cls, typing.Generic)
+            and len(cls.__parameters__) == 1
+            and isinstance(cls.__parameters__[0], ParamSpec)
```
JelleZijlstra, line 3031: Do we need to check both typing.ParamSpec and typing_extensions.ParamSpec?

JelleZijlstra: I think it might be useful to have a test on 3.10+ that uses `typing.Concatenate` and `typing.ParamSpec`, which are different from the `typing_extensions` version.

Daraan: Tests added & like noted in #489 for Concatenate a patch was necessary.

JelleZijlstra:
This is looking good, but could you fix the merge conflict?

(Sorry for the wait on these PRs!)

Daraan:
#489 was a superset of this one with more changes. As that one made it in first closing this one.

PR #504, opened 2024-11-12
Fix Pydantic third party integration tests

Viicos:
Fixes https://github.com/python/typing_extensions/issues/503


AlexWaygood (approved):
Thank you!

PR #506, opened 2024-11-17
Just use git for checkout

hauntsaninja:
actions/checkout is repeatedly flaky. See also https://github.com/actions/checkout/issues/1951

On the other hand, plain git clone seems to work just fine and never has connection issues checking out the default branch or whatever. If we do still see issues, it'll also be much easier to add retries 

Funnily enough, the YAML is shorter this way too

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -49,9 +49,7 @@ jobs:
     timeout-minutes: 60
     steps:
       - name: Checkout pydantic
-        uses: actions/checkout@v4
-        with:
-          repository: pydantic/pydantic
+        run: git clone https://github.com/pydantic/pydantic.git
       - name: Checkout typing_extensions
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

AlexWaygood (approved):
It's still mystifying to me why it flakes out on this repository so much, and so little on other repositories with similar setups... but if this works, it works!

hauntsaninja:
I'm not sure what other repos I've seen that use actions/checkout on other repos. mypy_primer just uses git clone. This is the problematic branch https://github.com/actions/checkout/blob/main/src/git-source-provider.ts#L138 (when using actions/checkout on your current repo, ref gets set)

AlexWaygood:
> I'm not sure what other repos I've seen that use actions/checkout on other repos.

We do it in flake8-pyi: https://github.com/PyCQA/flake8-pyi/blob/a528bc09991768c65fbbb6dd7936560d06df5682/.github/workflows/typeshed_primer.yml#L37

I don't remember ever seeing this issue over there, though admittedly that workflow runs less regularly than this one!

AlexWaygood:
We also do it in typeshed, and that workflow runs very regularly: https://github.com/python/typeshed/blob/2419d7520b821752c1181d2105a074ef24b9db81/.github/workflows/tests.yml#L181

brianschubert:
Looks like the workflow is still having timeout issues, even with plain git clones: https://github.com/python/typing_extensions/issues/508

hauntsaninja:
Hmm. Well, at least it's easy to add retries now: https://github.com/python/typing_extensions/pull/509

PR #512, opened 2024-11-27
Fix recursive use of Concatenate when mixing modules

Daraan:
Fixes that this code sample throws an error for 3.10:
```python
P = ParamSpec("P")
typing_extensions.Concatenate[int, P][typing.Concatenate[int, P]]
```

> TypeError: The last parameter to Concatenate should be a ParamSpec variable or ellipsis.


JelleZijlstra:
Thanks, could you add a changelog entry? Something like "Fix errors on Python 3.10 when `typing.Concatenate` and `typing_extensions.Concatenate` are used together."

JelleZijlstra (approved):

PR #516, opened 2024-12-06
Change issue text on thirdy_party failure to include exact run URL

brianschubert:
Based on [this SO answer](https://stackoverflow.com/a/70566764/), it's possible generate a URL for the current workflow run. We can use this in the issue body to save a few clicks and make it easier to find the right run when looking at old issues.

I tried this on my fork and it seems to be working great. Here's an example of a workflow run that prints its own URL using this setup (see the error in the dummy-print-url job): https://github.com/brianschubert/typing_extensions/actions/runs/12205086165


JelleZijlstra (approved):
Cool! This would be nice for the typeshed daily test too.

PR #517, opened 2024-12-08
Revert "Disable broken typed-argument-parser tests"

brianschubert:
Reverts python/typing_extensions#515

TAP tests were fixed upstream in https://github.com/swansonk14/typed-argument-parser/pull/157


AlexWaygood (approved):
Thanks!

PR #519, opened 2024-12-11
Start PEP 728 implementation

JelleZijlstra:
Early version but need to get this code off the laptop where it currently lives


src/test_typing_extensions.py
```diff
@@ -4681,23 +4709,39 @@ class AllTheThings(TypedDict):
...
+        self.assertEqual(Child.__optional_keys__, frozenset({'a'}))
+        self.assertEqual(Child.__readonly_keys__, frozenset({}))
+        self.assertEqual(Child.__mutable_keys__, frozenset({'a'}))
+
+    def test_extra_keys_readonly_explicit_closed(self):
+        class Base(TypedDict, closed=True):
+            __extra_items__: ReadOnly[str]
+
+        class Child(Base, closed=True):
             a: NotRequired[str]
```
Daraan, line 4745: `__required_keys__` and `__optional_keys__` will contain `__extra_itmes__` when using 3.14

```suggestion
    @skipIf(TYPING_3_14_0, "Only supported on <3.14")
    def test_extra_keys_readonly_explicit_closed(self):
        class Base(TypedDict, closed=True):
            __extra_items__: ReadOnly[str]

        class Child(Base, closed=True):
            a: NotRequired[str]
```

JelleZijlstra:
OK this is finally ready for review.

Daraan:
A few days ago I made a [PR #1](https://github.com/JelleZijlstra/typing_extensions/pull/1) against your fork branch. Addressing this statement from the PEP:

> If closed is not passed, the value of __closed__ is None.

Daraan:
I realized https://github.com/python/cpython/pull/109544 is missing

```diff
-            if not hasattr(tp_dict, '__total__'):
-                tp_dict.__total__ = total
+           tp_dict.__total__ = total
```

Resulting in a different behavior when using it outside of the spec, i.e. it being the value and not True.

```pyton
class TD_with_total(TypedDict):
   __total__ = "I want a total value here"
```

src/typing_extensions.py
```diff
@@ -984,13 +1042,27 @@ def __new__(cls, name, bases, ns, *, total=True, closed=False):
                 optional_keys.update(base_dict.get('__optional_keys__', ()))
                 readonly_keys.update(base_dict.get('__readonly_keys__', ()))
                 mutable_keys.update(base_dict.get('__mutable_keys__', ()))
-                base_extra_items_type = base_dict.get('__extra_items__', None)
-                if base_extra_items_type is not None:
-                    extra_items_type = base_extra_items_type
+                if getattr(base, "__closed__", None) and not closed:
```
Daraan, line 1045:
```suggestion
                if getattr(base, "__closed__", None) and closed is not None:
```

We want to not pass closed.

src/test_typing_extensions.py
```diff
@@ -4763,34 +4820,34 @@ class Base(TypedDict, closed=True):
         self.assertEqual(Base.__extra_items__, ReadOnly[Union[str, None]])
         self.assertTrue(Base.__closed__)
 
-        class Child(Base):
+        class Child(Base, closed=True):
```
Daraan, line 4823:
```suggestion
        class Child(Base):
```

We should test the case of a child with a closed parent (1/2)

src/test_typing_extensions.py
```diff
@@ -4763,34 +4820,34 @@ class Base(TypedDict, closed=True):
...
         self.assertEqual(Child.__optional_keys__, frozenset({}))
         self.assertEqual(Child.__readonly_keys__, frozenset({}))
-        self.assertEqual(Child.__mutable_keys__, frozenset({'a', "__extra_items__"}))
-        self.assertEqual(Child.__annotations__, {"__extra_items__": int, "a": int})
-        self.assertEqual(Child.__extra_items__, ReadOnly[Union[str, None]])
-        self.assertFalse(Child.__closed__)
+        self.assertEqual(Child.__mutable_keys__, frozenset({'a'}))
+        self.assertEqual(Child.__annotations__, {"a": int})
+        self.assertEqual(Child.__extra_items__, int)
+        self.assertTrue(Child.__closed__)
```
Daraan, line 4833:
```suggestion
        self.assertEqual(Child.__closed__, None)
```

We should test the case of a child with a closed parent (2/2)

Daraan (commented):
From my understanding, this example in the PEP should pass without any errors or warnings:

```
class BaseMovie(TypedDict, closed=True):
    name: str

class MovieA(BaseMovie):  # OK, still closed
    pass
```

Currently this *does* raise a warning; so I think the relevant check should be `closed is False` (see suggestion below).

This will cause quite a few `with self.assertWarns(DeprecationWarning):` to fail afterwards which then need a look-over.

---

From another side should we even check this, or should this only be handled by type-checkers?

> It is a type checker error to pass closed=False if a superclass has closed=True or sets extra_items.

src/typing_extensions.py
JelleZijlstra: Thanks, going to remove this check instead as you hinted at. I wanted to do this in https://github.com/python/peps/pull/4166/files but forgot to update this PR.

src/test_typing_extensions.py
JelleZijlstra: This test tests the legacy `__extra_items__: T` syntax, so I'd rather leave it alone, and instead add a new test working as you suggest.

src/test_typing_extensions.py
```diff
@@ -4763,34 +4819,68 @@ class Base(TypedDict, closed=True):
...
+        self.assertEqual(GrandChild.__extra_items__, Never)
+        self.assertTrue(GrandChild.__closed__)
+
     def test_implicit_extra_items(self):
         class Base(TypedDict):
             a: int
 
-        self.assertEqual(Base.__extra_items__, None)
+        self.assertEqual(Base.__extra_items__, NoExtraItems)
         self.assertFalse(Base.__closed__)
```
Daraan, line 4884:
```suggestion
        self.assertEqual(Base.__extra_items__, NoExtraItems)
        self.assertEqual(Base.__closed__, None)
```

src/test_typing_extensions.py
```diff
@@ -4799,6 +4889,10 @@ class ChildA(Base, closed=True):
         self.assertEqual(ChildA.__extra_items__, Never)
         self.assertTrue(ChildA.__closed__)
```
Daraan, line 4890:
```suggestion
        self.assertEqual(ChildA.__extra_items__, NoExtraItems)
        self.assertTrue(ChildA.__closed__)
```

Daraan (commented):
Looking at the runtime specifications once more. `__extra_items__` currently should also reflect the constructor and `closed=True => extra_items=Never` is only implicitly implied/equivalent. Suggested the necessary changes for that.

Also, suggested the `GrandGrandChild` as an example of an implicit closed subclass. I think such a thing was still missing.

src/test_typing_extensions.py
```diff
@@ -4304,29 +4342,30 @@ class ChildUnclosed(Closed, Unclosed):
             ...
 
         self.assertFalse(ChildUnclosed.__closed__)
```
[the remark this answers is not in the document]
JelleZijlstra: Good catch, I sort of assumed assertFalse was `assert x is False` but it's actually `assert not x`. I'm replacing all tests that use assertFalse/True with `__closed__`.

src/test_typing_extensions.py
JelleZijlstra: These should actually use `assertIs`.

src/test_typing_extensions.py
```diff
@@ -4763,34 +4819,68 @@ class Base(TypedDict, closed=True):
...
+            a: float
+
+        self.assertEqual(GrandChild.__required_keys__, frozenset({'a'}))
+        self.assertEqual(GrandChild.__optional_keys__, frozenset({}))
+        self.assertEqual(GrandChild.__readonly_keys__, frozenset({}))
+        self.assertEqual(GrandChild.__mutable_keys__, frozenset({'a'}))
+        self.assertEqual(GrandChild.__annotations__, {"a": float})
+        self.assertEqual(GrandChild.__extra_items__, Never)
+        self.assertTrue(GrandChild.__closed__)
+
```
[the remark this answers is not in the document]
JelleZijlstra: Added, thanks (but changed some assertEqual to assertIs)

JelleZijlstra:
> I realized https://github.com/python/cpython/pull/109544 is missing

Good catch, I guess when we made that change in CPython we didn't think about the fact that you could set something in the class body. Would you be interested in (1) adding a test case to CPython where there is an assignment to `__total__` in the class body and (2) backporting python/cpython#109544 to typing-extensions?

Daraan:
Is a change to the changelog necessary? Else looks good to me.

---

(1) you have seen already, (2) added #533

Daraan (approved):

src/typing_extensions.py
```diff
@@ -1029,8 +1086,7 @@ def __new__(cls, name, bases, ns, *, total=True, closed=False):
             tp_dict.__optional_keys__ = frozenset(optional_keys)
             tp_dict.__readonly_keys__ = frozenset(readonly_keys)
             tp_dict.__mutable_keys__ = frozenset(mutable_keys)
-            if not hasattr(tp_dict, '__total__'):
-                tp_dict.__total__ = total
+            tp_dict.__total__ = total
```
hauntsaninja, line 1089: don't see a test corresponding to this change

src/test_typing_extensions.py
```diff
@@ -4728,8 +4783,8 @@ class ExtraReadOnly(TypedDict):
         self.assertEqual(ExtraReadOnly.__optional_keys__, frozenset({}))
         self.assertEqual(ExtraReadOnly.__readonly_keys__, frozenset({'__extra_items__'}))
         self.assertEqual(ExtraReadOnly.__mutable_keys__, frozenset({}))
-        self.assertEqual(ExtraReadOnly.__extra_items__, None)
-        self.assertFalse(ExtraReadOnly.__closed__)
+        self.assertIs(ExtraReadOnly.__extra_items__, NoExtraItems)
```
hauntsaninja, line 4786: maybe change this test name to include legacy? kind of reads confusingly otherwise

src/typing_extensions.py
```diff
@@ -877,6 +878,59 @@ def inner(func):
     return inner
 
 
+if not hasattr(typing, "NoDefault") or not hasattr(typing, "NoExtraItems"):
+    class SingletonMeta(type):
```
hauntsaninja, line 882: kind of surprised we bothered. tbh if someone wants to do type(NoDefault).asdf = 123 i could live with that

hauntsaninja (approved):

src/typing_extensions.py
Daraan: This change+tests are covered by #533, a missing backport that was not yet added (see comments above).

src/typing_extensions.py
JelleZijlstra: I guess I did it initially for parity with the C version in CPython for NoDefault, where it's easier to not accept extra attributes, and now that I need another one it makes sense to follow the pattern.

src/test_typing_extensions.py
JelleZijlstra: Good point, done

src/typing_extensions.py
JelleZijlstra: Yes, I just updated the branch and now this change is gone

src/typing_extensions.py
```diff
@@ -877,6 +878,59 @@ def inner(func):
     return inner
 
 
+if not hasattr(typing, "NoDefault") or not hasattr(typing, "NoExtraItems"):
```
AlexWaygood, line 881: I'd be tempted to do something like this as it's a reasonably complex condition:

```diff
diff --git a/src/typing_extensions.py b/src/typing_extensions.py
index ad6664f..833bb7b 100644
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -878,7 +878,9 @@ def _ensure_subclassable(mro_entries):
     return inner
 
 
-if not hasattr(typing, "NoDefault") or not hasattr(typing, "NoExtraItems"):
+_NEEDS_SINGLETONMETA = not hasattr(typing, "NoDefault") or not hasattr(typing, "NoExtraItems")
+
+if _NEEDS_SINGLETONMETA:
     class SingletonMeta(type):
         def __setattr__(cls, attr, value):
             # TypeError is consistent with the behavior of NoneType
@@ -927,7 +929,7 @@ else:
     NoExtraItems = NoExtraItemsType()
     del NoExtraItemsType
 
-if not hasattr(typing, "NoDefault") or not hasattr(typing, "NoExtraItems"):
+if _NEEDS_SINGLETONMETA:
     del SingletonMeta
```

src/test_typing_extensions.py
```diff
@@ -4881,49 +4937,97 @@ class Base(TypedDict, closed=True):
...
+        self.assertIs(GrandGrandChild.__closed__, None)
 
     def test_implicit_extra_items(self):
         class Base(TypedDict):
             a: int
 
-        self.assertEqual(Base.__extra_items__, None)
-        self.assertFalse(Base.__closed__)
+        self.assertIs(Base.__extra_items__, NoExtraItems)
+        self.assertIs(Base.__closed__, None)
```
AlexWaygood, line 5014:
```suggestion
        self.assertIsNone(Base.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4881,49 +4937,97 @@ class Base(TypedDict, closed=True):
...
+        class Child(Base, extra_items=int):
+            a: str
+
+        self.assertEqual(Child.__required_keys__, frozenset({'a'}))
+        self.assertEqual(Child.__optional_keys__, frozenset({}))
+        self.assertEqual(Child.__readonly_keys__, frozenset({}))
+        self.assertEqual(Child.__mutable_keys__, frozenset({'a'}))
+        self.assertEqual(Child.__annotations__, {"a": str})
+        self.assertIs(Child.__extra_items__, int)
+        self.assertIs(Child.__closed__, None)
```
AlexWaygood, line 4986:
```suggestion
        self.assertIsNone(Child.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4881,49 +4937,97 @@ class Base(TypedDict, closed=True):
...
+        class Base(TypedDict, extra_items=ReadOnly[Union[str, None]]):
+            a: int
+
+        self.assertEqual(Base.__required_keys__, frozenset({"a"}))
+        self.assertEqual(Base.__optional_keys__, frozenset({}))
+        self.assertEqual(Base.__readonly_keys__, frozenset({}))
+        self.assertEqual(Base.__mutable_keys__, frozenset({"a"}))
+        self.assertEqual(Base.__annotations__, {"a": int})
+        self.assertEqual(Base.__extra_items__, ReadOnly[Union[str, None]])
+        self.assertIs(Base.__closed__, None)
```
AlexWaygood, line 4975:
```suggestion
        self.assertIsNone(Base.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4140,18 +4143,25 @@ def test_basics_keywords_syntax(self):
...
             TD = TypedDict("TD", cls=type, self=object, typename=str, _typename=int,
-                           fields=list, _fields=dict)
+                           fields=list, _fields=dict,
+                           closed=bool, extra_items=bool)
         self.assertEqual(TD.__name__, 'TD')
         self.assertEqual(TD.__annotations__, {'cls': type, 'self': object, 'typename': str,
-                                              '_typename': int, 'fields': list, '_fields': dict})
+                                              '_typename': int, 'fields': list, '_fields': dict,
+                                              'closed': bool, 'extra_items': bool})
+        self.assertIs(TD.__closed__, None)
```
AlexWaygood, line 4152:
```suggestion
        self.assertIsNone(TD.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4140,18 +4143,25 @@ def test_basics_keywords_syntax(self):
...
-               fields=[('bar', tuple)], _fields={'baz', set})
+               fields=[('bar', tuple)], _fields={'baz', set},
+               closed=None, extra_items="tea pot")
         self.assertEqual(a['cls'], str)
         self.assertEqual(a['self'], 42)
         self.assertEqual(a['typename'], 'foo')
         self.assertEqual(a['_typename'], 53)
         self.assertEqual(a['fields'], [('bar', tuple)])
         self.assertEqual(a['_fields'], {'baz', set})
+        self.assertEqual(a['closed'], None)
```
AlexWaygood, line 4163:
```suggestion
        self.assertIsNone(a['closed'])
```

src/test_typing_extensions.py
```diff
@@ -4448,6 +4440,53 @@ class ChildClosed(Unclosed, Closed):
                     class Wrong(*bases):
                         pass
 
+    def test_closed_values(self):
+        class Implicit(TypedDict): ...
+        class ExplicitTrue(TypedDict, closed=True): ...
+        class ExplicitFalse(TypedDict, closed=False): ...
+
+        self.assertIs(Implicit.__closed__, None)
```
AlexWaygood, line 4448:
```suggestion
        self.assertIsNone(Implicit.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4448,6 +4440,53 @@ class ChildClosed(Unclosed, Closed):
...
+        class Closed(TypedDict, closed=True):
+            __extra_items__: None
+
+        class Unclosed(TypedDict, closed=False):
+            ...
+
+        class ChildUnclosed(Closed, Unclosed):
+            ...
+
+        self.assertIs(ChildUnclosed.__closed__, None)
```
AlexWaygood, line 4464:
```suggestion
        self.assertIsNone(ChildUnclosed.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4448,6 +4440,53 @@ class ChildClosed(Unclosed, Closed):
...
+        self.assertEqual(ChildUnclosed.__extra_items__, NoExtraItems)
+
+        class ChildClosed(Unclosed, Closed):
+            ...
+
+        self.assertIs(ChildClosed.__closed__, None)
```
AlexWaygood, line 4470:
```suggestion
        self.assertIsNone(ChildClosed.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4840,16 +4895,16 @@ def test_extra_key_required(self):
...
             __extra_items__: ReadOnly[str]
 
         self.assertEqual(ExtraReadOnly.__required_keys__, frozenset({'__extra_items__'}))
         self.assertEqual(ExtraReadOnly.__optional_keys__, frozenset({}))
         self.assertEqual(ExtraReadOnly.__readonly_keys__, frozenset({'__extra_items__'}))
         self.assertEqual(ExtraReadOnly.__mutable_keys__, frozenset({}))
-        self.assertEqual(ExtraReadOnly.__extra_items__, None)
-        self.assertFalse(ExtraReadOnly.__closed__)
+        self.assertIs(ExtraReadOnly.__extra_items__, NoExtraItems)
+        self.assertIs(ExtraReadOnly.__closed__, None)
```
AlexWaygood, line 4907:
```suggestion
        self.assertIsNone(ExtraReadOnly.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4858,8 +4913,8 @@ class ExtraRequired(TypedDict):
         self.assertEqual(ExtraRequired.__optional_keys__, frozenset({}))
         self.assertEqual(ExtraRequired.__readonly_keys__, frozenset({}))
         self.assertEqual(ExtraRequired.__mutable_keys__, frozenset({'__extra_items__'}))
-        self.assertEqual(ExtraRequired.__extra_items__, None)
-        self.assertFalse(ExtraRequired.__closed__)
+        self.assertIs(ExtraRequired.__extra_items__, NoExtraItems)
+        self.assertIs(ExtraRequired.__closed__, None)
```
AlexWaygood, line 4917:
```suggestion
        self.assertIsNone(ExtraRequired.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4868,10 +4923,11 @@ class ExtraNotRequired(TypedDict):
         self.assertEqual(ExtraNotRequired.__optional_keys__, frozenset({'__extra_items__'}))
         self.assertEqual(ExtraNotRequired.__readonly_keys__, frozenset({}))
         self.assertEqual(ExtraNotRequired.__mutable_keys__, frozenset({'__extra_items__'}))
-        self.assertEqual(ExtraNotRequired.__extra_items__, None)
-        self.assertFalse(ExtraNotRequired.__closed__)
+        self.assertIs(ExtraNotRequired.__extra_items__, NoExtraItems)
+        self.assertIs(ExtraNotRequired.__closed__, None)
```
AlexWaygood, line 4927:
```suggestion
        self.assertIsNone(ExtraNotRequired.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4881,49 +4937,97 @@ class Base(TypedDict, closed=True):
...
+
+        class GrandGrandChild(GrandChild):
+            ...
+        self.assertEqual(GrandGrandChild.__required_keys__, frozenset({'a'}))
+        self.assertEqual(GrandGrandChild.__optional_keys__, frozenset({}))
+        self.assertEqual(GrandGrandChild.__readonly_keys__, frozenset({}))
+        self.assertEqual(GrandGrandChild.__mutable_keys__, frozenset({'a'}))
+        self.assertEqual(GrandGrandChild.__annotations__, {"a": float})
+        self.assertIs(GrandGrandChild.__extra_items__, NoExtraItems)
+        self.assertIs(GrandGrandChild.__closed__, None)
```
AlexWaygood, line 5007:
```suggestion
        self.assertIsNone(GrandGrandChild.__closed__)
```

src/test_typing_extensions.py
```diff
@@ -4448,6 +4440,53 @@ class ChildClosed(Unclosed, Closed):
...
+            ...
+
+        self.assertIs(ChildClosed.__closed__, None)
+        self.assertEqual(ChildClosed.__extra_items__, NoExtraItems)
+
+    def test_extra_items_class_arg(self):
+        class TD(TypedDict, extra_items=int):
+            a: str
+
+        self.assertEqual(TD.__extra_items__, int)
```
AlexWaygood, line 4477:
```suggestion
        self.assertIs(TD.__extra_items__, int)
```

src/typing_extensions.py
```diff
@@ -1105,9 +1171,14 @@ class Point2D(TypedDict):
...
-            if closed is not False and closed is not True:
+            # Support a field called "closed"
+            if closed is not False and closed is not True and closed is not None:
                 kwargs["closed"] = closed
-                closed = False
+                closed = None
+            # Or "extra_items"
+            if extra_items is not NoExtraItems:
+                kwargs["extra_items"] = extra_items
+                extra_items = NoExtraItems
```
AlexWaygood, line 1181: is this logic tested anywhere? I see a test here that might cover the `closed` logic, but I'm not sure I see one for the `extra_items` logic?

https://github.com/JelleZijlstra/typing_extensions/blob/4c20f9a62395b3d334e051b2854d5e17d82e1252/src/test_typing_extensions.py#L5032-L5041

AlexWaygood (approved):
Basically LGTM, a few nits

src/typing_extensions.py
JelleZijlstra: Good point, added a test case

src/typing_extensions.py
```diff
@@ -935,6 +995,8 @@ def __new__(cls, name, bases, ns, *, total=True, closed=False):
                 if type(base) is not _TypedDictMeta and base is not typing.Generic:
                     raise TypeError('cannot inherit from both a TypedDict type '
                                     'and a non-TypedDict base class')
+            if closed is not None and extra_items is not NoExtraItems:
+                raise TypeError(f"Cannot combine closed={closed!r} and extra_items")
```
AlexWaygood, line 999: I think this branch may also be uncovered, but feel free to merge and tackle this as a followup!

AlexWaygood (approved):

PR #521, opened 2024-12-13
Add 3.12.0 and 3.13.0 tests to CI

Daraan:
I realized that there is no test to check the early 3.12 range.


AlexWaygood (approved):
Thank you!

PR #524, opened 2024-12-21
Fix typos in `TypeIs` docstring

hamdanal:


AlexWaygood (commented):
Thanks! The same typo exists on line 2192, could you fix that one too?

AlexWaygood (approved):

PR #530, opened 2025-02-19
Update flit to use PEP 639

cdce8p:
Flit `3.11` was released today with basic support for PEP 639.
https://flit.pypa.io/en/stable/history.html#version-3-11

Followup to #507, replaces #451.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 # Build system requirements.
 [build-system]
-requires = ["flit_core >=3.4,<4"]
+requires = ["flit_core >=3.11,<4"]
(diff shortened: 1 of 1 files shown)


srittau (approved):

actionless:
this change causing the regression - pip-licenses is detecting the license not as PSF anymore but as unknown:

```
pip-licenses \
		--allow-only "\
Python Software Foundation License;\
"
license UNKNOWN not in allow-only licenses was found for package typing_extensions:4.13.1
```

actionless:
@cdce8p thanks for clarification!

PR #533, opened 2025-02-22
Remove unnecessary hasattr check from TypedDict

Daraan:
Adding @JelleZijlstra's PR https://github.com/python/cpython/pull/109544

Including related additional tests from me https://github.com/python/cpython/pull/130460 - not yet merged, hope that's fine.


JelleZijlstra (approved):
Thanks, would you mind adding a changelog entry since this does change user-visible behavior in an edge case?

JelleZijlstra (approved):

Issue #538, opened 2025-03-12
`_UnpackAlias` `__class__` attribute


Viicos:
Is there a reason to have the `__class__` attribute of the `_UnpackAlias` class set to `typing.TypeVar`?

https://github.com/python/typing_extensions/blob/3f47bf98d0b3b0149e55ba4e4eb8bdb855739a36/src/typing_extensions.py#L2659-L2660

It doesn't seem to be present on `typing`.

Daraan:
This is a needed for <3.11 when e.g. using `class Bar(Generic[typing_extensions.Unpack[Ts]])` to be compatible with `typing`'s `isinstance(typing_extensions.Unpack[...], TypeVar)` check during `Generic.__class_getitem__`.

I think for 3.11 we could safely remove the changed attribute.

Viicos:
Thanks, that clarifies it!

PR #539, opened 2025-03-12
Fix `isinstance(Unpack[Ts], TypeVar)` to be `False` in 3.11

Daraan:
In relation to #538 a minor fix for 3.11 only. For <3.11 it's currently a necessary trick to keep the `__class__ = typing.TypeVar` attribute allow using `Generic[Unpack[Ts]]`.

I am not sure if the additional test is that important, I slightly in favor of removing it again.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2657,7 +2657,7 @@ elif sys.version_info[:2] >= (3, 9):  # 3.9+
             self.__doc__ = _UNPACK_DOC
 
     class _UnpackAlias(typing._GenericAlias, _root=True):
-        if sys.version_info[:2] < (3, 11):
+        if sys.version_info < (3, 11):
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -2657,7 +2657,9 @@ def __init__(self, getitem):
             self.__doc__ = _UNPACK_DOC
 
     class _UnpackAlias(typing._GenericAlias, _root=True):
-        __class__ = typing.TypeVar
+        if sys.version_info[:2] < (3, 11):
```
JelleZijlstra, line 2660:
```suggestion
        if sys.version_info < (3, 11):
```
Minor but more consistent with how we usually write version checks in typed code

JelleZijlstra (approved):

PR #549, opened 2025-03-23
Bump intersphinx base for Python docs to 3.13

brianschubert:
I noticed that the docs are missing intersphinx links to the Python docs for features added to `typing` in 3.13. I was going to add some, but discovered this wasn't possible since intersphinx is currently set up to link to the Python 3.12 docs.

This PR updates intersphinx to link to the Python 3.13 docs, making e.g. `` :py:data:`typing.TypeIs` `` possible.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -27,7 +27,7 @@ extensions = ['sphinx.ext.intersphinx', '_extensions.gh_link']
 templates_path = ['_templates']
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 
-intersphinx_mapping = {'py': ('https://docs.python.org/3.13', None)}
+intersphinx_mapping = {'py': ('https://docs.python.org/3', None)}
 
 add_module_names = False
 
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -27,7 +27,7 @@
 templates_path = ['_templates']
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 
-intersphinx_mapping = {'py': ('https://docs.python.org/3.12', None)}
+intersphinx_mapping = {'py': ('https://docs.python.org/3.13', None)}
```
brianschubert, line 30: Alternatively, this could link to the latest 3.x docs if pinning to a particular version isn't important:
```suggestion
intersphinx_mapping = {'py': ('https://docs.python.org/3', None)}
```

AlexWaygood: I'm not a Sphinx expert, but that _sounds_ like it makes more sense to me, because then we don't have to remember to keep bumping this?

brianschubert: I think it makes sense, unless we're worried about link rot between Python versions? Maybe @JelleZijlstra remembers if it was set up this way for a reason in #166.

srittau (approved):

srittau (approved):

srittau:
Gnnn, now I'm not allowed to merge, since I made the last push.

AlexWaygood (approved):

PR #551, opened 2025-03-23
Fix 'Test and lint' workflow running on forks

### LINKED ISSUE #522: 3rd party tests conditional runs syntax
Inspired by `typing-extensions`' third-party tests, we are doing something similar in Pydantic.

However, the conditional syntax we used (identical to yours) seems to not be working as expected with GHA. We ended up using the `|` syntax instead, fixed here: https://github.com/pydantic/pydantic/pull/11162.

This might be why it was mentioned that the condition did not work as expected in ttps://github.com/python/typing_extensions/pull/226. Happy to fix this.

brianschubert:
Same thing as #525. The 'Third party tests' workflow no longer runs on forks, but the 'Test and lint' workflow also has scheduled tests which weren't fixed.

This should stop the remaining scheduled tests from running on forks. Closes #522.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,16 +24,11 @@ jobs:
   tests:
     name: Run tests
 
+    # if 'schedule' was the trigger,
+    # don't run it on contributors' forks
     if: >-
-      # if 'schedule' was the trigger,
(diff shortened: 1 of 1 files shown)


srittau (approved):
Thanks!

PR #565, opened 2025-04-01
Test Python 3.14 in CI

srittau:


brianschubert:
>but I don't think it's possible. 

I think it might be? I saw something similar in the SQLAlchemy tests: https://github.com/sqlalchemy/sqlalchemy/blob/main/.github/workflows/run-test.yaml#L121

srittau:
Huh, I tried that but VSCode flagged that as invalid and the GitHub docs didn't mention it. I'll change it.

Daraan:
Can/should we add a warning in a failing case?

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message

brianschubert:
> Can/should we add a warning in a failing case?

I think we get this for free - it seems an error annotation still gets emitted for the failing step even when the job doesn't fail: https://github.com/python/typing_extensions/actions/runs/14196009731

.github/workflows/ci.yml
```diff
@@ -69,6 +70,7 @@ jobs:
           cd src
           python --version  # just to make sure we're running the right one
           python -m unittest test_typing_extensions.py
+        continue-on-error: ${{ endsWith(matrix.python-version, '-dev') }}
```
JelleZijlstra, line 73: I think we can remove this now, we're almost in the beta phase and we should aim to support 3.14 from now on.

JelleZijlstra (approved):

JelleZijlstra:
Working on a PR to make the 3.14 tests pass

PR #567, opened 2025-04-01
Move CI to "ubuntu-latest"

JelleZijlstra:
GitHub is decommissioning Ubuntu 20.04. I wouldn't expect our tests to
have a lot of OS version dependencies, so let's try just running
ubuntu-latest everywhere.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,23 +54,23 @@ jobs:
           - "pypy3.9"
           - "pypy3.10"
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: astral-sh/setup-uv@v5
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-          enable-cache: false
+          allow-prereleases: true
 
       - name: Test typing_extensions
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
Oh, but the latest runner doesn't have some of the versions we want... I wonder if we should switch to uv to get Python binaries.

JelleZijlstra:
uv's Python is also problematic:

- 3.8.0 fails mysteriously with `Error: The process '/opt/hostedtoolcache/uv/0.6.11/x86_64/uv' failed with exit code 1` and no details. (I don't blame them, 3.8 is long out of support.)
- Most versions don't come with the CPython test suite pre-installed, which means we can't run CPython tests. Not an unreasonable decision for uv but seems like a deal breaker for us. We could clone the CPython repo ourselves and get the test suite that way, but that's more work.

JelleZijlstra:
Now we're down to:

- 3.8.0, 3.9.0, and 3.10.0 aren't available. Should probably drop 3.8.0 and for 3.9 and 3.10 use the earliest version that GH Actions does have.
- SQLAlchemy tests segfault for some reason

I don't have more time for this right now, but if anyone reading wants to work on a way to get CI back to a reliable state, feel free!

brianschubert:
> if anyone reading wants to work on a way to get CI back to a reliable state, feel free!

Put up #570 :)

PR #570, opened 2025-04-02
Move CI to "ubuntu-latest" (round 2)

brianschubert:
Picks up from [#567](https://github.com/python/typing_extensions/pull/567#issuecomment-2769885755)

* Drops Python 3.8.0
* Bumps 3.9.0 and 3.10.0 to the earliest versions [supported](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json) by the 22.04+ runners 
* Runs everything on ubuntu-latest
    * *except* the SQLAlchemy tests, which are kept at 22.04 (not 20.04, still supported) until the segfault mystery is resolved

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
         # For available versions, see:
         # https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
         python-version:
+          - "3.8"
           - "3.9"
           - "3.9.12"
           - "3.10"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,23 +54,23 @@ jobs:
           - "pypy3.9"
           - "pypy3.10"
 
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -38,12 +38,10 @@ jobs:
         # For available versions, see:
         # https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
         python-version:
-          - "3.8"
```
JelleZijlstra, line 41: Let's keep 3.8 (not 3.8.0) for now, we still want to test 3.8 for the upcoming bugfix release, and it seems to work on 22.04.

brianschubert: Sounds good - do we want to also include any earlier version that 22.04 supports?

JelleZijlstra: Feels not worth it, the testing on point releases is a bit of a spotcheck anyway. Thanks!

JelleZijlstra (approved):

PR #577, opened 2025-04-06
Enable Python 3.13 in cattrs 3rd party tests

provinzkraut:
Was disabled because msgspec didn't support 3.13. msgspec 0.19.0 has since been released with 3.13 support, and cattrs has been updated

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -271,8 +271,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # skip 3.13 because msgspec doesn't support 3.13 yet
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:


AlexWaygood:
Seems like there's now an exception being thrown from `time-machine`: https://github.com/python/typing_extensions/actions/runs/14292230292/job/40054964002?pr=577

provinzkraut:
This seems to have been fixed in time-machine `2.14.0`, but I'm not sure where this dependency gets pulled in.

AlexWaygood:
Errr it probably gets installed as part of this step? https://github.com/python/typing_extensions/blob/8092c3996f4902ad9c74ac2d1d8dd19371ecbaa3/.github/workflows/third_party.yml#L297

Tbh I have no idea if that's the correct incantation for running `cattrs`'s tests -- I've never used pdm and can't remember how I came up with that exact command when I originally wrote this workflow file! I think it was a mix of reading through pdm's docs and trial-and-error.

provinzkraut:
It's an issue with pendulum: https://github.com/python-pendulum/pendulum/issues/782.
Appears to have since been fixed, but the library is waiting for a release with that fix. I'll revisit this PR once a release lands.

AlexWaygood:
Sounds good. I'll mark this PR as draft for now 👍

AlexWaygood:
With a couple of additional hacks, this is now passing CI!

JelleZijlstra (approved):

Issue #560, opened 2025-03-26
typing-extensions 4.13.0 breaks pydantic (<2.10) and SQLAlchemy (latest) project when using custom type


Giuzzilla:
It seems that by upgrading `typing-extensions` from `4.12.2` to `4.13.0`, a project with a simple Pydantic model referencing a custom type breaks.

Reproducible example:

```python
from typing import Literal

from pydantic import BaseModel

type MyType = Literal["test1", "test2"]

class Test(BaseModel):
    a: int
    b: MyType
```

Throws:
```
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for MyType. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
```

I'm on Pydantic `2.9.0` and Python `3.12.9`

I'm not sure if this is an issue on pydantic side and whether it should be fixed there, just pretty sure that it starts to break with 4.13.0

EDIT (see conversation):
Pydantic seems fixed on >=2.10

SQLAlchemy is broken on the latest available version (`2.0.39`) with a very similar case:

```python
from typing import Literal

from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column

type MyType = Literal["test1", "test2"]


class Base(DeclarativeBase):
    pass


class MyModel(Base):
    __tablename__ = "my_model"

    id: Mapped[int] = mapped_column(primary_key=True)
    my_type: Mapped[MyType]
```

Giuzzilla:
Seems to be fine with the combination of Pydantic `2.10.6` + typing-extensions `4.13.0`

srittau:
This is interesting, as typing-extensions had no `Literal`-related changes between 4.12 and 4.13. Furthermore, in Python 3.12 it just re-exports `typing.Literal`, so I'm not sure what changed.

srittau:
But FWIW, I can reproduce the problem using Python 3.12.7 and pydantic 2.10.6 (working)/2.9.2 (not working).

Giuzzilla:
[...]

PR #578, opened 2025-04-06
Add 3rd party tests for litestar

provinzkraut:
Add [Litestar](https://github.com/litestar-org/litestar/) to the 3rd party tests.

The package relies significantly on runtime type inspections (and therefore `typing-extensions` :slightly_smiling_face:), and we've been [made aware](https://github.com/litestar-org/litestar/issues/4088) that it was also affected by #560. 
The issue had first been raised as a [Debian bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1101855) and was forwarded to us; An early(er) warning could have been useful to avoid this downstream issue.

The tests themselves don't take too long (~30-60 seconds), as we only need to run a subset of our test suite.

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -343,6 +343,33 @@ jobs:
             --force-dep "typing-extensions @ file://$(pwd)/../typing-extensions-latest" \
             -- -q --nomemory --notimingintensive
 
+
+  litestar:
+    name: litestar tests
+    needs: skip-schedule-on-fork
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.9", "3.10", "3.11", "3.12", "3.13" ]
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
These tests are nice and fast, the 3rd-party CI only runs once a night, and `litestar` is a very popular library -- so this seems pretty reasonable to me!

Each new library we add to this workflow adds a certain maintenance burden for us, as they do fail reasonably often due to changes beyond our control. I think this one meets the cost-benefit threshhold for it to be worth it, though.

AlexWaygood:
> The issue had first been raised as a [Debian bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1101855) and was forwarded to us; An early(er) warning could have been useful to avoid this downstream issue.

FYI: we also publish release candidates for every `typing_extensions` minor release. We'd love as many downstream projects as possible to test against these, so that we're aware of breakages before they happen!

provinzkraut:
> FYI: we also publish release candidates for every `typing_extensions` minor release. We'd love as many downstream projects as possible to test against these, so that we're aware of breakages before they happen!

Thanks for the pointer, I've added a workflow to test against typing-extension pre-releases on our side as well :)

Issue #572, opened 2025-04-03
support inline `TypedDict`s


DetachHead:
https://peps.python.org/pep-0764/

```py
from typing_extensions import TypedDict

foo: TypedDict[{"name": str}] = {"name": "asdf"}
```

this currently crashes at runtime:
```
TypeError: 'function' object is not subscriptable
```

JelleZijlstra:
cc @Viicos

JelleZijlstra:
Done in #580

PR #579, opened 2025-04-06
Implement support for PEP 764 (inline typed dictionaries)

Viicos:
Fixes https://github.com/python/typing_extensions/issues/572.

As discussed in the [PEP thread](https://discuss.python.org/t/78779), I've tried implementing inlined typed dictionaries as instances of a new `InlinedTypedDict` class, that would look like:

```python
class InlinedTypedDict:
    def __init__(self, fields):
        required_keys = set()
        optional_keys = set()
        readonly_keys = set()
        mutable_keys = set()

        for ann_key, ann_type in fields.items():
            qualifiers = set(_get_typeddict_qualifiers(ann_type))

            if NotRequired not in qualifiers:
                required_keys.add(ann_key)
            else:
                optional_keys.add(ann_key)

            if ReadOnly in qualifiers:
                readonly_keys.add(ann_key)
            else:
                mutable_keys.add(ann_key)

        assert required_keys.isdisjoint(optional_keys), (
            f"Required keys overlap with optional keys:"
            f" {required_keys=}, {optional_keys=}"
        )

        self.__required_keys__ = frozenset(required_keys)
        self.__optional_keys__ = frozenset(optional_keys)
        self.__readonly_keys__ = frozenset(readonly_keys)
        self.__mutable_keys__ = frozenset(mutable_keys)
        self.__total__ = True

    def __call__(self, *args, **kwargs):
        return dict(*args, **kwargs)
```

But I don't think it is worth the effort:
- We still need to set `__mro_entries__` on `InlinedTypedDict` to allow subclassing inlined TDs. To be compatible with the existing logic in `_TypedDictMeta.__new__`, we would most likely need to create an inner "real" TypedDict class matching the `InlinedTypedDict` instance.
- Although the logic `InlinedTypedDict.__init__()` is simpler as it is guaranteed to not have base classes, logic is still duplicated in two places.

Having `"<inlined TypedDict>"` as a `__name__` might be considered as a slightly ugly implementation detail, but considering the above it is worth the trade-off.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5066,6 +5066,38 @@ class TypedDictTests(BaseTestCase):
             class TD(TypedDict, closed=True, extra_items=range):
                 x: str
 
+    def test_inlined_too_many_arguments(self):
+        with self.assertRaises(TypeError):
+            TypedDict[{"a": int}, "extra"]
+
+    def test_inlined_not_a_dict(self):
+        with self.assertRaises(TypeError):
+            TypedDict["not_a_dict"]
+
+    def test_inlined(self):
+        TD = TypedDict[{
+            "a": int,
+            "b": Required[int],
+            "c": NotRequired[int],
+            "d": ReadOnly[int],
+        }]
+        self.assertIsSubclass(TD, dict)
+        self.assertIsSubclass(TD, typing.MutableMapping)
+        self.assertNotIsSubclass(TD, collections.abc.Sequence)
+        self.assertTrue(is_typeddict(TD))
+        self.assertEqual(TD.__name__, "<inlined TypedDict>")
+        self.assertEqual(TD.__module__, __name__)
+        self.assertEqual(TD.__bases__, (dict,))
+        self.assertEqual(TD.__total__, True)
+        self.assertEqual(TD.__required_keys__, {"a", "b", "d"})
+        self.assertEqual(TD.__optional_keys__, {"c"})
+        self.assertEqual(TD.__readonly_keys__, {"d"})
+        self.assertEqual(TD.__mutable_keys__, {"a", "b", "c"})
+
+        inst = TD(a=1, b=2, d=3)
+        self.assertIs(type(inst), dict)
+        self.assertEqual(inst["a"], 1)
+
 
 class AnnotatedTests(BaseTestCase):
 
@@ -6629,7 +6661,7 @@ class AllTests(BaseTestCase):
             exclude |= {
                 'TypeAliasType'
             }
-        if not typing_extensions._PEP_728_IMPLEMENTED:
+        if not typing_extensions._PEP_728_OR_764_IMPLEMENTED:
             exclude |= {'TypedDict', 'is_typeddict'}
         for item in typing_extensions.__all__:
             if item not in exclude and hasattr(typing, item):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5074,6 +5074,10 @@ class TypedDictTests(BaseTestCase):
         with self.assertRaises(TypeError):
             TypedDict["not_a_dict"]
 
+    def test_inlined_empty(self):
+        TD = TypedDict[{}]
+        self.assertEqual(TD.__required_keys__, set())
+
     def test_inlined(self):
         TD = TypedDict[{
             "a": int,
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -910,11 +910,11 @@ if _NEEDS_SINGLETONMETA:
     del SingletonMeta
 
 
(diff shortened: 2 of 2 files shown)


src/typing_extensions.py
```diff
@@ -1162,52 +1157,62 @@ class Point2D(TypedDict):
...
+                    stacklevel=2,
+                )
 
-        ns = {'__annotations__': dict(fields)}
-        module = _caller()
-        if module is not None:
-            # Setting correct module is necessary to make typed dict classes pickleable.
-            ns['__module__'] = module
+            ns = {'__annotations__': dict(fields)}
+            module = _caller(depth=3)
```
Viicos, line 1199: I think it would be great to have `_caller()` match the upstream, in particular having https://github.com/python/cpython/pull/99520. Let me know and I'll send a separate PR.

src/test_typing_extensions.py
```diff
@@ -5053,6 +5053,38 @@ def test_cannot_combine_closed_and_extra_items(self):
...
                 x: str
 
+    def test_inlined_too_many_arguments(self):
+        with self.assertRaises(TypeError):
+            TypedDict[{"a": int}, "extra"]
+
+    def test_inlined_not_a_dict(self):
+        with self.assertRaises(TypeError):
+            TypedDict["not_a_dict"]
+
```
brianschubert, line 5063: Maybe add a test for `TypedDict[{}]`?

brianschubert:
I only just picked up on the fact that the PEP uses the term *inline**d*** TypeDict; somehow I've been reading that as *inline* TypedDict this whole time (possible because that's the term the [mypy extension](https://mypy.readthedocs.io/en/stable/typed_dict.html#inline-typeddict-types) used?).

The term makes sense to me when I think about it, but admittedly when I was toying with this locally and saw `<class '__main__.<inlined TypedDict>'>` in the output, my brain jumped to interpreting it as something like "there was a TypedDict here, but it got _inlined_ away / converted into something else", sort of like an like an "inlined function" compiler optimization. Not that that interpretation would make sense here, but it's makes me wonder if a term like `<anonymous TypedDict>` might be clearer.

JelleZijlstra:
"Inlined" does read wrong to me, I'd also expect that to have something to do with function inlining. "Inline" or "anonymous" seems better.

(Sorry no review of the code yet.)

src/typing_extensions.py
JelleZijlstra: Yes, please do! Seems like it would have to be version-guarded.

src/typing_extensions.py
```diff
@@ -1,3 +1,4 @@
+# pyright: ignore[reportShadowedImports]
```
AlexWaygood, line 1: we don't run pyright in CI, so I'd prefer not to have pyright pragma comments in our code. (I know it breaks pyright's brain a little bit if you open `typing_extensions` or CPython in VSCode; I tend to just use a text editor or IDLE when working on these projects).

src/typing_extensions.py
```diff
@@ -1105,17 +1110,7 @@ def __subclasscheck__(cls, other):
...
-        typename,
-        fields=_marker,
-        /,
-        *,
-        total=True,
-        closed=None,
-        extra_items=NoExtraItems,
-        **kwargs
-    ):
+    class TypedDict(metaclass=_TypedDictMeta):
```
AlexWaygood, line 1113: I suppose it's necessary to make `TypedDict` a class so that you can subscript it? What about making it an instance of a `_SpecialForm` subclass, similar to [how `Annotated` is implemented in CPython](https://github.com/python/cpython/blob/e2476398ee9911b6b0b80e3ca182647805fde81f/Lib/typing.py#L2136-L2195)?

I doubt there's much difference between the two in terms of the user-facing behaviour, but conceptually it sort-of feels "wrong" for `TypedDict` to be a class. You use `TypedDict` to _create_ types; it is not in itself a type; calling `TypedDict` does not create an "instance of `TypedDict`"; `x: TypedDict` is an invalid type annotation; etc.

AlexWaygood (commented):
a couple of high-level comments!

Viicos:
I think this is a good point, I'll switch to _inline_ in the PEP document.

src/typing_extensions.py
Viicos: Yes sorry this is an oversight, I used it in development to avoid having pyright emitting warnings on all the file:
image

src/typing_extensions.py
Viicos: I'll give it a try.

Viicos: I opened https://github.com/python/typing_extensions/pull/580 to compare both.

Viicos:
Closing as https://github.com/python/typing_extensions/pull/579 seems to be the approach we'll stick to.

PR #580, opened 2025-04-07
Implement support for PEP 764 (inline typed dictionaries)

Viicos:
As suggested in https://github.com/python/typing_extensions/pull/579#discussion_r2030276351.

Only downside is that the parent frame to be fetched is different depending on which syntax (functional or inline) is used. This means we need an extra argument (called `__typing_is_inline__` in this PR) to differentiate the two. This isn't required in https://github.com/python/typing_extensions/pull/579.

There are also some issues on 3.8:

- `_ensure_subclassable()` assumes it applies on functions on pypy for Python 3.8:

    https://github.com/python/typing_extensions/blob/281d7b0ca6edad384e641d1066b759c280602919/src/typing_extensions.py#L866-L878

    which isn't the case here (the decorator is applied on the `_TypedDictSpecialForm` instance).

- `typing._SpecialForm` is implemented quite differently on 3.8, so we would need to redefine `__getitem__()` on `_TypedDictSpecialForm`.

~~Maybe support for 3.8 should be dropped first.~~ done in https://github.com/python/typing_extensions/pull/585.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5066,6 +5066,41 @@ class TypedDictTests(BaseTestCase):
             class TD(TypedDict, closed=True, extra_items=range):
                 x: str
 
+    def test_inlined_too_many_arguments(self):
+        with self.assertRaises(TypeError):
+            TypedDict[{"a": int}, "extra"]
+
+    def test_inlined_not_a_dict(self):
+        with self.assertRaises(TypeError):
+            TypedDict["not_a_dict"]
+
+    def test_inlined_empty(self):
+        TD = TypedDict[{}]
+        self.assertEqual(TD.__required_keys__, set())
+
+    def test_inlined(self):
+        TD = TypedDict[{
+            "a": int,
+            "b": Required[int],
+            "c": NotRequired[int],
+            "d": ReadOnly[int],
+        }]
+        self.assertIsSubclass(TD, dict)
+        self.assertIsSubclass(TD, typing.MutableMapping)
+        self.assertNotIsSubclass(TD, collections.abc.Sequence)
+        self.assertTrue(is_typeddict(TD))
+        self.assertEqual(TD.__name__, "<inlined TypedDict>")
+        self.assertEqual(TD.__module__, __name__)
+        self.assertEqual(TD.__bases__, (dict,))
+        self.assertEqual(TD.__total__, True)
+        self.assertEqual(TD.__required_keys__, {"a", "b", "d"})
+        self.assertEqual(TD.__optional_keys__, {"c"})
+        self.assertEqual(TD.__readonly_keys__, {"d"})
+        self.assertEqual(TD.__mutable_keys__, {"a", "b", "c"})
+
+        inst = TD(a=1, b=2, d=3)
+        self.assertIs(type(inst), dict)
+        self.assertEqual(inst["a"], 1)
 
 class AnnotatedTests(BaseTestCase):
 
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5066,19 +5066,29 @@ class TypedDictTests(BaseTestCase):
             class TD(TypedDict, closed=True, extra_items=range):
                 x: str
 
-    def test_inlined_too_many_arguments(self):
+    def test_typed_dict_signature(self):
+        self.assertListEqual(
+            list(inspect.signature(TypedDict).parameters),
+            ['typename', 'fields', 'total', 'closed', 'extra_items', 'kwargs']
+        )
+
+    def test_inline_too_many_arguments(self):
         with self.assertRaises(TypeError):
             TypedDict[{"a": int}, "extra"]
 
-    def test_inlined_not_a_dict(self):
+    def test_inline_not_a_dict(self):
         with self.assertRaises(TypeError):
             TypedDict["not_a_dict"]
 
-    def test_inlined_empty(self):
+    def test_inline_empty(self):
         TD = TypedDict[{}]
         self.assertEqual(TD.__required_keys__, set())
 
-    def test_inlined(self):
+    def test_inline_argument_as_tuple(self):
+        TD = TypedDict[({},)]
+        self.assertEqual(TD.__required_keys__, set())
+
+    def test_inline(self):
         TD = TypedDict[{
             "a": int,
             "b": Required[int],
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5080,9 +5080,19 @@ class TypedDictTests(BaseTestCase):
         with self.assertRaises(TypeError):
             TypedDict["not_a_dict"]
 
+        # a tuple of elements isn't allowed, even if the first element is a dict:
+        with self.assertRaises(TypeError):
+            TypedDict[({"key": int},)]
+
     def test_inline_empty(self):
         TD = TypedDict[{}]
+        self.assertTrue(TD.__total__)
+        self.assertTrue(TD.__closed__)
+        self.assertEqual(TD.__extra_items__, NoExtraItems)
         self.assertEqual(TD.__required_keys__, set())
+        self.assertEqual(TD.__optional_keys__, set())
+        self.assertEqual(TD.__readonly_keys__, set())
+        self.assertEqual(TD.__mutable_keys__,  set())
 
     def test_inline(self):
         TD = TypedDict[{
@@ -5096,9 +5106,15 @@ class TypedDictTests(BaseTestCase):
         self.assertNotIsSubclass(TD, collections.abc.Sequence)
         self.assertTrue(is_typeddict(TD))
         self.assertEqual(TD.__name__, "<inline TypedDict>")
+        self.assertEqual(
+            TD.__annotations__,
+            {"a": int, "b": Required[int], "c": NotRequired[int], "d": ReadOnly[int]},
+        )
         self.assertEqual(TD.__module__, __name__)
         self.assertEqual(TD.__bases__, (dict,))
-        self.assertEqual(TD.__total__, True)
+        self.assertTrue(TD.__total__)
+        self.assertTrue(TD.__closed__)
+        self.assertEqual(TD.__extra_items__, NoExtraItems)
         self.assertEqual(TD.__required_keys__, {"a", "b", "d"})
         self.assertEqual(TD.__optional_keys__, {"c"})
         self.assertEqual(TD.__readonly_keys__, {"d"})
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5086,8 +5086,8 @@ class TypedDictTests(BaseTestCase):
 
     def test_inline_empty(self):
         TD = TypedDict[{}]
-        self.assertTrue(TD.__total__)
-        self.assertTrue(TD.__closed__)
+        self.assertIs(TD.__total__, True)
+        self.assertIs(TD.__closed__, True)
         self.assertEqual(TD.__extra_items__, NoExtraItems)
         self.assertEqual(TD.__required_keys__, set())
         self.assertEqual(TD.__optional_keys__, set())
@@ -5112,8 +5112,8 @@ class TypedDictTests(BaseTestCase):
         )
         self.assertEqual(TD.__module__, __name__)
         self.assertEqual(TD.__bases__, (dict,))
-        self.assertTrue(TD.__total__)
-        self.assertTrue(TD.__closed__)
+        self.assertIs(TD.__total__, True)
+        self.assertIs(TD.__closed__, True)
         self.assertEqual(TD.__extra_items__, NoExtraItems)
         self.assertEqual(TD.__required_keys__, {"a", "b", "d"})
         self.assertEqual(TD.__optional_keys__, {"c"})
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -1078,17 +1078,73 @@ def __subclasscheck__(cls, other):
...
+                # Setting correct module is necessary to make typed dict classes
+                # pickleable.
+                ns['__module__'] = module
+
+            td = _TypedDictMeta(typename, (), ns, total=total, closed=closed,
+                                extra_items=extra_items)
+            td.__orig_bases__ = (TypedDict,)
+            return td
+
     @_ensure_subclassable(lambda bases: (_TypedDict,))
```
AlexWaygood, line 1145: you can just define `__mro_entries__` as a regular method if you do it this way, rather than having to use the `_ensure_subclassable` indirection

```suggestion

        def __mro_entries__(self, bases):
            return (_TypedDict,)

```

AlexWaygood:
I would be tempted to move the complexity out of `_TypedDictSpecialForm.__call__` into a separate function. It means that you can keep the `__typing_is_inline__` parameter out of `_TypedDictSpecialForm.__call__`, which means that it doesn't show up when users do `inspect.signature(TypedDict)`.

Currently on this PR we get the following:

```pycon
>>> from typing_extensions import TypedDict
>>> import inspect
>>> inspect.signature(TypedDict)
<Signature (typename, fields=<sentinel>, /, *, total=True, closed=None, extra_items=typing_extensions.NoExtraItems, __typing_is_inline__=False, **kwargs)>
```

but if you made this change:

<details>
<summary>Suggested patch</summary>

```diff
diff --git a/src/typing_extensions.py b/src/typing_extensions.py
index 9c727a8..dfbaa7d 100644
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -1079,6 +1079,68 @@ else:
     _TypedDict = type.__new__(_TypedDictMeta, 'TypedDict', (), {})
 
 
+    def _create_typeddict(
+        typename,
+        fields,
+        /,
+        *,
+        typing_is_inline,
+        total,
+        closed,
+        extra_items,
+        **kwargs
+    ):
+        if fields is _marker or fields is None:
+            if fields is _marker:
+                deprecated_thing = (
+                    "Failing to pass a value for the 'fields' parameter"
+                )
+            else:
+                deprecated_thing = "Passing `None` as the 'fields' parameter"
+
+            example = f"`{typename} = TypedDict({typename!r}, {{}})`"
+            deprecation_msg = (
+                f"{deprecated_thing} is deprecated and will be disallowed in "
+                "Python 3.15. To create a TypedDict class with 0 fields "
+                "using the functional syntax, pass an empty dictionary, e.g. "
+            ) + example + "."
+            warnings.warn(deprecation_msg, DeprecationWarning, stacklevel=2)
+            # Support a field called "closed"
+            if closed is not False and closed is not True and closed is not None:
+                kwargs["closed"] = closed
+                closed = None
+            # Or "extra_items"
+            if extra_items is not NoExtraItems:
+                kwargs["extra_items"] = extra_items
+                extra_items = NoExtraItems
+            fields = kwargs
+        elif kwargs:
+            raise TypeError("TypedDict takes either a dict or keyword arguments,"
+                            " but not both")
+        if kwargs:
+            if sys.version_info >= (3, 13):
+                raise TypeError("TypedDict takes no keyword arguments")
+            warnings.warn(
+                "The kwargs-based syntax for TypedDict definitions is deprecated "
+                "in Python 3.11, will be removed in Python 3.13, and may not be "
+                "understood by third-party type checkers.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+        ns = {'__annotations__': dict(fields)}
+        module = _caller(depth=5 if typing_is_inline else 3)
+        if module is not None:
+            # Setting correct module is necessary to make typed dict classes
+            # pickleable.
+            ns['__module__'] = module
+
+        td = _TypedDictMeta(typename, (), ns, total=total, closed=closed,
+                            extra_items=extra_items)
+        td.__orig_bases__ = (TypedDict,)
+        return td
+
+
     class _TypedDictSpecialForm(_ExtensionsSpecialForm, _root=True):
         def __call__(
             self,
@@ -1089,60 +1151,22 @@ else:
             total=True,
             closed=None,
             extra_items=NoExtraItems,
-            __typing_is_inline__=False,
             **kwargs
         ):
-            if fields is _marker or fields is None:
-                if fields is _marker:
-                    deprecated_thing = (
-                        "Failing to pass a value for the 'fields' parameter"
-                    )
-                else:
-                    deprecated_thing = "Passing `None` as the 'fields' parameter"
-
-                example = f"`{typename} = TypedDict({typename!r}, {{}})`"
-                deprecation_msg = (
-                    f"{deprecated_thing} is deprecated and will be disallowed in "
-                    "Python 3.15. To create a TypedDict class with 0 fields "
-                    "using the functional syntax, pass an empty dictionary, e.g. "
-                ) + example + "."
-                warnings.warn(deprecation_msg, DeprecationWarning, stacklevel=2)
-                # Support a field called "closed"
-                if closed is not False and closed is not True and closed is not None:
-                    kwargs["closed"] = closed
-                    closed = None
-                # Or "extra_items"
-                if extra_items is not NoExtraItems:
-                    kwargs["extra_items"] = extra_items
-                    extra_items = NoExtraItems
-                fields = kwargs
-            elif kwargs:
-                raise TypeError("TypedDict takes either a dict or keyword arguments,"
-                                " but not both")
-            if kwargs:
-                if sys.version_info >= (3, 13):
-                    raise TypeError("TypedDict takes no keyword arguments")
-                warnings.warn(
-                    "The kwargs-based syntax for TypedDict definitions is deprecated "
-                    "in Python 3.11, will be removed in Python 3.13, and may not be "
-                    "understood by third-party type checkers.",
-                    DeprecationWarning,
-                    stacklevel=2,
-                )
+            return _create_typeddict(
+                typename,
+                fields,
+                total=total,
+                closed=closed,
+                extra_items=extra_items,
+                typing_is_inline=False,
+                **kwargs
+            )
 
-            ns = {'__annotations__': dict(fields)}
-            module = _caller(depth=5 if __typing_is_inline__ else 2)
-            if module is not None:
-                # Setting correct module is necessary to make typed dict classes
-                # pickleable.
-                ns['__module__'] = module
+        def __mro_entries__(self, bases):
+            return (_TypedDict,)
 
-            td = _TypedDictMeta(typename, (), ns, total=total, closed=closed,
-                                extra_items=extra_items)
-            td.__orig_bases__ = (TypedDict,)
-            return td
 
-    @_ensure_subclassable(lambda bases: (_TypedDict,))
     @_TypedDictSpecialForm
     def TypedDict(self, args):
         """A simple typed namespace. At runtime it is equivalent to a plain dict.
@@ -1198,9 +1222,14 @@ else:
             raise TypeError(
                 "TypedDict[...] should be used with a single dict argument"
             )
-
-        # Delegate to _TypedDictSpecialForm.__call__:
-        return self("<inlined TypedDict>", args[0], __typing_is_inline__=True)
+        return _create_typeddict(
+            "<inlined TypedDict>",
+            args[0],
+            typing_is_inline=True,
+            total=True,
+            closed=None,
+            extra_items=NoExtraItems
+        )
```

</details>

Then it would look like this:

```pycon
>>> from typing_extensions import TypedDict
>>> import inspect
>>> inspect.signature(TypedDict)
<Signature (typename, fields=<sentinel>, /, *, total=True, closed=None, extra_items=typing_extensions.NoExtraItems, **kwargs)>
>>> exit()
```

(and all your tests still pass with that diff applied!)

src/test_typing_extensions.py
```diff
@@ -5066,6 +5066,41 @@ def test_cannot_combine_closed_and_extra_items(self):
...
+
+    def test_inlined_not_a_dict(self):
+        with self.assertRaises(TypeError):
+            TypedDict["not_a_dict"]
+
+    def test_inlined_empty(self):
+        TD = TypedDict[{}]
+        self.assertEqual(TD.__required_keys__, set())
+
+    def test_inlined(self):
```
AlexWaygood, line 5081:
```suggestion
    def test_inline_too_many_arguments(self):
        with self.assertRaises(TypeError):
            TypedDict[{"a": int}, "extra"]

    def test_inline_not_a_dict(self):
        with self.assertRaises(TypeError):
            TypedDict["not_a_dict"]

    def test_inline_empty(self):
        TD = TypedDict[{}]
        self.assertEqual(TD.__required_keys__, set())

    def test_inline(self):
```

src/test_typing_extensions.py
AlexWaygood, line 1: It might be nice to add some explicit tests for `inspect.signature(TypedDict)` as well?

src/typing_extensions.py
```diff
@@ -1135,52 +1205,22 @@ class Point2D(TypedDict):
...
-                "understood by third-party type checkers.",
-                DeprecationWarning,
-                stacklevel=2,
+        # This runs when creating inline TypedDicts:
+        if not isinstance(args, tuple):
+            args = (args,)
+        if len(args) != 1 or not isinstance(args[0], dict):
+            raise TypeError(
+                "TypedDict[...] should be used with a single dict argument"
             )
```
AlexWaygood, line 1214: can't this be written more simply as:

```suggestion
        if not isinstance(args, dict):
            raise TypeError(
                "TypedDict[...] should be used with a single dict argument"
            )
```

(and also change the third argument passed to `_create_typeddict` a few lines below from `args[0]` to `args`)

All your tests seem to pass with that change!

AlexWaygood (commented):
This looks great. A couple more minor comments:

src/typing_extensions.py
Viicos: This is to support passing arguments to `__getitem__()` as a tuple:

```python
TypedDict[({...},)]
```

While it's not that useful in our case as only one argument can be provided, this aligns with other special forms where tuples are supported (see https://github.com/python/cpython/commit/a35a30509820f956d6feeaa0dbf42e9ca82c12bb, where this was done for `Annotated`).

I'll add a test for this.

AlexWaygood: Hrm, I'd be okay with just saying `TypedDict[({"whatever": int},)]` is disallowed, TBH! Especially until/unless we have an actual use-case for it. It seems like a pretty weird thing to be doing. And it's disallowed by other typing special forms that only accept a single argument -- for example:

```pytb
>>> from typing import Optional
>>> Optional[(int,)]
  File "<python-input-1>", line 1, in <module>
    Optional[(int,)]
    ~~~~~~~~^^^^^^^^
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 432, in inner
    return func(*args, **kwds)
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 576, in __getitem__
    return self._getitem(self, parameters)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 809, in Optional
    arg = _type_check(parameters, f"{self} requires a single type.")
  File "/Users/alexw/.pyenv/versions/3.13.1/lib/python3.13/typing.py", line 202, in _type_check
    raise TypeError(f"{msg} Got {arg!r:.100}.")
TypeError: typing.Optional requires a single type. Got (<class 'int'>,).
```

Viicos: Ah, wasn't aware that Optional and such did not accept that

AlexWaygood (approved):
LGTM, thank you! Can you add an entry to `CHANGELOG.md`?

src/test_typing_extensions.py
```diff
@@ -5066,6 +5066,47 @@ def test_cannot_combine_closed_and_extra_items(self):
...
+            ['typename', 'fields', 'total', 'closed', 'extra_items', 'kwargs']
+        )
+
+    def test_inline_too_many_arguments(self):
+        with self.assertRaises(TypeError):
+            TypedDict[{"a": int}, "extra"]
+
+    def test_inline_not_a_dict(self):
+        with self.assertRaises(TypeError):
+            TypedDict["not_a_dict"]
```
AlexWaygood, line 5081: Could also add a test for a single-element tuple, to make it clear that it's intentionally disallowed (at least until we have a use case)?

```suggestion
            TypedDict["not_a_dict"]

        # a tuple of elements isn't allowed, even if the first element is a dict:
        with self.assertRaises(TypeError):
            TypedDict[({"key": int},)]
```

src/test_typing_extensions.py
```diff
@@ -5066,6 +5066,63 @@ def test_cannot_combine_closed_and_extra_items(self):
...
+        self.assertNotIsSubclass(TD, collections.abc.Sequence)
+        self.assertTrue(is_typeddict(TD))
+        self.assertEqual(TD.__name__, "<inline TypedDict>")
+        self.assertEqual(
+            TD.__annotations__,
+            {"a": int, "b": Required[int], "c": NotRequired[int], "d": ReadOnly[int]},
+        )
+        self.assertEqual(TD.__module__, __name__)
+        self.assertEqual(TD.__bases__, (dict,))
+        self.assertTrue(TD.__total__)
```
JelleZijlstra, line 5115: I saw you changed some `assertIs(..., True)` to `assertTrue(...)`. Note that `assertTrue` just checks for truthiness, not that the value is exactly `True`. I'd prefer to use the `assertIs` form to make sure we don't put some other truthy object here.

JelleZijlstra (approved):

AlexWaygood (approved):

PR #581, opened 2025-04-08
Do not shadow user arguments in generated `__new__` by `@deprecated`

Viicos:
Backport of: https://github.com/python/cpython/pull/132160


JelleZijlstra:
The test fails because on versions where `warnings.deprecated` exists but not your fix, we use the stdlib version. This can be fixed by adjusting the condition for when we use the stdlib version.

JelleZijlstra:
Would you mind adding an entry to CHANGELOG.md? (I'd do it myself but then I'd have to ping somebody else to review.)

JelleZijlstra (approved):

PR #582, opened 2025-04-09
Add Reader and Writer protocols

srittau:
Cf. python/cpython#127648

#565 would help with testing this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Unreleased
 
+New features:
+
+- Add `typing_extensions.Reader` and `typing_extensions.Writer`. Patch by
+  Sebastian Rittau.
+
+Bugfixes:
+
 - Fix `TypeError` when taking the union of `typing_extensions.TypeAliasType` and a
   `typing.TypeAliasType` on Python 3.12 and 3.13.
   Patch by [Joren Hammudoglu](https://github.com/jorenham).
@@ -7,6 +14,7 @@
 # Release 4.13.1 (April 3, 2025)
 
 Bugfixes:
+
 - Fix regression in 4.13.0 on Python 3.10.2 causing a `TypeError` when using `Concatenate`.
   Patch by [Daraan](https://github.com/Daraan).
 - Fix `TypeError` when using `evaluate_forward_ref` on Python 3.10.1-2 and 3.9.8-10.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,13 @@
 # Unreleased
 
+- Drop support for Python 3.8 (including PyPy-3.8). Patch by [Victorien Plot](https://github.com/Viicos).
+
 New features:
 
-- Add `typing_extensions.Reader` and `typing_extensions.Writer`. Patch by
-  Sebastian Rittau.
-
-Bugfixes:
-
-- Drop support for Python 3.8 (including PyPy-3.8). Patch by [Victorien Plot](https://github.com/Viicos).
 - Add support for inline typed dictionaries ([PEP 764](https://peps.python.org/pep-0764/)).
   Patch by [Victorien Plot](https://github.com/Viicos).
+- Add `typing_extensions.Reader` and `typing_extensions.Writer`. Patch by
+  Sebastian Rittau.
 
 # Release 4.13.2 (April 10, 2025)
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -659,6 +659,18 @@ Protocols
 
    .. versionadded:: 4.6.0
 
+.. class:: Reader
+
+    See :py:class:`io.Reader`. Added to the standard library in Python 3.14.
+
+    .. versionadded:: 4.14.0
+
+.. class:: Writer
+
+    See :py:class:`io.Writer`. Added to the standard library in Python 3.14.
+
+    .. versionadded:: 4.14.0
+
 Decorators
 ~~~~~~~~~~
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,12 +1,20 @@
 # Unreleased
 
+New features:
+
+- Add `typing_extensions.Reader` and `typing_extensions.Writer`. Patch by
+  Sebastian Rittau.
+
+Bugfixes:
```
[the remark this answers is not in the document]
srittau: Sounds good.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+
 - Drop support for Python 3.8 (including PyPy-3.8). Patch by [Victorien Plot](https://github.com/Viicos).
 - Add support for inline typed dictionaries ([PEP 764](https://peps.python.org/pep-0764/)).
   Patch by [Victorien Plot](https://github.com/Viicos).
```
AlexWaygood, line 12: Neither of these are really bugfixes -- inline TypedDicts is a feature; dropping support for Python 3.8 is an "other changes" kind of thing IMO

srittau: Ah yes, this is a merge artifact. Will fix.

src/typing_extensions.py
```diff
@@ -846,6 +849,43 @@ def __round__(self, ndigits: int = 0) -> T_co:
...
+        __slots__ = ()
+
+        @abc.abstractmethod
+        def write(self, data: T_contra, /) -> int:
+            """Write data to the output stream and return the number of items written."""
+
+
+
+
 _NEEDS_SINGLETONMETA = (
```
AlexWaygood, line 889: there's maybe a few too many empty lines here :P

```suggestion
        def write(self, data: T_contra, /) -> int:
            """Write data to the output stream and return the number of items written."""


_NEEDS_SINGLETONMETA = (
```

src/test_typing_extensions.py
```diff
@@ -4103,6 +4103,32 @@ def foo(self): pass
...
+        class MyReader:
+            def read(self, n: int) -> bytes:
+                return b""
+
+        class WrongReader:
+            def readx(self, n: int) -> bytes:
+                return b""
+
+        assert isinstance(MyReader(), typing_extensions.Reader)
+        assert not isinstance(WrongReader(), typing_extensions.Reader)
```
AlexWaygood, line 4117:
```suggestion
        self.assertIsInstance(MyReader(), typing_extensions.Reader)
        self.assertNotIsInstance(WrongReader(), typing_extensions.Reader)
```

src/test_typing_extensions.py
```diff
@@ -4103,6 +4103,32 @@ def foo(self): pass
...
+        class MyWriter:
+            def write(self, b: bytes) -> int:
+                return 0
+
+        class WrongWriter:
+            def writex(self, b: bytes) -> int:
+                return 0
+
+        assert isinstance(MyWriter(), typing_extensions.Writer)
+        assert not isinstance(WrongWriter(), typing_extensions.Writer)
```
AlexWaygood, line 4129:
```suggestion
        self.assertIsInstance(MyWriter(), typing_extensions.Writer)
        self.assertNotIsInstance(WrongWriter(), typing_extensions.Writer)
```

JelleZijlstra (approved):

AlexWaygood:
Should we add a mention of these to https://github.com/python/typing_extensions/blob/main/doc/index.rst ?

doc/index.rst
```diff
@@ -659,6 +659,18 @@ Protocols
 
    .. versionadded:: 4.6.0
 
+.. class:: Reader
+
+    See :py:class:`io.Reader`. Added to the standard library in Python 3.14.
```
srittau, line 664: Please note that these intersphinx links will only start working after Python 3.14 is released.

AlexWaygood: that won't be till the Autumn -- maybe we should hardcode a link to https://docs.python.org/3.14/library/io.html#io.Reader for now, instead of using intersphinx?

srittau: Or we could point sphinx to 3.14 by default for now.

AlexWaygood: that works for me!

src/typing_extensions.py
```diff
@@ -846,6 +849,41 @@ def __round__(self, ndigits: int = 0) -> T_co:
...
+        This protocol only supports blocking I/O.
+        """
+
+        __slots__ = ()
+
+        @abc.abstractmethod
+        def read(self, size: int = ..., /) -> T_co:
+            """Read data from the input stream and return it.
+
+            If size is specified, at most size items (bytes/characters) will be
```
AlexWaygood, line 869: I'd either use italics or backticks here. Don't have a strong opinion about which; I think it will render fine either way in VSCode-generated tooltips etc.!

```suggestion
            If *size* is specified, at most *size* items (bytes/characters) will be
```

src/typing_extensions.py
```diff
@@ -846,6 +849,41 @@ def __round__(self, ndigits: int = 0) -> T_co:
...
+        """Protocol for simple I/O writer instances.
+
+        This protocol only supports blocking I/O.
+        """
+
+        __slots__ = ()
+
+        @abc.abstractmethod
+        def write(self, data: T_contra, /) -> int:
+            """Write data to the output stream and return the number of items written."""
```
AlexWaygood, line 884:
```suggestion
            """Write *data* to the output stream and return the number of items written."""
```

AlexWaygood (approved):

Issue #559, opened 2025-03-26
License information missing in 4.13 package on PyPi


martin-traverse:
Hi,  this update came through automatically and broke the license scan for my project. We had to downgrade to 4.12 until it is fixed.

### Steps to reproduce

    python -m venv ./venv
    . venv/bin/activate
    pip install typing-extensions
    pip install pip-licenses
    pip-licenses
    
    Name               Version  License 
    typing_extensions  4.13.0   UNKNOWN 

Key one to get fixed since this library is at the bottom of the stack for a lot of software! Please let me know if you need any extra info.

martin-traverse:
For reference, from the previous version:

    pip install "typing-extensions < 4.13" 

    Name               Version  License                            
    typing_extensions  4.12.2   Python Software Foundation License

brianschubert:
Hi! This is an issue with pip-licenses, not typing-extensions. It seems that pip-licenses does not support [PEP 639](https://peps.python.org/pep-0639/) license metadata, which typing-extensions uses since #507.

There's an open issue on pip-licenses tracker to add PEP 639 support: https://github.com/raimon49/pip-licenses/issues/225, plus an open PR: https://github.com/raimon49/pip-licenses/pull/213

Just as confirmation, I checked the typing-extensions 4.13.0 wheel on PyPI. It has the license info I would expect:
* `typing_extensions-4.13.0.dist-info/licenses/METADATA` contains the `License-Expression: PSF-2.0` and `License-File: LICENSE` fields
* `typing_extensions-4.13.0.dist-info/licenses/LICENSE` is present and populated with the right license file

Daraan:
(this is not an official reply)

With https://github.com/python/typing_extensions/pull/507 the behavior was changed to match [PEP 639](https://peps.python.org/pep-0639/) recommendations. `pip-license` is still expects the old format, its migration is still an open issue (https://github.com/raimon49/pip-licenses/issues/225) but a PR (https://github.com/raimon49/pip-licenses/pull/213) already exists; also a fork that implements it if you are interested.
However, maintenance seems low at the moment, best try to follow up with them.

srittau:
I just confirmed that correct license information is present in both the published .tar.gz file and the wheel. I am closing this per @brianschubert and @Daraan and hope that pip-licenses will be fixed soon.

martin-traverse:
Hi - the metadata in the wheel might be correct, but, it is a problem for us if there are no tools that actually support I!

[...]

PR #584, opened 2025-04-10
fix: update pyproject.toml with OSI approved license

scottschreckengaust:
This will add the OSI approved license to the package

https://github.com/pypa/trove-classifiers/blob/7e3e9e0c1135492b5a630d60d8abac15fb1d81ee/src/trove_classifiers/__init__.py#L336C5-L336C68

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Environment :: Console",
     "Intended Audience :: Developers",
+    "License :: OSI Approved :: Python Software Foundation License",
     "Operating System :: OS Independent",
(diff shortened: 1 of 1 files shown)


brianschubert:
Hi! This trove classifier was in fact recently removed in #530. Using trove classifiers for license metadata was [deprecated by PEP 639](https://peps.python.org/pep-0639/#deprecate-license-classifiers). Instead, typing_extensions now uses the PEP 639 `License-Expression` and `License-File` fields.

If you're suggesting this because a tool you're using doesn't work without the trove classifier, then the information in #559 and the issues linked therein may be helpful.

scottschreckengaust:
Thank you @brianschubert #559 is helpful as I was looking at `pip-licenses`.

Issue #568, opened 2025-04-01
Drop support for Python 3.8


JelleZijlstra:
Python 3.8 has been EOL for a while now and it's getting harder to get it running in CI. Let's drop support in the next feature release.

Viicos:
@JelleZijlstra can I work on this? If we go for the alternative implementation in https://github.com/python/typing_extensions/pull/580, it will make things easier.

JelleZijlstra:
Go ahead!

PR #585, opened 2025-04-16
Drop support for Python 3.8

Viicos:
Fixes https://github.com/python/typing_extensions/issues/568.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5157,7 +5157,7 @@ class AnnotatedTests(BaseTestCase):
         A.x = 5
         self.assertEqual(C.x, 5)
 
-    @skipIf(sys.version_info[:2] == (3, 10), "Waiting for bpo-46491 bugfix.")
+    @skipIf(sys.version_info[:2] == (3, 10), "Waiting for https://github.com/python/cpython/issues/90649 bugfix.")
     def test_special_form_containment(self):
         class C:
             classvar: Annotated[ClassVar[int], "a decoration"] = 4
@@ -5482,7 +5482,7 @@ class ParamSpecTests(BaseTestCase):
         self.assertTrue(hasattr(P, 'args'))
         self.assertTrue(hasattr(P, 'kwargs'))
 
-    @skipIf((3, 10, 0) <= sys.version_info[:3] <= (3, 10, 2), "Needs bpo-46676.")
+    @skipIf((3, 10, 0) <= sys.version_info[:3] <= (3, 10, 2), "Needs https://github.com/python/cpython/issues/90834.")
     def test_args_kwargs(self):
         P = ParamSpec('P')
         P_2 = ParamSpec('P_2')
@@ -7202,7 +7202,7 @@ class TypeVarTests(BaseTestCase):
             TypeVar('X', bound=(1, 2))
 
     def test_missing__name__(self):
-        # See bpo-39942
+        # See https://github.com/python/cpython/issues/84123
         code = ("import typing\n"
                 "T = typing.TypeVar('T')\n"
                 )
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -110,7 +110,6 @@ VT = TypeVar("VT")
 
 # Flags used to mark tests that only apply after a specific
 # version of the typing module.
-TYPING_3_9_0 = sys.version_info[:3] >= (3, 9, 0)
 TYPING_3_10_0 = sys.version_info[:3] >= (3, 10, 0)
 
 # 3.11 makes runtime type checks (_type_check) more lenient.
@@ -1779,8 +1778,7 @@ class GetUtilitiesTestCase(TestCase):
         self.assertIs(get_origin(List), list)
         self.assertIs(get_origin(Tuple), tuple)
         self.assertIs(get_origin(Callable), collections.abc.Callable)
-        if sys.version_info >= (3, 9):
-            self.assertIs(get_origin(list[int]), list)
+        self.assertIs(get_origin(list[int]), list)
         self.assertIs(get_origin(list), None)
         self.assertIs(get_origin(P.args), P)
         self.assertIs(get_origin(P.kwargs), P)
@@ -1817,20 +1815,18 @@ class GetUtilitiesTestCase(TestCase):
         self.assertEqual(get_args(List), ())
         self.assertEqual(get_args(Tuple), ())
         self.assertEqual(get_args(Callable), ())
-        if sys.version_info >= (3, 9):
-            self.assertEqual(get_args(list[int]), (int,))
+        self.assertEqual(get_args(list[int]), (int,))
         self.assertEqual(get_args(list), ())
-        if sys.version_info >= (3, 9):
-            # Support Python versions with and without the fix for
-            # https://bugs.python.org/issue42195
-            # The first variant is for 3.9.2+, the second for 3.9.0 and 1
-            self.assertIn(get_args(collections.abc.Callable[[int], str]),
-                          (([int], str), ([[int]], str)))
-            self.assertIn(get_args(collections.abc.Callable[[], str]),
-                          (([], str), ([[]], str)))
-            self.assertEqual(get_args(collections.abc.Callable[..., str]), (..., str))
+        # Support Python versions with and without the fix for
+        # https://bugs.python.org/issue42195
+        # The first variant is for 3.9.2+, the second for 3.9.0 and 1
+        self.assertIn(get_args(collections.abc.Callable[[int], str]),
+                        (([int], str), ([[int]], str)))
+        self.assertIn(get_args(collections.abc.Callable[[], str]),
+                        (([], str), ([[]], str)))
+        self.assertEqual(get_args(collections.abc.Callable[..., str]), (..., str))
         P = ParamSpec('P')
-        # In 3.9 and lower we use typing_extensions's hacky implementation
+        # In 3.9 we use typing_extensions's hacky implementation
         # of ParamSpec, which gets incorrectly wrapped in a list
         self.assertIn(get_args(Callable[P, int]), [(P, int), ([P], int)])
         self.assertEqual(get_args(Required[int]), (int,))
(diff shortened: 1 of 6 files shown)


Viicos:
Not sure if anything should be done here?

https://github.com/python/typing_extensions/blob/4525e9dbbd177b4ef8a84f55ff5fe127582a071d/src/typing_extensions.py#L1337-L1341

cc @Daraan

Daraan:
> Not sure if anything should be done here?
> 
> https://github.com/python/typing_extensions/blob/4525e9dbbd177b4ef8a84f55ff5fe127582a071d/src/typing_extensions.py#L1337-L1341
> 
> cc @Daraan

3.9 introduced `_UnionGenericAlias`. The code can now be adjusted to use it directly like this:

```diff
 def _could_be_inserted_optional(t): 
     """detects Union[..., None] pattern""" 
-     # 3.8+ compatible checking before _UnionGenericAlias 
-     if get_origin(t) is not Union: 
+     if not isinstance(t, typing._UnionGenericAlias):
         return False 
```

src/test_typing_extensions.py
```diff
@@ -5163,7 +5157,7 @@ class C:
         A.x = 5
         self.assertEqual(C.x, 5)
 
-    @skipIf(sys.version_info[:2] in ((3, 9), (3, 10)), "Waiting for bpo-46491 bugfix.")
+    @skipIf(sys.version_info[:2] == (3, 10), "Waiting for bpo-46491 bugfix.")
```
AlexWaygood, line 5160: While we're here, let's maybe update this to a GitHub reference? It looks like the old BPO issue was migrated to https://github.com/python/cpython/issues/90649, so that would be

```suggestion
    @skipIf(sys.version_info[:2] == (3, 10), "Waiting for CPython gh-90649 bugfix.")
```

src/typing_extensions.py
```diff
@@ -585,7 +568,7 @@ class _ProtocolMeta(type(typing.Protocol)):
         # but is necessary for several reasons...
         #
         # NOTE: DO NOT call super() in any methods in this class
-        # That would call the methods on typing._ProtocolMeta on Python 3.8-3.11
+        # That would call the methods on typing._ProtocolMeta on Python 3.9-3.11
```
AlexWaygood, line 571: this avoids us having to update this again when we drop support for Python 3.9 😆

```suggestion
        # That would call the methods on typing._ProtocolMeta on Python <=3.11
```

src/typing_extensions.py
```diff
@@ -786,7 +769,7 @@ def close(self): ...
 runtime = runtime_checkable
 
 
-# Our version of runtime-checkable protocols is faster on Python 3.8-3.11
+# Our version of runtime-checkable protocols is faster on Python 3.9-3.11
```
AlexWaygood, line 772:
```suggestion
# Our version of runtime-checkable protocols is faster on Python <=3.11
```

src/typing_extensions.py
```diff
@@ -864,17 +847,9 @@ def __round__(self, ndigits: int = 0) -> T_co:
...
-            }
-            t = type(func.__name__, (), cls_dict)
-            return functools.update_wrapper(t(), func)
-        else:
-            func.__mro_entries__ = mro_entries
-            return func
+    def inner(obj):
+        obj.__mro_entries__ = mro_entries
+        return obj
     return inner
```
AlexWaygood, line 853: We _could_ get rid of this `_ensure_subclassable` decorator altogether now, and monkey-patch `__mro_entries__` onto the `TypedDict` and `NamedTuple` functions manually (the same as CPython does, and the same as we used to do before https://github.com/python/typing_extensions/commit/dcdc53f62ecbea8a0fc9a42f66746f970ab101bb).

But aesthetically I sort-of prefer it as a decorator, so I'm happy to keep it the way you have it currently!

src/typing_extensions.py
```diff
@@ -1209,10 +1182,7 @@ class Point2D(TypedDict):
         td.__orig_bases__ = (TypedDict,)
         return td
 
-    if hasattr(typing, "_TypedDictMeta"):
-        _TYPEDDICT_TYPES = (typing._TypedDictMeta, _TypedDictMeta)
-    else:
-        _TYPEDDICT_TYPES = (_TypedDictMeta,)
+    _TYPEDDICT_TYPES = (typing._TypedDictMeta, _TypedDictMeta,)
```
AlexWaygood, line 1185: tiny nit

```suggestion
    _TYPEDDICT_TYPES = (typing._TypedDictMeta, _TypedDictMeta)
```

src/typing_extensions.py
```diff
@@ -1381,17 +1336,12 @@ def _clean_optional(obj, hints, globalns=None, localns=None):
                         localns = globalns
                 elif localns is None:
                     localns = globalns
-                if sys.version_info < (3, 9):
-                    original_value = ForwardRef(original_value)
                 else:
                     original_value = ForwardRef(
                         original_value,
                         is_argument=not isinstance(obj, _types.ModuleType)
                     )
```
AlexWaygood, line 1343: for 1:1 correspondence with the previous logic here, I think this should be

```suggestion

                original_value = ForwardRef(
                    original_value,
                    is_argument=not isinstance(obj, _types.ModuleType)
                )
```

but it's interesting that no tests failed...!

src/typing_extensions.py
```diff
@@ -1402,129 +1352,15 @@ def _clean_optional(obj, hints, globalns=None, localns=None):
...
-        # 3.9+
-        from typing import _BaseGenericAlias
-    except ImportError:
-        _BaseGenericAlias = typing._GenericAlias
-    try:
-        # 3.9+
-        from typing import GenericAlias as _typing_GenericAlias
-    except ImportError:
-        _typing_GenericAlias = typing._GenericAlias
+    from typing import GenericAlias as _typing_GenericAlias
```
AlexWaygood, line 1362: Hmm, I'm not sure this import ever did the thing it claims it was doing. I assumed that this was importing the `_GenericAlias` class from the `typing` module, but actually it's importing the `types.GenericAlias` class that's incidentally/accidentally re-exported from `typing`:

```pycon
>>> from typing import GenericAlias
>>> GenericAlias
<class 'types.GenericAlias'>
```

I think I'd prefer it if we just replaced all uses of `_typing_GenericAlias` with `types.GenericAlias`. It seems much less prone to confusion

src/typing_extensions.py
```diff
@@ -1543,7 +1379,7 @@ def get_origin(tp):
         """
         if isinstance(tp, _AnnotatedAlias):
             return Annotated
-        if isinstance(tp, (typing._GenericAlias, _typing_GenericAlias, _BaseGenericAlias,
+        if isinstance(tp, (_typing_GenericAlias, _BaseGenericAlias,
```
AlexWaygood, line 1382: is this change correct? As I mentioned above, `typing._GenericAlias` is a different class to `_typing_GenericAlias` -- the latter actually appears to be `types.GenericAlias`

AlexWaygood (commented):
Thank you, this overall looks great! I have a smattering of nits and a couple of substantive comments:

src/typing_extensions.py
JelleZijlstra: Seems like we need more tests here too

Viicos: The existing `isinstance()` check had `typing._GenericAlias`, which was useless as `typing._GenericAlias` is a subclass of `_BaseGenericAlias` and this one is already in the tuple.

Thus I removed `typing._GenericAlias`, matching [the current implementation](https://github.com/python/cpython/blob/3.9/Lib/typing.py#L1538).

src/typing_extensions.py
```diff
@@ -1,3 +1,4 @@
+# pyright: ignore
```
AlexWaygood, line 1:
```suggestion
```

src/typing_extensions.py
```diff
@@ -1543,7 +1378,7 @@ def get_origin(tp):
         """
         if isinstance(tp, _AnnotatedAlias):
```
AlexWaygood, line 1379: I think we may as well just use the fully qualified form, and then we can get rid of `from typing import _AnnotatedAlias, _BaseGenericAlias` statement a few lines above altogether 

```suggestion
        if isinstance(tp, typing._AnnotatedAlias):
```

src/typing_extensions.py
```diff
@@ -1543,7 +1378,7 @@ def get_origin(tp):
...
             return Annotated
-        if isinstance(tp, (typing._GenericAlias, _typing_GenericAlias, _BaseGenericAlias,
+        if isinstance(tp, (_types.GenericAlias, _BaseGenericAlias,
```
AlexWaygood, line 1381:
```suggestion
        if isinstance(tp, (_types.GenericAlias, typing._BaseGenericAlias,
```

src/typing_extensions.py
```diff
@@ -1563,9 +1398,7 @@ def get_args(tp):
         """
         if isinstance(tp, _AnnotatedAlias):
```
AlexWaygood, line 1399:
```suggestion
        if isinstance(tp, typing._AnnotatedAlias):
```

src/typing_extensions.py
Viicos: We do have one already:

https://github.com/python/typing_extensions/blob/4525e9dbbd177b4ef8a84f55ff5fe127582a071d/src/test_typing_extensions.py#L1783

AlexWaygood: Thanks!

src/typing_extensions.py
```diff
@@ -1541,9 +1373,9 @@ def get_origin(tp):
             get_origin(List[Tuple[T, T]][int]) == list
             get_origin(P.args) is P
         """
-        if isinstance(tp, _AnnotatedAlias):
+        if isinstance(tp, typing._AnnotatedAlias):
             return Annotated
-        if isinstance(tp, (typing._GenericAlias, _typing_GenericAlias, _BaseGenericAlias,
+        if isinstance(tp, ( typing._BaseGenericAlias,
```
AlexWaygood, line 1378:
```suggestion
        if isinstance(tp, (typing._BaseGenericAlias,
```

AlexWaygood: I think my edited suggestion should fix the failing test on 3.9 :-)

AlexWaygood (approved):
LGTM

PR #592, opened 2025-05-05
Fix tests on Python 3.14

JelleZijlstra:
* Support deferred evaluation in `typing_extensions.TypedDict`
* Account for ForwardRef caching being removed (python/cpython#129465). `test_nested_strings` in our test suite was passing accidentally due to ForwardRef caching.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -5152,6 +5152,65 @@ class TypedDictTests(BaseTestCase):
         self.assertIs(type(inst), dict)
         self.assertEqual(inst["a"], 1)
 
+    def test_annotations(self):
+        # _type_check is applied
+        with self.assertRaisesRegex(TypeError, "Plain typing.Final is not valid as type argument"):
+            class X(TypedDict):
+                a: Final
+
+        # _type_convert is applied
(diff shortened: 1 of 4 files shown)


src/test_typing_extensions.py
```diff
@@ -103,6 +103,11 @@
     runtime_checkable,
 )
 
+if sys.version_info >= (3, 14):
+    from test.support import EqualToForwardRef
```
AlexWaygood, line 107: this is going to mean that the tests can't be run on e.g. a Conda build -- Conda builds don't include the `test` module in the stdlib

JelleZijlstra: I guess I can inline it

srittau (approved):
LG to my typing-internals-untrained eyes

AlexWaygood (approved):

cdce8p:
I'm currently looking into making https://github.com/Fatal1ty/mashumaro compatible with Python 3.14.0b1. Was wondering why a lot of tests with `typing_extensions.TypedDict` failed until I discovered this change.

It would probably make testing for downstream projects a lot easier if the changes here could be released. Even just a pre-release would help.

Issue #591, opened 2025-05-05
Add support for PEP 661 sentinels


JelleZijlstra:
We should add a PEP 661-style Sentinel class; see https://discuss.python.org/t/pep-661-sentinel-values/9126/243

cc @Viicos @taleinat.

(For clarity: I won't work on this myself but I'll review a PR.)

Viicos:
@taleinat I can look at it this week and port your draft implementation linked in the PEP there.

taleinat:
The PEP has been [deferred to Python 3.15](https://discuss.python.org/t/pep-661-sentinel-values/9126/243), so I don't think that should be added yet.

Viicos:
Yes, but as with typing PEPs, we don't necessarily have to wait for 3.15 to be released (assuming the PEP would be accepted) to have it included in typing-extensions.

To give more context, having it added here early (even if the PEP is still in draft) can allow users/libraries (such as in Pydantic) to experiment with it, and potentially catch issues or provide valuable feedback. Of course, this won't be added in Pydantic as a stable feature until the PEP gets accepted.

AlexWaygood:
Yes, we usually add experimental versions of not-yet-accepted typing features well in advance of their PEPs being accepted

taleinat:
I wasn't aware of that, cool! In that case, feel free to go ahead.

I'll mention that it seems likely that the internal registry of sentinels will be removed, to be replaced by simple attribute lookup on the relevant module (perhaps with a mechanism to avoid threading race conditions). In this case, sentinels may be required to be set as module-level attributes.

And in any case, the name of the sentinel will be required to exactly match the name of the variable it is assigned to; a mismatch should be considered an error.

PR #594, opened 2025-05-08
Add support for sentinels (PEP 661)

Viicos:
Fixes https://github.com/python/typing_extensions/issues/591.

Regarding the pickling behavior, no matter if the sentinel is defined at the module level or in a nested scope, this seems tricky to handle. Ideally, I'd like to have the following working:

```python
sentinel = Sentinel('sentinel')

class A:
    def __init__(self):
        self.s = sentinel

a = A()
a.s is sentinel  # True

import pickle

a_r = pickle.loads(pickle.dumps(a))

a_r.s is sentinel  # False
```

Considering unplicking data can be done across interpreter runs, this isn't really achievable, but still will be surprising.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -4238,10 +4238,10 @@ class Sentinel:
             raise TypeError(f"{type(self).__name__!r} object is not callable")
 
     def __or__(self, other):
-        return Union[self, other]
+        return typing.Union[self, other]
 
     def __ror__(self, other):
-        return Union[other, self]
+        return typing.Union[other, self]
 
     def __getstate__(self):
         raise TypeError(f"Cannot pickle {type(self).__name__!r} object")
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -4238,7 +4238,7 @@ class Sentinel:
     def __init__(
         self,
         name: str,
-        repr: str | None = None,
+        repr: typing.Optional[str] = None,
     ):
         self._name = name
         self._repr = repr if repr is not None else f'<{name.split(".")[-1]}>'
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -1,4 +1,3 @@
-# pyright: ignore
 import abc
 import builtins
 import collections
@@ -4227,12 +4226,10 @@ else:
 class Sentinel:
     """Create a unique sentinel object.
 
-    *name* should be the fully-qualified name of the variable to which the
-    return value shall be assigned.
+    *name* should be the name of the variable to which the return value shall be assigned.
 
     *repr*, if supplied, will be used for the repr of the sentinel object.
-    If not provided, "<name>" will be used (with any leading class names
-    removed).
+    If not provided, "<name>" will be used.
     """
 
     def __init__(
@@ -4241,7 +4238,7 @@ class Sentinel:
         repr: typing.Optional[str] = None,
     ):
         self._name = name
-        self._repr = repr if repr is not None else f'<{name.split(".")[-1]}>'
+        self._repr = repr if repr is not None else f'<{name}>'
 
     def __repr__(self):
         return self._repr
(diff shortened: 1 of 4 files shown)


TeamSpen210:
For pickling, there's some builtin behaviour which might be perfect for this. If `__reduce__` returns the sentinel name, and `__module__` is defined (via `sys._getframemodulename()`), `pickle` will import the module then lookup attributes to find the value. It'll even check and error on pickling if the sentinel isn't present there. This'll make your example work, by looking up `__main__.sentinel`. If that global wasn't present, `self.s = Sentinel("a.s")` would also work.

JelleZijlstra:
I'm worried the pickling behavior the PEP wants may change over time. One option would be to make them not pickleable for now. That way, we can always add pickling as a feature later, but if we add pickling now and need to change the behavior later, that may cause issues for users.

Viicos:
Yes anyway this will be better discussed in the PEP thread. I removed the `__reduce__()` implementation, let me know if you actually want to raise an error.

Viicos:
Yes, makes sense, added a type error following prior art: https://github.com/search?q=repo%3Apython%2Fcpython+%22Cannot+pickle%22&type=code.

src/typing_extensions.py
```diff
@@ -1,3 +1,4 @@
+# pyright: ignore
```
JelleZijlstra, line 1:
```suggestion
```

src/typing_extensions.py
```diff
@@ -4222,6 +4224,44 @@ def evaluate_forward_ref(
         )
 
 
+class Sentinel:
+    """Create a unique sentinel object.
+
+    *name* should be the fully-qualified name of the variable to which the
```
JelleZijlstra, line 4230: Fully-qualified? That doesn't match the PEP which does `MISSING = Sentinel("MISSING")`.

JelleZijlstra (commented):
Thanks! Could you also add this to the documentation?

src/typing_extensions.py
Viicos: This was copied verbatim from https://github.com/taleinat/python-stdlib-sentinels/, updated.

Viicos: Actually the repr construction by splitting dots don't make sense anymore, so I also updated the logic in the last logic.

JelleZijlstra (approved):

src/typing_extensions.py
TeamSpen210: It'd make sense to be fully qualified inside the module, so `Sentinel("SomeClass.CONST")`. Pickle expects that if we decide to use it. It's recommended in the PEP under "Additional Notes", but not required.

JelleZijlstra:
Would you mind fixing the lint? I think we need to make `Union` unconditionally defined.

src/typing_extensions.py
Viicos: > It'd make sense to be fully qualified inside the module, so `Sentinel("SomeClass.CONST")`. Pickle expects that if we decide to use it. It's recommended in the PEP under "Additional Notes", but not required.

This will have to be decided in the PEP discussion.

JelleZijlstra (approved):

HexDecimal:
Sorry. I've used dataclasses so often that I forgot the default behavior of Python classes. Here I am making several wrong assumptions about how equality works.

> We'll likely add Pickle support later but I do not believe it is essential.

I've used sentinels in my own Python projects and pickling them is essential. I currently use [sentinel-value](https://pypi.org/project/sentinel-value/) but I'd wish to use something more standard. I don't really need to be a hurry though.

> We're waiting for the PEP author to decide what it will look like.

Does this count? https://github.com/python/typing_extensions/issues/591#issuecomment-2863310895
I think attribute lookup appears to be very forward compatible, one could seamlessly replace a `typing_extensions.Sentinel` definition with a `typing.Sentinel` and the replacement class will be loaded instead of the class originally pickled.

PR #596, opened 2025-05-10
Update PEP 649/749 implementation

JelleZijlstra:
Now that PEP 749 has been accepted we can just rely on annotationlib.

This also updates the values in the Format enum and the default value for some `format` arguments.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -952,9 +952,19 @@ Enums
       for the annotations. This format is identical to the return value for
       the function under earlier versions of Python.
 
+   .. attribute:: VALUE_WITH_FAKE_GLOBALS
+
+      Equal to 2. Special value used to signal that an annotate function is being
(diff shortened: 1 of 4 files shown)


srittau (approved):

doc/index.rst
```diff
@@ -796,7 +796,7 @@ Functions
    This parameter must be provided (though it may be an empty tuple) if *owner*
    is not given and the forward reference does not already have an owner set.
    *format* specifies the format of the annotation and is a member of
-   the :class:`Format` enum.
+   the :class:`Format` enum, defaulting :attr:`Format.VALUE`.
```
AlexWaygood, line 799:
```suggestion
   the :class:`Format` enum, defaulting to :attr:`Format.VALUE`.
```

AlexWaygood (approved):

AlexWaygood (approved):

Issue #528, opened 2025-02-10
Discussion: Upgrade repo version after a release to `dev` to not match release version


Daraan:
A while ago to do some tests I wanted to upgrade my current `typing_extensions` version to the current GitHub version by using
`pip install git+https://github.com/python/typing_extensions.git`.

I also blame `pip` here because it does not tell that it doesn't install anything. `pip` did not perform an upgrade because the current release, which I had already installed, has the same *version* as the GitHub repo (4.12.2).  
Only after a while I realized that nothing was upgraded and I had to use `--force-reinstall` to do it.

I want to discuss if in the future, after some merged PRs, the version of the GitHub repo should be upgraded to a +1dev to not be equivalent to the latest release version.

JelleZijlstra:
Sounds good. We can update the release instructions in the repo to mention this.

PR #601, opened 2025-05-21
Add development versioning scheming

Daraan:
In relation to #528 I've written a proposal on how and vaguely when to add development versions.

I've oriented myself on the [packaging](https://packaging.python.org/en/latest/discussions/versioning/#valid-version-numbers) versioning, and decided against the hyphen `-` suffix for pre-releases.

Further input and ideas are welcome.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,10 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
+## Development versions
+After meaningful updates, but without concrete plans for a release, the version can pre increased once in [pyproject.toml](/pyproject.toml) file and appended with a `.dev` suffix, e.g. `4.0.1.dev`.
+Further development versions adjust their suffix in following increasing way `.dev1`, `.dev2`, ...
+
 # Type stubs
 
 A stub file for `typing_extensions` is maintained
@@ -54,7 +58,7 @@ may have installed.
 # Workflow for PyPI releases
 
 - Make sure you follow the versioning policy in the documentation
-  (e.g., release candidates before any feature release)
+  (e.g., release candidates before any feature release, do not release development versions)
 
 - Ensure that GitHub Actions reports no errors.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,8 +25,8 @@ Starting with version 4.0.0, `typing_extensions` uses
 for more detail.
 
 ## Development versions
-After meaningful updates, but without concrete plans for a release, the version can pre increased once in [pyproject.toml](/pyproject.toml) file and appended with a `.dev` suffix, e.g. `4.0.1.dev`.
-Further development versions adjust their suffix in following increasing way `.dev1`, `.dev2`, ...
+After a release the version is increased once in the [pyproject.toml](/pyproject.toml) file and appended with a `.dev` suffix, e.g. `4.0.1.dev`.
+Further subsequent updates, i.e. to `.dev2`, are not planned between releases.
 
 # Type stubs
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,8 +24,9 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
-## Development versions
-After a release the version is increased once in the [pyproject.toml](/pyproject.toml) file and appended with a `.dev` suffix, e.g. `4.0.1.dev`.
+## Development version
+After a release the version is increased once in [pyproject.toml](/pyproject.toml) and
+appended with a `.dev` suffix, e.g. `4.0.1.dev`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -24,6 +24,10 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
+## Development versions
+After meaningful updates, but without concrete plans for a release, the version can pre increased once in [pyproject.toml](/pyproject.toml) file and appended with a `.dev` suffix, e.g. `4.0.1.dev`.
+Further development versions adjust their suffix in following increasing way `.dev1`, `.dev2`, ...
```
srittau, line 29: I would suggest that we (and by that I mean our release manager, @JelleZijlstra), increase the version to `.dev` after each release, even without any updates in between. This makes is clear that the current version is not identical to the released PyPI package, and we don't forget to increment the version number after a "meaningful" update.

CONTRIBUTING.md
```diff
@@ -24,6 +24,11 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
+## Development version
+After a release the version is increased once in [pyproject.toml](/pyproject.toml) and
+appended with a `.dev` suffix, e.g. `4.0.1.dev`.
```
brianschubert, line 29: In other projects, I've seen dev versions be named relative to the next "planned release", which can mean either a patch release or a major (feature) release. So for example, after releasing `4.0.1`, the version number might get bumped to `4.1.0.dev`. I think that's how [htop does it](https://github.com/htop-dev/htop/commit/3b4f9bad10d462310442aff078f761c58bfc103c) as an example. I'm not sure whether doing something similar here would make sense or not. 

FWIW, my personal weak preference is using a `-dev` suffix for generic dev version numbers that don't get bumped. The same style is used in e.g. [mypy_extensions](https://github.com/python/mypy_extensions/blob/8d272bb1264a6114610a3df9b43f87247c856d92/pyproject.toml#L7). It's a purely cosmetic difference -- `-dev`, `.dev`, `dev0`, etc all get [normalized to the exact same version number](https://packaging.python.org/en/latest/specifications/version-specifiers/#development-release-separators).

CONTRIBUTING.md
```diff
@@ -68,3 +73,6 @@ may have installed.
 
 - Release automation will finish the release. You'll have to manually
   approve the last step before upload.
+
+- After the release has been published on PyPI the version of the repositiory
+  is upgraded to the next [development versions](#development-version).
```
brianschubert, line 78: As a sanity check for whoever is making the release, I think it would be helpful to include an example of how the version number should be updated. E.g. "After making the release, update the version number in `typing_extensions/pyproject.toml` to a `dev` version of the next planned release. For example, change `4.1.1` to `4.1.2.dev`."

CONTRIBUTING.md
```diff
@@ -54,7 +59,7 @@ may have installed.
 # Workflow for PyPI releases
 
 - Make sure you follow the versioning policy in the documentation
-  (e.g., release candidates before any feature release)
+  (e.g., release candidates before any feature release, do not release development versions)
```
brianschubert, line 62: I'm not sure this reminder is necessary. Would someone even think of doing this? I don't think creating a dev release accidentally is a concern, since the rest of the instructions (e.g. making a tag) would make that hard

CONTRIBUTING.md
```diff
@@ -24,6 +24,11 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
+## Development version
+After a release the version is increased once in [pyproject.toml](/pyproject.toml) and
+appended with a `.dev` suffix, e.g. `4.0.1.dev`.
+Further subsequent updates, i.e. to `.dev2`, are not planned between releases.
+
```
brianschubert, line 31: I think this is implied by the previous description. Talking about what the project _doesn't_ do might be confusing

CONTRIBUTING.md
Daraan: Thank you very much for the input and references, also sounds good and logical to me it had also crossed my mind. At that point it made more sense to me to avoid a downgrade in between:  
 `4.0.0 (release) -> 4.1.0.dev -> 4.0.1 (patch-release)` vs. subsequent increase even if there is not patch  
 `4.0.0 (release) -> 4.0.1.dev -> 4.1.0 (no-patch-release)`.  
Subsequent would allow `pip install --upgrade typing_extensions` to work as expected. On the other hand, in this situation pip will at least tell you that the requirement is already satisfied with `4.1.0.dev` and you see that no upgrade to `4.0.1` was done.

I think we need to hear from the others first as well.

CONTRIBUTING.md
Daraan: Good point with the tag. I also would not expect someone doing it, just explicit better an implicit?

In case we always increase in at least minor steps (`4.1.0dev`) I think we should note here or above that `4.1.0.dev -> 4.0.1` is the desired progress in case of patch versions.

PR #602, opened 2025-05-24
More fixes for 3.14 and 3.15

JelleZijlstra:
python/cpython#133772 (TypedDict change)

python/cpython#133822 (NamedTuple change for 3.15)

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -4402,6 +4402,38 @@ class TypedDictTests(BaseTestCase):
             'voice': str,
         }
 
+    def test_inheritance_pep563(self):
+        def _make_td(future, class_name, annos, base, extra_names=None):
+            lines = []
+            if future:
+                lines.append('from __future__ import annotations')
+            lines.append('from typing import TypedDict')
+            lines.append(f'class {class_name}({base}):')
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -4402,6 +4402,39 @@ class Cat(Animal):
...
+                lines.append('from __future__ import annotations')
+            lines.append('from typing import TypedDict')
+            lines.append(f'class {class_name}({base}):')
+            for name, anno in annos.items():
+                lines.append(f'    {name}: {anno}')
+            code = '\n'.join(lines)
+            ns = {**extra_names} if extra_names else {}
+            exec(code, ns)
+            return ns[class_name]
+
```
srittau, line 4419: This is all very magical. Wouldn't it make more sense to just hard-code these four fairly short code samples?

srittau: Ah, it's eight. So maybe use a template at least?

JelleZijlstra: I feel that would be more complicated; it would require conditionals on half the lines of the template and `textwrap.dedent`.

srittau (approved):

PR #605, opened 2025-05-25
Remove PEP-604 methods from `Sentinel` on Python <3.10

AlexWaygood:
We don't generally try to "backport PEP 604" on Python <3.10; this is more consistent with our features

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Unreleased
+
+- Remove `__or__` and `__ror__` methods from `typing_extensions.Sentinel`
+  on Python versions <3.10. PEP 604 was introduced in Python 3.10, and
+  `typing_extensions` does not generally attempt to backport PEP-604 methods
+  to prior versions.
+
 # Release 4.14.0rc1 (May 24, 2025)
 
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Could you add a note to the changelog?

AlexWaygood:
> Note that `typing_extensions._SpecialForm` also unconditionally has `__or__`.

Yes, the word "generally" in the changelog entry is doing some work here 😄 as you say, though, `_SpecialForm.__(r)or__` is probably not doing much harm, and it might cause backwards-compat issues to remove it now 🤷

AlexWaygood:
> It's suspicious that no tests failed, would you mind adding a simple 3.10+ test that `__or__` works?

it actually is already tested on Python 3.10+, but the test is already skipped on Python <=3.9!

https://github.com/python/typing_extensions/blob/44de568f73a93f29e52c2fc2d5f149305a4a3bae/src/test_typing_extensions.py#L9148-L9156

JelleZijlstra (approved):

Viicos:
Good catch, I always fall into this recursion trap!

PR #607, opened 2025-05-25
Use pre-commit for linting

AlexWaygood:
This PR adds a pre-commit-config file to the repo and removes our linting GitHub Actions workflow. It shouldn't be merged unless we:
1. Agree that we want this!
2. Either:
  a. Enable pre-commit.ci on this repo or
  b. Replace the GitHub Actions workflow with one that uses https://github.com/pre-commit/action

My main motivation for proposing this change is that there are various linting tools that I think would be useful for this project that exist conveniently as pre-commit hooks, and which can be run easily in isolated environments via pre-commit. There are actually several other tools that I'd like to add (zizmor, a security linter, would be very relevant here, I think) but which have several complaints on the repo as-is -- if this is accepted, I'll add them in followup PRs so it's easy to identify exactly why we're making each change.

The downside of this change is that we're once again using tools written in Python to lint this repo. pre-commit doesn't currently appear to have a direct or indirect dependency on typing_extensions, but that could easily change in the future. It's fairly easy to simply run pre-commit using uv or pipx to ensure that pre-commit itself is installed into a standalone virtual environment, but this does require a little bit more explanation in our CONTRIBUTING guide.


AlexWaygood:
Is this something other maintainers think would be useful? Or should I close this? :-)

JelleZijlstra (approved):
Let's do it!

srittau (approved):

PR #616, opened 2025-06-04
Do not use slots for `_TypedDictSpecialForm`

Viicos:
Fixes https://github.com/python/typing_extensions/issues/615.


AlexWaygood (approved):
Thanks!

PR #617, opened 2025-06-04
Support for pickling sentinel objects as singletons

### LINKED ISSUE #720: Missing pickle support for Sentinel
Support for pickle is expressed as a motivation in [PEP 661](https://peps.python.org/pep-0661/), but currently `typing_extensions.Sentinel` implements an explicit rejection of pickle support.

The reason for the rejection was that the suggested pickle implementation could change in the future, **but the solution to that would've been to use pickles singleton support which would've deferred any specific pickle implementation to any hypothetical future object including swapping out the typing_extensions sentinel with a future builtin sentinel.** It was understandable to worry about this since I've seen many others get it wrong, but **singletons are inherently forward compatible.**

Pickling as a singleton was suggested right away by TeamSpen210 in #594, but was ignored possibly due to being too contrived to understand possibly due to singletons themselves not being a well documented part of pickle.

I've made PR's to fix this but I may have given the impression that I'm just making them for fun, but I actually need pickle support. I strongly want to move away from sentinel libraries which have a poor reduce implementation which locks pickled data to that library.

Obviously the default for pickling instances is bad and was rightly rejected, and many of the typical options to reduce an instance are also rightly rejected, but singletons simply reference an objects fully qualified `__module__.__name__` and do not care what that object is later replaced with.

The solution to support singletons is to either assign `self.__module__` and return the sentinels qualified name as lone string from `__reduce__` like in PR #617, or to treat new sentinels as types rather than instances and rely on the singleton support inherit in pickling types like in PR #715.

### LINKED ISSUE #742: Adapt Sentinel implementation to accepted version of PEP 661
PEP-661 was accepted, but it is implemented a little differently than our `typing_extensions.Sentinel`. Let's see how we can make the typing-extensions version better match the version that will be in Python 3.15, while keeping backward compatibility for users who may already be using `typing_extensions.Sentinel`.

Here are the relevant differences:

* We have `typing_extensions.Sentinel`; 3.15 will have `builtins.sentinel` (lowercase). Let's not worry about the name. On Python 3.15 we'll re-export `builtins.sentinel` as `typing_extensions.Sentinel`.
* Our Sentinel is subclassable; `builtins.sentinel` will not be. Let's add an `__init_subclass__` that emits a DeprecationWarning.
* We added an extra `repr` argument to control the repr. Let's emit a DeprecationWarning if it is passed.
* The default repr is different: `<name>` for us, just `name` for the builtin. Let's just change this.
* We made our Sentinel not picklable; the builtin is picklable by name. So let's add a `__module__` attribute inferred from the caller (can reuse `_set_module`) and a `__reduce__` that just returns `self.__name__`.
* The builtin doesn't support weakrefs; our version does because it's in Python. I don't know of a way to deprecate our support, so let's just document this difference.
* The builtin exposes a `__name__` attribute; we have `_name`. Again let's just change this.
* There are some observable changes from the builtin being in C (e.g., the type is immutable), but let's not worry about that.

There are three open PRs about changing Sentinel. #715 and #732 make changes that introduce further incompatibilities, so I'm going to close those. #617 goes in the right direction but also introduces some further incompatible changes.

HexDecimal:
My attempt at implementing [PEP 661](https://peps.python.org/pep-0661/) since I was unhappy with #594. I'm hoping that this is a PR of decent quality and not just me desperately wanting pickle support.

Changes made to Sentinel:

- ~~Added `module_name` parameter following PEP implementation and tweaking that to use the local `_caller` helper function. Breaking change due to `repr` becoming a keyword parameter.~~ `module_name` removed due to changes in PEP 661.
- `name` requires qualified names to support pickle singletons, this is tested and documented.
- Added `__reduce__` to track Sentinel by name and module_name.
- Added copy and pickle tests. There were several other use-cases for pickle from the PEP 661 discussion. The PEP requires that the sentinel identity is preserved during these operations.
- Updated documentation for Sentinel.

For a while I still supported the `repr` parameter and after following the [PEP 661 discussion](https://discuss.python.org/t/pep-661-sentinel-values/9126) I ended up implementing the truthiness tweaks mentioned there, but then I ended up scrapping all of that so that I could follow the PEP more closely, but they would be easy to reintroduce later if desired.

I'm unsure of how to mention version changes in the docs. Replacing `repr` with `module_name` is technically a breaking change.

Closes #720
Closes #742

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1076,9 +1076,6 @@ Sentinel objects
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
 
-   A sentinel is bound to the module it is created within,
-   sentinels are not equal to similar named sentinels from other modules.
-
    Assigning attributes to a sentinel including `__weakref__` is forbidden.
 
    Example::
@@ -1093,16 +1090,18 @@ Sentinel objects
       ...
       >>> func(MISSING)
 
-   Sentinels defined inside a class scope should use a :term:`qualified name`.
-
-   Example::
-
-      >>> class MyClass:
-      ...     MISSING = sentinel('MyClass.MISSING')
-
    .. versionadded:: 4.14.0
 
       See :pep:`661`
+
+   .. versionchanged:: 4.16.0
+
+      Now supports pickle and will be reduced as a singleton.
+      Renamed from `Sentinel` to `sentinel`, `Sentinel` is deprecated.
+      Automatic `repr` string no longer has angle brackets.
+      `repr` parameter was deprecated.
+      `name` as a keyword is deprecated.
+      Subclasssing and attribute assignment are deprecated.
 
 
 Pure aliases
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1076,7 +1076,7 @@ Sentinel objects
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
 
-   Assigning attributes to a sentinel including `__weakref__` is forbidden.
+   Assigning attributes to a sentinel is deprecated.
 
    Example::
 
@@ -1096,12 +1096,15 @@ Sentinel objects
 
    .. versionchanged:: 4.16.0
 
+      The implementation of this class has been updated to conform to
+      the accepted version of :pep:`661`.
+
       Now supports pickle and will be reduced as a singleton.
       Renamed from `Sentinel` to `sentinel`, `Sentinel` is deprecated.
       Automatic `repr` string no longer has angle brackets.
       `repr` parameter was deprecated.
       `name` as a keyword is deprecated.
-      Subclasssing and attribute assignment are deprecated.
+      Subclassing and attribute assignment are deprecated.
 
 
 Pure aliases
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1071,16 +1071,15 @@ Capsule objects
 Sentinel objects
 ~~~~~~~~~~~~~~~~
 
-.. class:: sentinel(name, /, *, repr=None)
+.. class:: sentinel(name, /)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
 
-   If *repr* is provided, it will be used for the :meth:`~object.__repr__`
-   of the sentinel object. If not provided, ``"<name>"`` will be used.
-
    A sentinel is bound to the module it is created within,
    sentinels are not equal to similar named sentinels from other modules.
+
+   Assigning attributes to a sentinel including `__weakref__` is forbidden.
 
    Example::
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1071,7 +1071,7 @@ Capsule objects
 Sentinel objects
 ~~~~~~~~~~~~~~~~
 
-.. class:: Sentinel(name, module_name=None, *, repr=None)
+.. class:: Sentinel(name, *, module_name=None, repr=None)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1071,10 +1071,13 @@ Capsule objects
 Sentinel objects
 ~~~~~~~~~~~~~~~~
 
-.. class:: Sentinel(name, repr=None)
+.. class:: Sentinel(name, module_name=None, *, repr=None)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
+
+   *module_name* is the module where the sentinel is defined.
+   Defaults to the current modules ``__name__``.
 
    If *repr* is provided, it will be used for the :meth:`~object.__repr__`
    of the sentinel object. If not provided, ``"<name>"`` will be used.
@@ -1090,6 +1093,13 @@ Sentinel objects
       ...         assert_type(arg, int)
       ...
       >>> func(MISSING)
+
+   Sentinels defined inside a class scope should use a :term:`qualified name`.
+
+   Example::
+
+      >>> class MyClass:
+      ...     MISSING = Sentinel('MyClass.MISSING')
 
    .. versionadded:: 4.14.0
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1071,21 +1071,21 @@ Capsule objects
 Sentinel objects
 ~~~~~~~~~~~~~~~~
 
-.. class:: Sentinel(name, *, module_name=None, repr=None)
+.. class:: sentinel(name, /, *, repr=None)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
 
-   *module_name* is the module where the sentinel is defined.
-   Defaults to the current modules ``__name__``.
-
    If *repr* is provided, it will be used for the :meth:`~object.__repr__`
    of the sentinel object. If not provided, ``"<name>"`` will be used.
 
+   A sentinel is bound to the module it is created within,
+   sentinels are not equal to similar named sentinels from other modules.
+
    Example::
 
-      >>> from typing_extensions import Sentinel, assert_type
-      >>> MISSING = Sentinel('MISSING')
+      >>> from typing_extensions import sentinel, assert_type
+      >>> MISSING = sentinel('MISSING')
       >>> def func(arg: int | MISSING = MISSING) -> None:
       ...     if arg is MISSING:
       ...         assert_type(arg, MISSING)
@@ -1099,7 +1099,7 @@ Sentinel objects
    Example::
 
       >>> class MyClass:
-      ...     MISSING = Sentinel('MyClass.MISSING')
+      ...     MISSING = sentinel('MyClass.MISSING')
 
    .. versionadded:: 4.14.0
 
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -164,6 +164,9 @@ class Sentinel:
 
     *name* should be the name of the variable to which the return value shall be assigned.
 
+    *module_name* is the module where the sentinel is defined.
+    Defaults to the current modules ``__name__``.
+
     *repr*, if supplied, will be used for the repr of the sentinel object.
     If not provided, "<name>" will be used.
     """
@@ -171,10 +174,15 @@ class Sentinel:
     def __init__(
         self,
         name: str,
+        module_name: typing.Optional[str] = None,
+        *,
         repr: typing.Optional[str] = None,
     ):
         self._name = name
         self._repr = repr if repr is not None else f'<{name}>'
+
+        # For pickling as a singleton:
+        self.__module__ = module_name if module_name is not None else _caller()
 
     def __repr__(self):
         return self._repr
@@ -193,11 +201,12 @@ class Sentinel:
         def __ror__(self, other):
             return typing.Union[other, self]
 
-    def __getstate__(self):
-        raise TypeError(f"Cannot pickle {type(self).__name__!r} object")
+    def __reduce__(self) -> str:
+        """Reduce this sentinel to a singleton."""
+        return self._name  # Module is taken from the __module__ attribute
 
 
-_marker = Sentinel("sentinel")
+_marker = Sentinel("sentinel", __name__)
 
 # The functions below are modified copies of typing internal helpers.
 # They are needed by _ProtocolMeta and they provide support for PEP 646.
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -174,8 +174,8 @@ class Sentinel:
     def __init__(
         self,
         name: str,
-        module_name: typing.Optional[str] = None,
         *,
+        module_name: typing.Optional[str] = None,
         repr: typing.Optional[str] = None,
     ):
         self._name = name
@@ -206,7 +206,7 @@ class Sentinel:
         return self._name  # Module is taken from the __module__ attribute
 
 
-_marker = Sentinel("sentinel", __name__)
+_marker = Sentinel("sentinel", module_name=__name__)
 
 # The functions below are modified copies of typing internal helpers.
 # They are needed by _ProtocolMeta and they provide support for PEP 646.
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -176,80 +176,87 @@ def _caller(depth=1, default='__main__'):
 # Placeholder for sentinel methods, because sentinels can not have their own sentinels
 _sentinel_placeholder = object()
 
+if hasattr(builtins, "sentinel"):  # 3.15+
+    sentinel = builtins.sentinel
+else:
+    class sentinel:
+        """Create a unique sentinel object.
 
-class sentinel:
-    """Create a unique sentinel object.
+        *name* should be the name of the variable to which the return value
+        shall be assigned.
+        """
 
-    *name* should be the name of the variable to which the return value shall be assigned.
-    """
+        def __init__(
+            self,
+            __name: str = _sentinel_placeholder,
+            /,
+            repr: typing.Optional[str] = None,
+            *,
+            name: str = _sentinel_placeholder,
+        ) -> None:
+            if name is not _sentinel_placeholder:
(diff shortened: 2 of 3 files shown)


Viicos:
> Replacing `repr` with `module_name` is technically a breaking change.

I think breaking changes are expected for draft PEPs, so imo no deprecation process should be used. Users should be aware that changes are expected. However, I think we can:
- Improve [the documentation about draft PEPs](https://github.com/python/typing_extensions/blob/59d2c20858ac527516ebad5a89c05af514dac94a/doc/index.rst?plain=1#L16-L21), stating that implementation can change.
- Explicitly state that Sentinels are from a draft PEP in the `Sentinel` class documentation, and isn't stable yet.

HexDecimal:
> I'm not willing to remove the `repr` parameter from `typing_extensions.Sentinel` without a deprecation. However, I'm OK to re-export the builtin Sentinel class in 3.15 (if it makes it in), even if it doesn't have this parameter.

Then I will change `repr` into a keyword parameter.

If compatibility is important enough then I can also have the code assume `module_name` is actually `repr` if it does not refer to a existing module. Would this be necessary?

Configuration like `repr` would be defined only with the initial Sentinel object. It wouldn't get stored in the reduce function. This allows changes in `repr` to be reflected in unpickled sentinels. This means that removing `repr` later will not break serialization.

HexDecimal:
I've replaced the `__reduce__` method with a version using pickle's singleton support. This is the most conservative and inoffensive option for a new reduce function since it doesn't add a custom unpickle function, is forward compatible with any future method of pickling, is the standard method of handling singletons via pickle, and is generally strict. This does not support pickling anonymous sentinels which will now raise `pickle.PicklingError` instead of `TypeError`. I can revert this if the previous behavior was more desired, but I personally only need to pickle sentinels which have a top-level definition, and the more I work with them the more that anonymous sentinels feel counter intuitive.

I've attempted to add to the discussion of PEP 661. Anonymous sentinels bring a lot of issues which need to be addressed.
https://discuss.python.org/t/pep-661-sentinel-values/9126/251

Before this is merged I'd like to add a `strict` keyword parameter to Sentinel which prevents sentinels from being accidentally named after existing top-level objects which are not Sentinel as well as ensuring that `repr` is never given conflicting values. I'll need feedback on if this is appropriate.

Viicos:
Thanks @HexDecimal. I just tested your implementation in Pydantic to implement an `UNSET` sentinel, and [pickling seems to work as expected](https://github.com/pydantic/pydantic/pull/11883/files#diff-b446ead7cd1bc63e9765e6faf001b14ea47dc50c2c33ea4fa7965fe548b24d2fR50-R58).

src/typing_extensions.py
```diff
@@ -4155,24 +4156,87 @@ def evaluate_forward_ref(
...
+        return _sentinel_registry.setdefault(registry_key, sentinel)
+
+    @staticmethod
+    def _import_sentinel(name: str, module_name: str):
+        """Return object `name` imported from `module_name`, otherwise return None."""
+        if not module_name:
+            return None
+        try:
+            module = importlib.import_module(module_name)
+            return operator.attrgetter(name)(module)
```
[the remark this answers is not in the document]
HexDecimal: This suggestion breaks qualified names and current tests. `operator.attrgetter` handles dots while `getattr` does not.

HexDecimal:
Note on behavior regarding the sentinel registry and imported sentinels:
```py
class MISSING:
    pass

assert Sentinel("MISSING") is MISSING  # Sentinel named after an existing non-Sentinel object
```
Which of these makes the most sense for the above code?

1. Register and return the existing class as the sentinel identity, the assert will pass (current PR implementation)
2. Return a new Sentinel identity, the assert will fail (old implementation)
3. Raise a warning, then do 1
4. Raise `TypeError` because `MISSING` is not a `Sentinel` (what I recommend, the most strict option)
5. Add a `strict` parameter to `Sentinel` to decide between 1 and 4

---

2nd behavior involving registered sentinel parameters:

```py
MISSING = Sentinel("MISSING", repr="foo")

assert MISSING is Sentinel("MISSING")  # Implicit 'repr' conflicts with previous 'repr'
assert MISSING is Sentinel("MISSING", repr="bar")  # Explicit 'repr' conflicts with previous 'repr'
```
Which of these makes the most sense for the above code?

1. `repr="foo"` is locked in for all subsequent sentinels, subsequent `repr` is always ignored, asserts pass (current implementation)
2. Both asserts raise `ValueError` as long as `repr="foo"` is not given again (my recommendation, the most strict option)
3. `repr` only checked if explicitly given, 1st assert passes, 2nd assert raises `ValueError`
4. Check `repr` like 2 but raise warnings instead of errors
5. Add `strict` parameter to decide between 1 and 2

---

The truthiness of sentinels is still being discussed. One thing which is unclear is if conversions to bool are common enough to be an issue in the first place (because sentinels are normally checked using identity comparison and little else). It'd be interesting to experiment with the most strict option and disable conversions to bool entirely:

```py
def __bool__(self) -> Never:
    raise TypeError(...)
```

HexDecimal:
> Constructing a class shouldn't look around in the globals for other stuff that might be using the same name.

That was necessary to handle unpickling until I finally implemented the correct reduce function, but at this point I can remove that code now and revert to the old behavior. I'd accept that the other options are unnecessarily handholdy.

src/typing_extensions.py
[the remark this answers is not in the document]
HexDecimal: Relevant code removed during refactor.

HexDecimal:
I've narrowed the scope of this PR down to only what's needed to support pickling sentinels. Unfortunately this still means that `repr` needs to be touched to allow for the `module_name` parameter. Thankfully a lot of other changes were no longer needed after this latest rebase.

Has anyone mentioned that `_marker` is a terribly obfuscated name for a sentinel?

src/typing_extensions.py
```diff
@@ -164,18 +164,26 @@ class Sentinel:
...
+    Defaults to the current modules ``__name__``.
+
     *repr*, if supplied, will be used for the repr of the sentinel object.
     If not provided, "<name>" will be used.
     """
 
     def __init__(
         self,
         name: str,
+        module_name: typing.Optional[str] = None,
```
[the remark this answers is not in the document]
HexDecimal: `module_name` is now keyword-only.

JelleZijlstra:
This is mostly good but we shouldn't add a module_name argument since PEP-661 doesn't have that; see #742.

@HexDecimal thanks for the persistent work here! Are you interested in the remaining work of making Sentinel forward-compatible with the accepted version of PEP 661, as outlined in #742?

HexDecimal:
I will work on the PR right now.

HexDecimal:
I've had to move `overload` and `deprecated` in order to perform high-quality deprecations of older sentinel features. I've also had to move `_caller` in order to remove `module_name` from sentinels.

HexDecimal:
I'm having trouble debugging this error: https://github.com/python/typing_extensions/actions/runs/24916592995/job/72969774720?pr=617
Any advice for this? Only affects Python 3.11.15. Deleting doctests affected this this PR does not resolve it.

HexDecimal:
I attempted to deprecate the `Sentinel` class using a `@deprecated` decorator but this broke Python 3.11.15 specifically, not 3.11.0 or 3.12.x or any other version.

HexDecimal:
* Renamed `Sentinel` to `sentinel`, Old name kept, I wanted to deprecate the old name [but was prevented due to this error in Python 3.11.15](https://github.com/python/typing_extensions/actions/runs/24916592995/job/72969774720?pr=617), this specific version does not handle PEP 702 deprecation of classes.
* Deprecated the `repr` parameter and removed it from the documentation. Parameter deprecations are enforced via PEP 702 style deprecations, including enforcing `name` as a positional-only parameter.
* Deprecated subclassing of `sentinel`
* Renamed `_name` attribute to `__name__`
* Strongly discouraged assigning attributes or using `__weakref__` via the documentation. I'm assuming this will switch to `__slots__` later.

Hopefully the code quality here is acceptable. Skipping deprecations would've cleaned this up a lot. Accounting for PEP 702 ahead of time would've prevented code from being shuffled around.

doc/index.rst
```diff
@@ -1071,18 +1071,20 @@ Capsule objects
 Sentinel objects
 ~~~~~~~~~~~~~~~~
 
-.. class:: Sentinel(name, repr=None)
+.. class:: sentinel(name, /)
```
JelleZijlstra, line 1074: We should keep it named as Sentinel, I don't think it's worth trying to rename it. We'll just have the slightly odd case that `typing_extensions.Sentinel == builtins.sentinel`.

doc/index.rst
JelleZijlstra, line 1096: Add a versionchanged 4.16.0 with a summary of the changes in this version.

doc/index.rst
```diff
@@ -1091,6 +1093,13 @@ Sentinel objects
       ...
       >>> func(MISSING)
 
+   Sentinels defined inside a class scope should use a :term:`qualified name`.
```
JelleZijlstra: I'm not sure we need to mention this, it's a fairly unusual use case and typing-extensions docs don't need to be very precise; we can refer to the CPython docs for more.

src/typing_extensions.py
```diff
@@ -159,22 +160,270 @@
...
+class sentinel:
     """Create a unique sentinel object.
 
     *name* should be the name of the variable to which the return value shall be assigned.
-
-    *repr*, if supplied, will be used for the repr of the sentinel object.
-    If not provided, "<name>" will be used.
     """
 
+    @overload
```
JelleZijlstra: I'm not sure the overloads are worth it, since type checkers won't look at the source for this package. It's enough to raise a runtime DeprecationWarning, and we can attempt type checker deprecations in typeshed.

I think we can write the signature something like this:

```python
def __init__(self, _name: str = _marker, /, repr: Optional[str] = None, *, name: str = _marker) -> None:
    if name is not _marker:
       warnings.warn(DeprecationWarning, "Passing `name` as a keyword argument is deprecated")
       _name = name
     if _name is _marker:
         raise TypeError("name is required")
     if repr is not None:
         warnings.warn(DeprecationWarning, ...)
```

And in typeshed we can use overload+deprecated as you suggest.

HexDecimal: I think you're right. I thought PEP 702 would look more clean, and it does, but this is useless outside of typeshed.

I'll revert my changes and switch to this.

doc/index.rst
HexDecimal: Using the correct name is important. Switching from `typing_extensions` to native Python and backporting to older versions should be a simple addition or removal of `from typing_extensions import sentinel`. Using Sentinel with the wrong case needs to be actively discouraged or else things will get worse for code migration.

doc/index.rst
HexDecimal: That makes sense. The reference to PEP 661 should be enough. Anybody else will learn about the qualified name requirement when they try to use pickle.

src/typing_extensions.py
HexDecimal: Note that sentinels can not be used as defaults for the sentinels own methods. This all happens at import-time.

JelleZijlstra: True, good point. We'll have to use `object()`.

doc/index.rst
JelleZijlstra: Fair, I'd like to hear more opinions on this though. I'll post in the issue.

doc/index.rst
```diff
@@ -1095,6 +1094,15 @@ Sentinel objects
 
       See :pep:`661`
 
+   .. versionchanged:: 4.16.0
+
+      Now supports pickle and will be reduced as a singleton.
```
JelleZijlstra, line 1102: Add a line: "The implementation of this class was updated to conform to the accepted version of :pep:`661`."

src/typing_extensions.py
```diff
@@ -159,22 +160,75 @@
...
+    except (AttributeError, ValueError):  # For platforms without _getframe()
+        pass
+    return None
+
+
+# Placeholder for sentinel methods, because sentinels can not have their own sentinels
+_sentinel_placeholder = object()
+
+
+class sentinel:
```
JelleZijlstra: Can you make this re-export builtins.sentinel on 3.15, so something like `if hasattr(builtins, "sentinel"): sentinel = builtins.sentinel`

src/typing_extensions.py
```diff
@@ -159,22 +160,75 @@
...
         repr: typing.Optional[str] = None,
-    ):
-        self._name = name
-        self._repr = repr if repr is not None else f'<{name}>'
+        *,
+        name: str = _sentinel_placeholder,
+    ) -> None:
+        if name is not _sentinel_placeholder:
+            warnings.warn(
+                "'name' is positional-only and must not be a keyword parameter",
```
JelleZijlstra:
```suggestion
                "Passing 'name' as a keyword argument is deprecated; pass it positionally instead.",
```

src/typing_extensions.py
```diff
@@ -159,22 +160,75 @@
...
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            __name = name
+        if __name is _sentinel_placeholder:
+            raise TypeError("First parameter 'name' is required")
+        if repr is not None:
+            warnings.warn(
+                "'repr' is deprecated and must be removed",
```
JelleZijlstra:
```suggestion
                "The 'repr' parameter is deprecated and will be removed in Python 3.15.",
```

src/typing_extensions.py
```diff
@@ -159,22 +160,75 @@
...
+
+        self.__name__ = __name
+        self._repr = repr if repr is not None else __name
+
+        # For pickling as a singleton:
+        self.__module__ = _caller()
+
+    def __init_subclass__(cls):
+        warnings.warn(
+            "Subclassing sentinel is forbidden by PEP 661",
```
JelleZijlstra:
```suggestion
            "Subclassing sentinel is deprecated and will be disallowed in Python 3.15",
```

src/typing_extensions.py
```diff
@@ -159,22 +160,75 @@
...
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        super().__init_subclass__()
+
+    def __setattr__(self, attr: str, value: object) -> None:
+        if attr not in {"__name__", "_repr", "__module__"}:
+            warnings.warn(
+                "Setting attributes on sentinel is deprecated",
```
JelleZijlstra:
```suggestion
                f"Setting attribute {attr!r} on sentinel objects is deprecated and will be disallowed in Python 3.15.",
```

doc/index.rst
```diff
@@ -1071,18 +1071,17 @@ Capsule objects
...
 
-.. class:: Sentinel(name, repr=None)
+.. class:: sentinel(name, /)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
 
-   If *repr* is provided, it will be used for the :meth:`~object.__repr__`
-   of the sentinel object. If not provided, ``"<name>"`` will be used.
+   Assigning attributes to a sentinel including `__weakref__` is forbidden.
```
JelleZijlstra:
```suggestion
   Assigning attributes to a sentinel is deprecated.
```

doc/index.rst
```diff
@@ -1095,6 +1094,15 @@ Sentinel objects
...
       See :pep:`661`
 
+   .. versionchanged:: 4.16.0
+
+      Now supports pickle and will be reduced as a singleton.
+      Renamed from `Sentinel` to `sentinel`, `Sentinel` is deprecated.
+      Automatic `repr` string no longer has angle brackets.
+      `repr` parameter was deprecated.
+      `name` as a keyword is deprecated.
+      Subclasssing and attribute assignment are deprecated.
```
JelleZijlstra:
```suggestion
      Subclassing and attribute assignment are deprecated.
```

src/typing_extensions.py
HexDecimal: Thanks. I was wondering when I was supposed to add this.

`hasattr(builtins, "sentinel")` is correct. I can't imagine any other method being reliable.

doc/index.rst
HexDecimal: Added along with your other suggestions.

JelleZijlstra (approved):

Issue #520, opened 2024-12-13
Add test coverage monitoring


JelleZijlstra:
Getting typing-extensions to 100% test coverage seems like a realistic and useful goal. Doing this ensures that we have thorough tests and also gives more confidence when adding new code.

Contributions towards this goal are welcome. We should start with running a coverage tool to get a sense of how much coverage we're currently missing, then add tests to cover the missing bits, then set up CI so that it enforces 100% coverage. A complication might be that lots of code is version-specific: coverage computation should ideally sum across all supported versions of Python.

AlexWaygood:
The covdefaults coverage plugin might be useful for us. It allows you to suppress coverage calculations on a version branch only if you're running a specific Python version: https://github.com/asottile/covdefaults?tab=readme-ov-file#version-specific--pragma-no-cover

PR #623, opened 2025-07-07
Add Coverage workflow

Daraan:
Added a coverage workflow (resolves: #520).

- [ ] Add PR comment if wanted? - but needs permission, that I can likely not set up (?)
 - https://github.com/marketplace/actions/code-coverage-summary
 - [ ] (Future) Extend PR comment to be more meaningful currently reports only "src" and no specific lines.

Currently the logic is to upload one coverage file from each version test.

### `pip install coverage` & `coverage -m unittest`

A crucial point about this workflow is that it installs and runs the test with `coverage -m unittest` and not `python -m unittest` anymore. `coverage` does (currently) not install `typing_extensions` so I think that is a safe `pip install` (at the moment).  I tried to add a weak test that assures that `typing_extensions.__file__` is indeed the one in `src/` and not `site-packages`, but that test fails on the pypi installation

---

Do you prefer any changes here? e.g. run first with `python` when install and run again with coverage?

Does anyone know why the pypy 3.9 and 3.10 tests fail here? Some kind of code leakage, missing monkeypatch?

EDIT:
- Changed to not run coverage on the pypi versions


AlexWaygood:
if it's a pain to run PyPy tests under coverage, I think it would be fine to do the pypy tests as a separate CI job that aren't run under coverage. None of our code is currently PyPy-specific (we had a workaround for a PyPy bug for a while but it was a tiny branch of code).

.github/workflows/ci.yml
```diff
@@ -64,14 +64,41 @@ jobs:
...
+      - name: Test typing_extensions with coverage
+        if: ${{ !startsWith(matrix.python-version, 'pypy') }}
+        run: |
+          # Be wary of running `pip install` here, since it becomes easy for us to
+          # accidentally pick up typing_extensions as installed by a dependency
+          cd src
+          python --version  # just to make sure we're running the right one
+          export COVERAGE_FILE=.coverage_${{ matrix.python-version }}
+          # Run with coverage omit files that are executed in tempfiles
+          coverage run --omit ann_module*,inspect* -m unittest test_typing_extensions.py
```
AlexWaygood, line 82: nit: I'd do this, otherwise the `python --version` command we ran immediately above doesn't have much value :-)

```suggestion
          python -m coverage run --omit ann_module*,inspect* -m unittest test_typing_extensions.py
```

.github/workflows/ci.yml
```diff
@@ -64,14 +64,41 @@ jobs:
...
+
+      - name: Test typing_extensions with coverage
+        if: ${{ !startsWith(matrix.python-version, 'pypy') }}
+        run: |
+          # Be wary of running `pip install` here, since it becomes easy for us to
+          # accidentally pick up typing_extensions as installed by a dependency
+          cd src
+          python --version  # just to make sure we're running the right one
+          export COVERAGE_FILE=.coverage_${{ matrix.python-version }}
+          # Run with coverage omit files that are executed in tempfiles
```
AlexWaygood, line 81:
```suggestion
          # Run tests under coverage, omitting "data files" in this repo
          # and ephemeral files created in temporary directories in tests
```

.github/workflows/ci.yml
```diff
@@ -109,3 +136,61 @@ jobs:
...
+  report-coverage:
+    name: Report coverage
+
+    runs-on: ubuntu-latest
+
+    needs: [tests]
+    permissions:
+      contents: read
+      pull-requests: write
+
```
AlexWaygood, line 149:
```suggestion
```

src/test_typing_extensions.py
```diff
@@ -6887,6 +6887,7 @@ def test_typing_extensions_compiles_with_opt(self):
             self.fail('Module does not compile with optimize=2 (-OO flag).')
 
 
+
```
AlexWaygood, line 6890:
```suggestion
```

AlexWaygood (commented):
Nice, thank you!

AlexWaygood:
Could you fix the pre-commit failures?

AlexWaygood (approved):
LGTM if we can get the PR comment thing working! (I don't think I have the necessary permissions for that either)

Daraan:
According to https://docs.github.com/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication, a PR from a fork can never have write access (expect the repo itself).

A relevant bit:

> Finally, if the workflow was triggered by a pull request from a forked repository, and the Send write tokens to workflows from pull requests setting is not selected, the permissions are adjusted to change any write permissions to read only.

---

Maintainers note from the action on the error: https://github.com/marocchino/sticky-pull-request-comment/tree/v2/#error-resource-not-accessible-by-integration

> check your Settings > Actions > General > Workflow permissions, and make sure to enable read and write permissions.

The alternative is to provide a `GITHUB_TOKEN` (defaults to ${{ github.token }}) currently.

.github/workflows/coverage_report.yml
```diff
@@ -0,0 +1,62 @@
...
+
+jobs:
+  cov-report:
+    permissions:
+        contents: read
+        pull-requests: write
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request'
+      && github.repository == github.event.workflow_run.repository.full_name
```
woodruffw, line 19: Unfortunately I don't think this works -- `workflow_run` always runs in the context of the "parent" repository, so the second hand of this `&&` expression always evaluate to `true` (I think).

.github/workflows/coverage_report.yml
```diff
@@ -0,0 +1,62 @@
...
+    # The github.repository == ... check will prevent this workflow from running on forks.
+    steps:
+      - name: Get PR number
+        id: get_pr_number
+        env:
+          SHA: "${{ github.event.workflow_run.head_sha }}"
```
woodruffw, line 25: Do you have any docs for this field? I couldn't find anything that clearly documents that it would point to the `HEAD` to the PR's branch, rather than the `HEAD` of the main branch.

(It's also not clear to me how this works when PRs contain some shared history or have identical commits due to cherry picking, so this feels pretty risky.)

.github/workflows/coverage_report.yml
woodruffw: (But also, if the intention is to limit this to first-party PRs, then I think this `workflow_run` can be removed entirely and you can move these steps with their first-party check to `ci.yml` instead 🙂)

JelleZijlstra: For what it's worth, if this works only for first-party PRs that would be a dealbreaker to me; we get lots of third-party PRs and even maintainer PRs often come from forks.

PR #627, opened 2025-07-07
Add sphinx-lint as a pre-commit hook

AlexWaygood:
The tool checks for correctness issues in `.rst` files, and our documentation uses reStructured text. It doesn't currently find any issues in our docs.


srittau (approved):
I also noticed that currently we don't build the docs as part of the CI process. We should probably do that as well to spot any potential problems.

AlexWaygood:
> I also noticed that currently we don't build the docs as part of the CI process. We should probably do that as well to spot any potential problems.

and docs previews as well would be wonderful!

PR #629, opened 2025-07-08
Upgrade ruff version and rules

Daraan:
In relation to the pre-commit PR #628. A while ago I started a list of ruff rules that need to be ignored in the future. With the latest version of ruff (0.12) - if we add it - some of the preview rules became active in the standard set (e.g., UP045, RUF041) but shouldn't be applied to typing_extensions.
Some of the rules I've added in this PR are still in preview, I think it makes sense to add them pro-actively.

Possibly we want RUF023 (sorted __slots__) only on a per line ignore.

```
    "RUF012",  # Use ClassVar for mutables (active now)
    "RUF022",  # Unsorted __all__  (active now)
    "RUF023",  # Unsorted __slots__ (active now)
    "RUF031",  # parentheses for tuples in subscripts (preview)
    "RUF041",  # nested Literal (active now)
    "RUF036",  # None not at end of Union (preview)
    "B903",  # Use dataclass / namedtuple  (preview)
```


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.6
+    rev: v0.12.2
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
```
AlexWaygood, line 6: let's also make it so that pre-commit.ci doesn't push Ruff autofixes to our branches, as per https://github.com/python/typing_extensions/pull/628#issuecomment-3046623599

```suggestion
```

AlexWaygood (commented):
Thank you!!

src/_typed_dict_test_helper.py
```diff
@@ -16,7 +16,7 @@ class Foo(TypedDict):
 
 
 class FooGeneric(TypedDict, Generic[T]):
-    a: Optional[T]
+    a: Optional[T]  # noqa: UP045 while < (3, 10)
```
AlexWaygood, line 19: UP045 was previously part of UP007, but was split into its own rule in a recent release. We have UP007 ignored for the whole repo; I think it probably makes sense to have UP045 ignored for the whole repo as well TBH

AlexWaygood (approved):
great, thank you!

PR #633, opened 2025-07-19
Update cattrs tests

Tinche:
Updates the CI configuration for cattrs to use uv.

### DIFF
--- a/.github/workflows/third_party.yml
+++ b/.github/workflows/third_party.yml
@@ -291,19 +291,16 @@ jobs:
         with:
           path: typing-extensions-latest
           persist-credentials: false
-      - name: Install pdm for cattrs
-        run: pip install pdm
-      - name: Add latest typing-extensions as a dependency
-        run: |
-          cd cattrs
-          pdm remove typing-extensions
-          pdm add --dev ../typing-extensions-latest
-      - name: Install cattrs test dependencies
-        run: cd cattrs; pdm install --dev -G :all
-      - name: List all installed dependencies
-        run: cd cattrs; pdm list -vv
-      - name: Run cattrs tests
-        run: cd cattrs; pdm run pytest tests
+      - name: Install uv
+        run: curl -LsSf https://astral.sh/uv/install.sh | sh
+      - name: Add local version of typing_extensions as a dependency
+        run: cd cattrs; uv add --editable ../typing-extensions-latest
+      - name: Install test dependencies
+        run: cd cattrs; uv sync --group test --all-extras
+      - name: List installed dependencies
+        run: cd cattrs; uv pip list
+      - name: Run tests
+        run: cd cattrs; uv run pytest tests
 
   sqlalchemy:
     name: sqlalchemy tests


Tinche:
I see actions/setup-python@v5 is being used here. In my own CI I've just switched to letting uv handle the Python installation for me.

JelleZijlstra:
Thanks! I'd be open to switching to setup-uv here.

AlexWaygood:
Yeah, I think when we first started using uv in parts of this workflow, uv wasn't yet able to install Python for you and setup-uv didn't yet exist (just the curl command). I'm also open to modernising it a bit!

Tinche:
I switched everything to `astral-sh/setup-uv`, let me know how it looks!

.github/workflows/third_party.yml
```diff
@@ -100,15 +95,15 @@ jobs:
...
-        run: uv pip install --system "typing-extensions @ ./typing-extensions-latest"
+        run: cd typing_inspect; uv pip install "typing-extensions @ ../typing-extensions-latest"
       - name: List all installed dependencies
-        run: uv pip freeze
+        run: cd typing_inspect; uv pip freeze
       - name: Run typing_inspect tests
         run: |
-          cd typing_inspect
-          pytest
+          cd typing_inspect; uv run pytest
```
AlexWaygood, line 106: one (possibly unwarranted) concern I have here is that uv will implicitly do another sync here to ensure the venv is up-to-date with the pyproject.toml file before it executes the `pytest` command. I don't know why it _would_, but it feels like there's a chance that uv could decide that the `main` branch install of typing_extensions doesn't satisfy the declared dependencies in the pyproject.toml file, leading it to uninstall the `main` branch of typing_extensions that we installed just a few lines above and replace it with a version of typing_extensions from PyPI.

For this specific workflow, I'd feel more comfortable sticking to the low-level `uv pip` commands here for the projects that aren't using uv as a project management tool (avoiding `uv run`). I know exactly what `uv pip` doing under the hood 😆

If we kept the venv at the workflow root rather than in the `typing_inspect` directory, I think we could fix the issues you were seeing previously by just doing an old-school activation of the virtual environment here, avoiding `uv run`:

```suggestion
          source .venv/bin/activate
          cd typing_inspect
          pytest
```

(This would be an alternative solution to https://github.com/python/typing_extensions/pull/633/commits/5c8c5f8f18650505eda41faa504a1a0315388c75 -- you'd need to revert that commit)

The same comment applies to the pycroscope, mypy and typeguard jobs

Tinche: This may not be relevant to the problem at hand, but I personally try to avoid explicit venv activation when possible. Apart from being less stateful, I'm mentally adjusting to treating virtualenvs like cattle where different project commands may change or recreate them at will. For example, in cattrs if you do `just --set python 3.9 test`, if the current virtualenv isn't 3.9 it will get nuked and replaced with a 3.9 one. This kind of workflow is unorthodox but I'm really growing to it.

Another solution here would be to just run pytest directly with `../.venv/bin/pytest` instead of using uv, but I think `uv run --no-project` looks cleaner ;)

AlexWaygood: Eh, this workflow is already pretty darn stateful with all the `cd`-ing around -- IMO one tiny venv activation would be preferable to so many `cd`s 😆 but I don't have a strong opinion; this seems fine to me now!

PR #634, opened 2025-07-22
Add @solid_base (PEP 800)

JelleZijlstra:


src/typing_extensions.py
```diff
@@ -315,6 +316,32 @@ class Other(Leaf):  # Error reported by type checker
...
+            @solid_base
+            class Solid2: pass
+
+            class Solid3(Solid1, Solid2): pass  # Type checker error
+
+        Type checkers can use solid bases to detect unreachable code
+        and determine when two types can overlap.
+
+        See PEP 800."""
+        cls.__solid_base__ = True
```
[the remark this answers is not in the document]
AlexWaygood: yeah, it just seems sort-of odd that most pre-existing solid bases without `__slots__` would not allow you to set the attribute, and it also seems more likely than usual that you might try to make a pure-Python class might be immutable if you're adding the `solid_base` decorator to it

src/typing_extensions.py
```diff
@@ -315,6 +316,33 @@ class Other(Leaf):  # Error reported by type checker
         return f
 
 
+if hasattr(typing, "solid_base"):
```
Viicos, line 319:
```suggestion
if hasattr(typing, "solid_base"):  # 3.15
```

To more easily remove the conditional when dropping support for Python versions.

JelleZijlstra: That's optimistic :)

AlexWaygood (approved):
I still don't see any harm in being more cautious about trying to set the attribute, and I think it has some benefits, so I'd still prefer protecting it in a `try`/`except` block. But I'm also happy with this being landed as-is if you disagree :-)

Issue #619, opened 2025-06-05
Self issue with PySide (fix proposal)


noamgot:
Hello
While the issue I describe is a very weird phenomenon that's not directly related to typing_extensions, I would like to propose a solution that might be able to solve this.

**Note:** this is an issue related to python <=3.10

### Minimal reproducable example
```python
import typing

print(hasattr(typing, "Self"))

import PySide6 # noqa

print(hasattr(typing, "Self"))
```
This prints:
```
False
True
```
In other words, `PySide6` manipulates python's built-in typing module. 

### Observation
Examining `typing_extension.py`, I found this section
https://github.com/python/typing_extensions/blob/59d2c20858ac527516ebad5a89c05af514dac94a/src/typing_extensions.py#L2277-L2281

### Proposal
Would it make sense to check `sys.version_info` instead of checking the existance of `Self` attribute?
This is already done in several places, e.g.:
https://github.com/python/typing_extensions/blob/59d2c20858ac527516ebad5a89c05af514dac94a/src/typing_extensions.py#L192-L196

IMHO it looks like a more robust solution (that can solve this specific problem)

I would love to hear your thoughts. I'd be happy to open a PR if needed.

Thanks!

noamgot:
For completeness and future readers, I opened a ticket for the Qt team, follow here: https://bugreports.qt.io/browse/PYSIDE-3135

PR #640, opened 2025-08-03
check-python-version-instead-of-hasattr

alon-ba:
why it is important - because other packages could for some reason monkey patch this variable:

https://github.com/pyqtgraph/pyqtgraph/issues/3352


brianschubert:
Previous discussion in #619

If you're using PySide6 for qt bindings, then the monkey patching issue should be fixed in v6.9.2: https://bugreports.qt.io/browse/PYSIDE-3135

PR #644, opened 2025-08-12
DOC: add missing `versionadded` metadata for `TypeAlias`

neutrinoceros:
According to git blame, `TypeAlias` was added in https://github.com/python/typing_extensions/commit/d362ccb6504bf52618796822e8e6ffa9091575a4, which GitHub's UI indicates was included as early as 3.10.0.0

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -363,6 +363,8 @@ Special typing primitives
 
    See :py:data:`typing.TypeAlias` and :pep:`613`. In ``typing`` since 3.10.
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
We didn't track version changes before 4.0 very well. There are many other objects added before 4.0 that also don't have a `versionadded` note, so I'd rather not add it for just one of them.

4.0 is almost four years old at this point so I don't see much point in tracking down exactly when in the 3.* series each object was added.

PR #651, opened 2025-08-22
Delete many tests

AlexWaygood:
(testing to see if I get a comment from the new coverage workflow when I make a PR from a branch on the upstream repo)


Daraan:
> Oh interesting -- it looks like the bot previously posted a comment after the first commit, that it's now deleted?

Yes this is controlled by the `recreate: true` option.

PR #652, opened 2025-08-22
Delete many tests

AlexWaygood:
The same as #651, but this time I'm making the PR from a branch on my fork. For security reasons, this should *not* trigger a comment from the bot.


AlexWaygood:
Nice: the coverage report is posted in the logs here, but there is no comment posted, which is what we wanted: https://github.com/python/typing_extensions/actions/runs/17158064310/job/48679934651?pr=652

Daraan:
You could try further to get coverage <90% and test if the tests really fail :)

AlexWaygood:
Down to 89%, and now the CI job correctly fails 🎉

But this makes me think that we should raise the threshold below which the job fails. It seems like we can delete _many_ tests before the job starts failing, which doesn't seem great.

Daraan:
> Down to 89%, and now the CI job correctly fails 🎉
> 
> But this makes me think that we should raise the threshold below which the job fails. It seems like we can delete _many_ tests before the job starts failing, which doesn't seem great.

Thanks for testing.

Mixing: 
~4000 lines in `typing_extensions` (priority)
~9000 lines in `test_typing_extensions` of which nearly all are executed skews the percentage into the positive. I would now propose to exclude the test file from the calculation.

AlexWaygood:
> I would now propose to exclude the test file from the calculation.

I don't think we should do that -- it's way too easy to get into a situation where you accidentally don't run tests at all in Python, and including the test file in the consideration guards against that. See https://nedbatchelder.com/blog/201908/dont_omit_tests_from_coverage.html

Daraan:
With ~3s more to the workflow we could also add an html report, which is also pretty nice for exploring via GUI (needs to be manually downloaded):
grafik

https://github.com/Daraan/typing_extensions/actions/runs/17159151595/job/48683686044

zipped html report from above's run
https://github.com/Daraan/typing_extensions/actions/runs/17159151595/artifacts/3829560437

---

As a follow up the report could be piped into GitHub Pages without much effort (I think). Leaving this here if someone wants to check it out:
- https://dev.to/mbarzeev/auto-publish-your-test-coverage-report-on-github-pages-35be
- https://readmedium.com/publishing-coverage-report-html-report-on-github-pages-f5d29de47cfb

I think it would be sufficient to extend the current ci workflow on a `branches: main` `push` event to upload the html artifact to Pages.

PR #654, opened 2025-08-22
Coverage: Implement fail_under

Daraan:
Set the fail threshold to 96% and implemented it by `fail_under` in pyproject.toml.

The fail will now be raised by `coverage report` and not the action. I set up the action to always run even if coverage report fails (still need to test first if it works as intended).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -216,7 +216,7 @@ jobs:
         # -> need another workflow with access to secret token
         if: >-
           ${{
-            always() 
+            always()
             && github.event_name == 'pull_request'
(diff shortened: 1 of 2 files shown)


Daraan:
Works like expected.

I just realized that coverage `fail_under` and the actions `fail_below_min` were not in sync. E.g a 96% coverage was ok for `coverage` but the action failed.
See run: https://github.com/Daraan/typing_extensions/actions/runs/17160726679/job/48689088817

Therefore I adjusted the threshold by -1 to 95%.
It would also be an option to not set `fail_below_min` but that then would not post the *minimum allowed* message.

PR #657, opened 2025-08-24
Use PEP 661 `Sentinel` for internal sentinel

brianschubert:
:dog: 
c.f. https://github.com/python/typing_extensions/pull/656#discussion_r2296730805


brianschubert:
Huh, incidentally the `typing` analog of this sentinel seems to be dead code now: https://github.com/python/cpython/blob/96b7a2eba423b42320f15fd4974740e3e930bb8b/Lib/typing.py#L456-L459

It's last use (and it's global instance) were removed in https://github.com/python/cpython/pull/136332

brianschubert:
Opened https://github.com/python/cpython/pull/138120

JelleZijlstra (approved):

hauntsaninja:
Hmm, would we revert if PEP 661 was rejected?

JelleZijlstra:
If so, we'd likely keep Sentinel around in typing-extensions indefinitely for compatibility, and there's no real compatibility problem with changing this internal sentinel back to an internal class.

PR #667, opened 2025-09-01
Fix how type parameters are collected from parametrized `Protocol` bases

### LINKED ISSUE #636: TypeError when omitting a Protocol type argument with default
This is pretty much the same issue as https://github.com/python/cpython/issues/137191, but with a slightly different error message:

```py
from typing_extensions import Generic, Protocol, TypeVar

T1 = TypeVar("T1")
T2 = TypeVar("T2", default=object)

class A(Protocol[T1]): ...
class B1(A[T2], Protocol, Generic[T1, T2]): ...  # the workaround
class B2(A[T2], Protocol[T1, T2]): ...  # the problem

B1[str]  # ok
B2[str]  # TypeError
```

on `3.12.11`:

```pytb
  File "/home/joren/huh.py", line 11, in <module>
    B2[str]  # TypeError
    ~~^^^^^
  File "/home/joren/.pyenv/versions/3.12.11/lib/python3.12/typing.py", line 398, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/joren/.pyenv/versions/3.12.11/lib/python3.12/typing.py", line 1110, in _generic_class_getitem
  File "/home/joren/.pyenv/versions/3.12.11/lib/python3.12/site-packages/typing_extensions.py", line 3085, in _check_generic
    raise TypeError(f"Too {'many' if alen > elen else 'few'} arguments"
TypeError: Too few arguments for <class '__main__.B2'>; actual 1, expected at least 1
```

on `3.9.23`:

```pytb
  File "/home/joren/huh.py", line 11, in <module>
    B2[str]  # TypeError
  File "/home/joren/.pyenv/versions/3.9.23/lib/python3.9/typing.py", line 277, in inner
    return func(*args, **kwds)
  File "/home/joren/.pyenv/versions/3.9.23/lib/python3.9/typing.py", line 1004, in __class_getitem__
  File "/home/joren/.pyenv/versions/3.9.23/lib/python3.9/site-packages/typing_extensions.py", line 3049, in _check_generic
    raise TypeError(f"Too {'many' if alen > elen else 'few'} {things}"
TypeError: Too few parameters for <class '__main__.B2'>; actual 1, expected at least 1
```

And since `typing_extensions` occurs in the traceback, I figured that this is not the same issue as https://github.com/python/cpython/issues/137191.

brianschubert:
Fixes #636, backport of https://github.com/python/cpython/pull/137281

Adding some inline comments about the implementation below.

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -3203,7 +3203,13 @@ def _is_unpacked_typevartuple(x) -> bool:
     )
 
 
-# Python 3.11+ _collect_type_vars was renamed to _collect_parameters
+# - Python 3.11+ _collect_type_vars was renamed to _collect_parameters.
+#   Breakpoint: https://github.com/python/cpython/pull/31143
+# - Python 3.13+ _collect_parameters was renamed to _collect_type_parameters.
+#   Breakpoint: https://github.com/python/cpython/pull/118900
+# - Monkey patch Generic.__init_subclass__ on <3.15 to fix type parameter
+#   collection from Protocol bases with listed parameters.
+#   Breakpoint: https://github.com/python/cpython/pull/137281
 if hasattr(typing, '_collect_type_vars'):
     def _collect_type_vars(types, typevar_types=None):
         """Collect all type variable contained in types in order of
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -3253,21 +3259,82 @@ def _collect_type_vars(types, typevar_types=None):
...
+            error = (Generic in cls.__bases__ and
+                     cls.__name__ != 'Protocol' and
+                     type(cls) not in (_TypedDictMeta, typing._TypedDictMeta))
+        if error:
+            raise TypeError("Cannot inherit from plain Generic")
+        if '__orig_bases__' in cls.__dict__:
+            typevar_types = (TypeVar, typing.TypeVar, ParamSpec)
+            if hasattr(typing, "ParamSpec"):  # Python 3.10+
+                typevar_types += (typing.ParamSpec,)
+            tvars = _collect_type_vars(cls.__orig_bases__, typevar_types)
```
brianschubert, line 3277: note: for compatibility with both 3.9 and 3.10+, c.f. https://github.com/python/cpython/pull/26091

src/typing_extensions.py
```diff
@@ -3253,21 +3259,82 @@ def _collect_type_vars(types, typevar_types=None):
...
     typing._collect_type_vars = _collect_type_vars
-else:
-    def _collect_parameters(args):
+    typing.Generic.__init_subclass__ = classmethod(_generic_init_subclass)
+elif sys.version_info < (3, 15):
+    def _collect_parameters(
+        args,
+        *,
+        enforce_default_ordering=_marker,
+        validate_all=False,
```
brianschubert, line 3314: note: since `typing` only explicitly passes `validate_all` from inside `_generic_init_subclass` (which this monkey patches below), there's no need for `sys._getframe` hacks like what was done for `enforce_default_ordering` in #392.

src/typing_extensions.py
```diff
@@ -3253,21 +3259,82 @@ def _collect_type_vars(types, typevar_types=None):
...
+            >>> _collect_parameters((list[T], Generic[P, T]))
+            (~P, ~T)
         """
         parameters = []
 
         # A required TypeVarLike cannot appear after a TypeVarLike with default
         # if it was a direct call to `Generic[]` or `Protocol[]`
-        enforce_default_ordering = _has_generic_or_protocol_as_origin()
+        if enforce_default_ordering is _marker:
+            enforce_default_ordering = _has_generic_or_protocol_as_origin()
```
brianschubert, line 3336: note: using a placeholder default value for `enforce_default_ordering` allows the same function to work as `_collect_parameters` on Python <3.13 and as `_collect_type_parameters` on Python 3.13+. Another alternative would be to split this into separate <3.13 and 3.13+ implementations in respective version branches

src/typing_extensions.py
```diff
@@ -3253,21 +3259,82 @@ def _collect_type_vars(types, typevar_types=None):
...
-else:
-    def _collect_parameters(args):
+    typing.Generic.__init_subclass__ = classmethod(_generic_init_subclass)
+elif sys.version_info < (3, 15):
+    def _collect_parameters(
+        args,
+        *,
+        enforce_default_ordering=_marker,
+        validate_all=False,
+    ):
```
brianschubert, line 3315: I saw that there have been previous issues where monkey patching `typing` internals can break future versions of `typing` if new parameters are added (c.f. https://github.com/python/cpython/pull/118900). I think there's low risk of that here, since this only monkey patches internal functions on Python <3.15, and it seems unlikely that new parameter would be backported to 3.14 at this point. But if we wanted to help guard against that, I suppose adding an ignored `**kwargs` parameter might be an option?

PR #668, opened 2025-09-01
Ensure `isinstance` is unaffected by `sys.setprofile` on PyPy; run PyPy tests under coverage

### LINKED ISSUE #661: PyPy: `isinstance` returns different result with `sys.setprofile`
I took a look into the errors from #623 when running PyPy tests under coverage. It seems that the ghost of #318 has come back to haunt us :ghost:!

Under PyPy 3.9 and PyPy 3.10, both `_UnpackAlias` and `_ConcatenateGenericAlias` have `__class__` removed from their class dictionaries when a profiling function has been set.

Here's the output of the script from https://github.com/python/typing_extensions/issues/318#issuecomment-2125837556 under PyPy 3.9:
```text
$ uv run --no-project python --version
Python 3.9.15 (21401ebc2df332b6be6e3d364a985e951a72bbbd, Dec 05 2022, 18:37:07)
[PyPy 7.3.10 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)]

$ uv run --no-project ./gh318_repro.py
Without profiling function
Global class:  True <class '__main__.A'> Local class:  True <class '__main__.A'>

With profiling function
Global class:  True <class '__main__.A'> Local class:  False None

Without profiling function again
Global class:  True <class '__main__.A'> Local class:  True <class '__main__.A'>
```

And similarly, the value of `_ConcatenateGenericAlias.__class__` with and without coverage:
```shell
$ cat temp.py 
from typing_extensions import Concatenate, ParamSpec

P = ParamSpec("P")
C = Concatenate[int, P]
print(type(C).__dict__.get("__class__"))

$ uv run --no-project ./temp.py
<class 'typing._GenericAlias'>

$ uv run --no-project coverage run ./temp.py
None
```

brianschubert:
Fixes #661

The same workaround as #407 seems to work on PyPy as well. Avoiding the `super()` calls prevents `__class__` from being removed from `<type>.__dict__` when a profiling function has been set.

Now the PyPy tests can be run under coverage along with the other tests.

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -6211,6 +6211,47 @@ class ConcatenateTests(BaseTestCase):
             self.assertTrue(typing._is_param_expr(concat))
             self.assertTrue(typing._is_param_expr(typing_concat))
 
+    def test_isinstance_results_unaffected_by_presence_of_tracing_function(self):
+        # See https://github.com/python/typing_extensions/issues/661
+
+        code = textwrap.dedent(
+            """\
+            import sys, typing
+
+            def trace_call(*args):
+                return trace_call
+
+            def run():
+                sys.modules.pop("typing_extensions", None)
+                from typing_extensions import Concatenate
+                return isinstance(Concatenate[...], typing._GenericAlias)
+            isinstance_result_1 = run()
+            sys.setprofile(trace_call)
+            isinstance_result_2 = run()
+            sys.stdout.write(f"{isinstance_result_1} {isinstance_result_2}")
+            """
+        )
+
+        # Run this in an isolated process or it pollutes the environment
+        # and makes other tests fail:
+        try:
+            proc = subprocess.run(
+                [sys.executable, "-c", code], check=True, capture_output=True, text=True,
+            )
+        except subprocess.CalledProcessError as exc:
+            print("stdout", exc.stdout, sep="\n")
+            print("stderr", exc.stderr, sep="\n")
+            raise
+
+        # Sanity checks that assert the test is working as expected
+        self.assertIsInstance(proc.stdout, str)
+        result1, result2 = proc.stdout.split(" ")
+        self.assertIn(result1, {"True", "False"})
+        self.assertIn(result2, {"True", "False"})
+
+        # The actual test:
+        self.assertEqual(result1, result2)
+
 class TypeGuardTests(BaseTestCase):
     def test_basics(self):
         TypeGuard[int]  # OK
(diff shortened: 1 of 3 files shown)


src/test_typing_extensions.py
```diff
@@ -6211,6 +6211,47 @@ def test_is_param_expr(self):
...
 
+    def test_isinstance_results_unaffected_by_presence_of_tracing_function(self):
+        # See https://github.com/python/typing_extensions/issues/661
+
+        code = textwrap.dedent(
+            """\
+            import sys, typing
+
+            def trace_call(*args):
+                return trace_call
```
brianschubert, line 6222: Curiously, if the trace function accesses `f_locals`, e.g.
```python
def trace_call(frame, event, arg):
    frame.f_locals
    return trace_call
```
then the test fails on Python 3.12.0 (and only this version) with this odd error:

```python-traceback
 /home/runner/work/typing_extensions/typing_extensions/src/typing_extensions.py:4303: RuntimeWarning: assigning None to unbound local 'name'
  {name: getattr(typing, name) for name in _typing_names if hasattr(typing, name)}
  File "<string>", line 13, in <module>
  File "<string>", line 9, in run
  File "/home/runner/work/typing_extensions/typing_extensions/src/typing_extensions.py", line 4303, in <module>
    {name: getattr(typing, name) for name in _typing_names if hasattr(typing, name)}
           ^^^^^^^^^^^^^^^^^^^^^
TypeError: attribute name must be string, not 'NoneType'
```
https://github.com/brianschubert/typing_extensions/actions/runs/17386017836/job/49352484071?pr=5

Note that it only fails on the 3.12<b>.0</b> tests, not the 3.12 tests (which pass).

The same is true for the existing test in `ParamSpecTests` that these tests were based on.

src/typing_extensions.py
```diff
@@ -2545,7 +2545,7 @@ def __typing_is_unpacked_typevartuple__(self):
         def __getitem__(self, args):
             if self.__typing_is_unpacked_typevartuple__:
                 return args
-            return super().__getitem__(args)
+            return typing._GenericAlias.__getitem__(self, args)
```
AlexWaygood, line 2548:
```suggestion
            # Cannot use `super().__getitem__` here because of the `__class__` assignment
            # in the class body on Python <=3.11
            # (https://github.com/python/typing_extensions/issues/661)
            return typing._GenericAlias.__getitem__(self, args)
```

src/typing_extensions.py
```diff
@@ -1986,7 +1986,7 @@ class _ConcatenateGenericAlias(list):
         __class__ = typing._GenericAlias
 
         def __init__(self, origin, args):
-            super().__init__(args)
+            list.__init__(self, args)
```
AlexWaygood, line 1989:
```suggestion
            # Cannot use `super().__init__` here because of the `__class__` assignment
            # in the class body (https://github.com/python/typing_extensions/issues/661)
            list.__init__(self, args)
```

AlexWaygood (approved):
Thank you! I'm... sort-of surprised these work on CPython on Python 3.9, given that we know that this `__class__` assignment also does funny things to `super()` calls on older versions of Python. Or do these new tests fail on older versions of CPython on `main` too?

AlexWaygood (approved):
Thank you!

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Unreleased
 
+- Fix incorrect behaviour on PyPy 3.9 and PyPy 3.10 that meant that
+  calling `isinstance` with `typing_extensions.Concatenate[...]` or
+  `typing_extensions.Unpack[...]` as the first argument could have a different
+  result in some situations depending on whether or not a profiling function had been
+  set using `sys.setprofile`. Patch by Brian Schubert.
```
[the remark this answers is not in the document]
AlexWaygood: that LGTM!

AlexWaygood (approved):

PR #674, opened 2025-09-09
Add Codecov step to CI

brianschubert:
c.f. https://github.com/python/typing_extensions/issues/520#issuecomment-3267713399

Curious if this will "just work." This is based on the Codecov setup used in the bedevere repo.


AlexWaygood (approved):
LGTM but I'll wait for a second approval from @JelleZijlstra or @srittau before landing

JelleZijlstra (approved):

PR #676, opened 2025-09-11
Expand protocol allowlist to match CPythons

JanEricNitschke:
Discovered by running the CPython 3.12 test_typing.py over typing_extensions.

https://github.com/JanEricNitschke/typing_extensions/actions/runs/17633620675/job/50105726688#step:7:38

Originally failing test: https://github.com/python/cpython/blob/3.12/Lib/test/test_typing.py#L4047

Current situation in 3.14: https://github.com/python/cpython/blob/3.14/Lib/test/test_typing.py#L4373

And the current allow list: https://github.com/python/cpython/blob/main/Lib/typing.py#L1894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Add support for `AsyncIterator`, `io.Reader`, `io.Writer` and `os.PathLike` protocols.
 - Fix incorrect behaviour on Python 3.9 and Python 3.10 that meant that
   calling `isinstance` with `typing_extensions.Concatenate[...]` or
   `typing_extensions.Unpack[...]` as the first argument could have a different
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Add support for `AsyncIterator`, `io.Reader`, `io.Writer` and `os.PathLike` protocols.
```
```suggestion
- Add support for `AsyncIterator`, `io.Reader`, `io.Writer` and `os.PathLike` protocols
  as bases for other protocls.
```

JanEricNitschke:
Anything i can/should still do here?

JelleZijlstra:
Due to repo rules somebody else needs to approve after I made changes. @srittau @AlexWaygood are either of you willing to look at the open PRs and approve a few?

hauntsaninja (approved):

PR #678, opened 2025-09-11
Ensure init called in protocol subclasses

JanEricNitschke:
Discovered through CPython 3.12 test_typing.py: https://github.com/JanEricNitschke/typing_extensions/actions/runs/17639167194/job/50121940225#step:7:36

Tests are from here:
- https://github.com/python/cpython/pull/27545
- https://github.com/python/cpython/pull/28121
- https://github.com/python/cpython/pull/28121

Code is from current CPython: 
- https://github.com/python/cpython/blob/main/Lib/typing.py#L1843
- https://github.com/python/cpython/blob/main/Lib/typing.py#L2121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Fix `__init__` in subclasses of protocols.
 - Fix incorrect behaviour on Python 3.9 and Python 3.10 that meant that
   calling `isinstance` with `typing_extensions.Concatenate[...]` or
   `typing_extensions.Unpack[...]` as the first argument could have a different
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Fix `__init__` in subclasses of protocols.
```
```suggestion
- Non-protocol subclasses of `Protocol` ignore now the
  `__init__` method inherited from protocol base classes.
```
Align with the CPython NEWS entry

JelleZijlstra (approved):

JanEricNitschke:
Anything i can/should still do here? Conflict is only in the changelog.

PR #679, opened 2025-09-13
Fix error message when subclassing TypeVarTuple

JanEricNitschke:
Issue discovered by running CPython 3.12 test_typing.py on typing_extensions: https://github.com/JanEricNitschke/typing_extensions/actions/runs/17692936537/job/50288896717#step:7:120

Test from here: https://github.com/python/cpython/blob/main/Lib/test/test_typing.py#L1526

Regex from here (adjusted to also handle typing_extensions): https://github.com/python/cpython/blob/main/Lib/test/test_typing.py#L61

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -2637,7 +2637,9 @@ elif hasattr(typing, "TypeVarTuple"):  # 3.11+
             return tvt
 
         def __init_subclass__(self, *args, **kwds):
-            raise TypeError("Cannot subclass special typing classes")
+            raise TypeError(
+                f"type '{__name__}.TypeVarTuple' is not an acceptable base type"
+            )
 
 else:  # <=3.10
     class TypeVarTuple(_DefaultMixin):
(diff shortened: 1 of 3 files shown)


src/typing_extensions.py
```diff
@@ -2634,10 +2634,18 @@ def _typevartuple_prepare_subst(alias, args):
                 )
 
             tvt.__typing_prepare_subst__ = _typevartuple_prepare_subst
+
+            def __mro_entries__(bases):
+                raise TypeError("Cannot subclass an instance of TypeVarTuple.")
+            tvt.__mro_entries__ =  __mro_entries__
```
JelleZijlstra, line 2640:
```suggestion
            tvt.__mro_entries__ = __mro_entries__
```

JelleZijlstra:
@copilot resolve the merge conflicts in this pull request

JanEricNitschke:
Anything i can/should still do here? Conflict is only in the changelog.

PR #680, opened 2025-09-13
Fix doctest for get_protocol_members

JanEricNitschke:
Dicovered by running Cpython 3.12 test_typing.py on typing_extensions: https://github.com/JanEricNitschke/typing_extensions/actions/runs/17693805118/job/50290793287#step:7:115

"New" doctest from here: https://github.com/python/cpython/blob/805e3368d6d07e58430654d1365283924fdf4143/Lib/typing.py#L3737

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # Unreleased
 
-- Fix doctest of `typing_extensions.get_protocol_members`.
 - Fix incorrect behaviour on Python 3.9 and Python 3.10 that meant that
   calling `isinstance` with `typing_extensions.Concatenate[...]` or
   `typing_extensions.Unpack[...]` as the first argument could have a different


AlexWaygood:
Cpython runs these doctests in CI but I'm not sure we do — might be worth it for us to do the same?

JanEricNitschke:
> Cpython runs these doctests in CI but I'm not sure we do — might be worth it for us to do the same?

Done

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Fix doctest of `typing_extensions.get_protocol_members`.
```
JelleZijlstra, line 3: Not sure this is worth mentioning since it's not really a user-visible change. (Well, the docstring changed, but that doesn't feel that worth mentioning.)

JanEricNitschke: Makes sense, got a bit overeager. Should i revert it?

AlexWaygood: I agree, let's remove the changelog entry!

PR #710, opened 2025-12-22
Add tests for LiteralString compatibility with str and bytes

u-harshitha007:
This PR adds runtime tests covering basic `LiteralString` behavior:

- Verifies that `LiteralString` is accepted where `str` is expected
- Verifies that `LiteralString` is not compatible with `bytes`

These tests align with the intended semantics of `LiteralString`
and improve coverage in `test_typing_extensions.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,4 +41,3 @@ See [CONTRIBUTING.md](https://github.com/python/typing_extensions/blob/main/CONT
 for how to contribute to `typing_extensions`.
 
 <!-- Contribution consistency update -->
-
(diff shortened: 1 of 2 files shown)


u-harshitha007:
Thanks for pointing that out — that makes sense.

You're right that these tests don't assert any runtime behavior provided
by typing_extensions itself. I'll remove these tests and look for cases
where runtime behavior is actually defined (or move this kind of example
to documentation instead).

Appreciate the clarification.

PR #715, opened 2026-01-04
Implement sentinels as types, support isinstance and match statments

### LINKED ISSUE #720: Missing pickle support for Sentinel

HexDecimal:
Two main things here:
- The Sentinel class now uses the builtin `type` function to dynamically generate new sentinel types.
- The new meta class `_SentinelMeta` is setup so that sentinels can be used with themselves in `isinstance` checks.

This new type logic ended up being simpler than the workarounds required to make sentinels pretend they are types so I ended up removing those workarounds. I'm not fully experienced with their usage but they were covered by tests.

This PR is based on #617 so `repr` being made a keyword parameter is included. With sentinels being types now it's more important to get `__module__` correct anyways. Pickle is supported and tested.

`self.__name__` takes the place of `self._name` now that sentinels are types.

Updated documentation for sentinels, in addition to #617 it now explains `isinstance` and `match` statements.

Improved the error message for calling a sentinel to include the name of the sentinel.

`isinstance(MySentinel, typing_extensions.Sentinel)` currently returns `False`, but that can be changed in `__instancecheck__` if necessary.

Relevant documentation:
- [metaclasses](https://docs.python.org/3/reference/datamodel.html#metaclasses)
- [`__instancecheck__`](https://docs.python.org/3/reference/datamodel.html#type.__instancecheck__)
- [`type` function](https://docs.python.org/3/library/functions.html#type)

Closes #720

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1056,10 +1056,13 @@ Capsule objects
 Sentinel objects
 ~~~~~~~~~~~~~~~~
 
-.. class:: Sentinel(name, repr=None)
+.. class:: Sentinel(name, module_name=None, *, repr=None)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
+
+   *module_name* is the module where the sentinel is defined.
+   Defaults to the current modules ``__name__``.
 
    If *repr* is provided, it will be used for the :meth:`~object.__repr__`
    of the sentinel object. If not provided, ``"<name>"`` will be used.
@@ -1075,6 +1078,13 @@ Sentinel objects
       ...         assert_type(arg, int)
       ...
       >>> func(MISSING)
+
+   Sentinels defined inside a class scope should use a :term:`qualified name`.
+
+   Example::
+
+      >>> class MyClass:
+      ...     MISSING = Sentinel('MyClass.MISSING')
 
    .. versionadded:: 4.14.0
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1067,17 +1067,28 @@ Sentinel objects
    If *repr* is provided, it will be used for the :meth:`~object.__repr__`
    of the sentinel object. If not provided, ``"<name>"`` will be used.
 
+   Sentinels can be tested using :ref:`is`, :func:`isinstance`,
+   or :ref:`match`.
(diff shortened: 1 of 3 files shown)


HexDecimal:
> This should be in PEP 661 first before we guarantee support for additional behaviors in typing-extensions.

I guess I got ahead of myself following the PEP 661 discussions.

The only additional behavior here is `__instancecheck__`. Would the technical refactoring of this PR be acceptable if this feature were removed? Or is there something fundamentally wrong with this implementation? Honestly I'm not too familiar with metaclasses.

Does this apply to #617 too? Because I thought that was following the specification after I removed all the extraneous changes in the latest push, but I never received feedback on that which is why I made a separate PR for this one because I couldn't tell if anything was wrong with #617 or if simply nobody was free to look at it at the time.

JelleZijlstra:
We won't be implementing sentinels as classes; see #742.

Issue #700, opened 2025-11-25
Tests  fails with Python 3.15.0a2


hegjon:
```
FAIL: test_alias_names_still_exist (test_typing_extensions.AllTests.test_alias_names_still_exist)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-typing-extensions-4.15.0-build/typing_extensions-4.15.0/src/test_typing_extensions.py", line 6895, in test_alias_names_still_exist
    self.assertTrue(
    ~~~~~~~~~~~~~~~^
        hasattr(typing_extensions, name),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        f"{name} no longer exists in typing",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
AssertionError: False is not true : no_type_check_decorator no longer exists in typing

======================================================================
FAIL: test_all_names_in___all__ (test_typing_extensions.AllTests.test_all_names_in___all__)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-typing-extensions-4.15.0-build/typing_extensions-4.15.0/src/test_typing_extensions.py", line 6854, in test_all_names_in___all__
    self.assertEqual({*exclude, *typing_extensions.__all__},
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                     actual_names)
                     ^^^^^^^^^^^^^
AssertionError: Items in the first set but not the second:
'no_type_check_decorator'

======================================================================
FAIL: test_typing_extensions_includes_standard (test_typing_extensions.AllTests.test_typing_extensions_includes_standard)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-typing-extensions-4.15.0-build/typing_extensions-4.15.0/src/test_typing_extensions.py", line 6836, in test_typing_extensions_includes_standard
    self.assertTrue(hasattr(typing_extensions, name))
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: False is not true
```

More information: https://bugzilla.redhat.com/show_bug.cgi?id=2416981

Daraan:
[...]

PR #723, opened 2026-01-14
Remove no_type_check_decorator from _typing_names

Daraan:
Resolves: #700

Tested with 3.15.0a4

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -4302,10 +4302,12 @@ _typing_names = [
     # and some users rely on it.
     "_AnnotatedAlias",
 ]
+
 # Breakpoint: https://github.com/python/cpython/pull/133602
 if sys.version_info < (3, 15, 0):
     _typing_names.append("no_type_check_decorator")
     __all__.append("no_type_check_decorator")
+
 globals().update(
     {name: getattr(typing, name) for name in _typing_names if hasattr(typing, name)}
 )


src/typing_extensions.py
```diff
@@ -4298,11 +4298,14 @@ def type_repr(value):
...
-    "no_type_check_decorator",
     # This is private, but it was defined by typing_extensions for a long time
     # and some users rely on it.
     "_AnnotatedAlias",
 ]
+# Breakpoint: https://github.com/python/cpython/pull/133602
+if sys.version_info < (3, 15, 0):
+    _typing_names.append("no_type_check_decorator")
+    __all__.append("no_type_check_decorator")
 globals().update(
```
AlexWaygood: formatting nit (sorry!)

```suggestion
]

# Breakpoint: https://github.com/python/cpython/pull/133602
if sys.version_info < (3, 15, 0):
    _typing_names.append("no_type_check_decorator")
    __all__.append("no_type_check_decorator")

globals().update(
```

srittau:
Since this was never implemented by any type checkers as far as I know, we could even skip the import for all Python versions and just add something like (pseudo-code):

```python
def __getattr__(name):
    if name == "no_type_check_decorator":
        # print deprecation warning
        return lambda f: f
```

or just

```python
@deprecated("...")
def no_type_check_decorator(f):
    return f
```

JelleZijlstra:
I think this is fine as is. It's OK if behavior changes depending on the Python version.

PR #727, opened 2026-01-28
docs: add security implications of annotations

JelleZijlstra:
I received a report alleging that the use of `eval()` here is a security vulnerability. There is a caution about this already in the stdlib at https://docs.python.org/3.14/library/annotationlib.html#security-implications-of-introspecting-annotations but I suppose we should put it in the typing-extensions docs too.


srittau (approved):
Good to explicitly document this.

PR #732, opened 2026-03-22
Add pickle support for Sentinel via singleton registry

r266-tech:
## Summary

This PR adds pickle support for `typing_extensions.Sentinel` objects, fixing #720.

Sentinel objects now support pickling/unpickling using a module-level registry keyed by `(module_name, name)`. Unpickled sentinels preserve object identity (`is` checks pass).

## Changes

- Add `_sentinel_registry` dict for singleton tracking
- Convert `__init__` to `__new__` with registry-based singleton pattern
- Replace `__getstate__` (which raised `TypeError`) with `__reduce__`
- `__reduce__` returns `(cls, (name, None, module_name))` for reconstruction
- Auto-detect calling module via `inspect.currentframe()`
- Update tests: verify pickle roundtrip, identity preservation, singleton behavior

## Approach

This implements the singleton registry pattern discussed in the issue and consistent with the PEP 661 reference implementation. When a sentinel is created, it is registered in `_sentinel_registry` keyed by `(module_name, name)`. On unpickling, `Sentinel(name, module_name=module_name)` is called, which looks up the existing singleton from the registry rather than creating a new object.

## Motivation

As noted in #720, pickle support is expressed as a motivation in PEP 661. The current explicit rejection (`__getstate__` raises TypeError) blocks legitimate use cases where sentinel values need to be serialized. Singletons are inherently forward-compatible — they simply reference an object by its qualified name and do not care what that object is later replaced with.

## Tests

All 554 existing tests pass (13 skipped for version-specific features). New tests added:
- `test_sentinel_picklable` — verifies pickle roundtrip across all protocols
- `test_sentinel_pickle_preserves_identity` — verifies `is` checks pass after unpickling
- `test_sentinel_singleton` — verifies same-name sentinels return the same object

### DIFF
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -9570,13 +9570,24 @@ class TestSentinels(BaseTestCase):
         ):
             sentinel()
 
-    def test_sentinel_not_picklable(self):
+    def test_sentinel_picklable(self):
         sentinel = Sentinel('sentinel')
-        with self.assertRaisesRegex(
-            TypeError,
-            "Cannot pickle 'Sentinel' object"
-        ):
-            pickle.dumps(sentinel)
+        for proto in range(pickle.HIGHEST_PROTOCOL + 1):
+            pickled = pickle.dumps(sentinel, protocol=proto)
+            loaded = pickle.loads(pickled)
+            self.assertIs(loaded, sentinel)
+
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
This is not the direction we need to go; see #742.

PR #741, opened 2026-04-23
Type variable tuple variance

KotlinIsland:
https://github.com/python/cpython/pull/148212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   Python 3.9. The `typing` implementation has always raised an error, and the
   `typing_extensions` implementation has raised an error on Python 3.10+ since
   `typing_extensions` v4.6.0. Patch by Brian Schubert.
+- Add `bound` and variance parameters to `TypeVarTuple`.
 
 # Release 4.15.0 (August 25, 2025)
 
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -1762,7 +1762,7 @@ class GetTypeHintTests(BaseTestCase):
             annotation              : annotation,
             Optional[int]           : Optional[int],
             Optional[List[str]]     : Optional[List[str]],
-            Optional[annotation]     : Optional[annotation],
+            Optional[annotation]    : Optional[annotation],
             Union[str, None, str]   : Optional[str],
             Unpack[Tuple[int, None]]: Unpack[Tuple[int, None]],
         }
@@ -1780,6 +1780,8 @@ class GetTypeHintTests(BaseTestCase):
             Union[str, "Union[None, StrAlias]"]: Optional[str],
             Union["annotation", T_default]     : Union[annotation, T_default],
             Annotated["annotation", "nested"]  : Annotated[Union[int, None], "data", "nested"],
+            # Note: A starred *Ts will use typing.Unpack in 3.11+ see Issue #485
+            Unpack[Ts]                         : Unpack[Ts],
         }
         # Note: A starred *Ts will use typing.Unpack in 3.11+ see Issue #485
         if TYPING_3_15_0:
@@ -6610,7 +6612,10 @@ class UnpackTests(BaseTestCase):
     @skipIf(TYPING_3_15_0, "repr changed in 3.15")
     def test_repr(self):
         Ts = TypeVarTuple('Ts')
-        self.assertEqual(repr(Unpack[Ts]), f'{Unpack.__module__}.Unpack[Ts]')
+        if not hasattr(typing, 'TypeVarTuple') or sys.version_info >= (3, 15):
+            self.assertEqual(repr(Unpack[Ts]), f'{Unpack.__module__}.Unpack[~Ts]')
+        else:
+            self.assertEqual(repr(Unpack[Ts]), f'{Unpack.__module__}.Unpack[Ts]')
 
     @skipUnless(TYPING_3_15_0, "repr changed in 3.15")
     def test_repr_py315(self):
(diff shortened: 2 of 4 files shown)


src/typing_extensions.py
JelleZijlstra: This needs to change now, we currently re-export typing.TypeVarTuple on 3.13 and 3.14 but with this change we can do so only on 3.15+.

JelleZijlstra (commented):
- Can you also set `tvt.__bound__ = bound` for consistency?
- Can you add tests and a changelog entry?

src/typing_extensions.py
```diff
@@ -2689,8 +2702,16 @@ def get_shape(self) -> Tuple[*Ts]:
         def __iter__(self):
             yield self.__unpacked__
 
-        def __init__(self, name, *, default=NoDefault):
+        def __init__(self, name, *, bound=None, covariant=False, contravariant=False,
+                     infer_variance=False, default=NoDefault):
             self.__name__ = name
+            self.__covariant__ = bool(covariant)
```
JelleZijlstra, line 2784: Why does this branch call bool() on the variance-related arguments and `_type_check` on bound and the above one doesn't? We should have things behave the same way across versions.

src/test_typing_extensions.py
```diff
@@ -6750,7 +6753,43 @@ def test_basic_plain(self):
...
+        Ts_2 = TypeVarTuple('Ts_2')
+        if not hasattr(typing, 'TypeVarTuple') or sys.version_info >= (3, 15):
+            self.assertEqual(repr(Ts), '~Ts')
+            self.assertEqual(repr(Ts_2), '~Ts_2')
+
+            self.assertEqual(repr(Ts_co), '+Ts_co')
+            self.assertEqual(repr(Ts_contra), '-Ts_contra')
+            self.assertEqual(repr(Ts_infer), 'Ts_infer')
+        else:
+            # Not worth creating our own version of TypeVarTuple
```
JelleZijlstra: I don't understand this comment, we do create our own version in order to add the variance arguments, so we might as well update the repr too.

In general the tests ideally shouldn't have version-dependent branches: it's a goal for typing-extensions for all versions to behave the same.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
   Python 3.9. The `typing` implementation has always raised an error, and the
   `typing_extensions` implementation has raised an error on Python 3.10+ since
   `typing_extensions` v4.6.0. Patch by Brian Schubert.
+- add `bound` and variance parameters to `TypeVarTuple`
```
```suggestion
- Add `bound` and variance parameters to `TypeVarTuple`.
```

src/test_typing_extensions.py
KotlinIsland: it's adapted from the `ParamSpec` test which does similar. we do create a class for `TypeVarTuple`, but it doesn't return `Self`, so we can't update the repr

src/typing_extensions.py
KotlinIsland: the implementation of `ParamSpec` also has this discrepancy, i will update it accordingly

should `_set_default` also invoke `_type_check`?

src/test_typing_extensions.py
KotlinIsland: https://github.com/python/typing_extensions/blob/83caa5908b408560b7e30d60052c2e4da31b0556/src/test_typing_extensions.py#L5769

src/typing_extensions.py
JelleZijlstra: Sorry I didn't come back here for a while. I think bool() is right for the boolean fields since it matches what we do for TypeVar. However, _type_check is actually wrong for ParamSpec and TypeVarTuple bounds, since it assumes a single type. This is actually a bug in CPython too, it rejects TypeVarTuple bounds like `(int, str)` that should be valid.

KotlinIsland: > it rejects TypeVarTuple bounds like `(int, str)` that should be valid.

is this in the spec? perhaps this is unrelated to this change

JelleZijlstra: Ah right, we're not actually adding support for bounds here, only for variance. Might as well get this fixed in the runtime though.

JelleZijlstra: https://github.com/python/cpython/issues/151955

JelleZijlstra (approved):

src/test_typing_extensions.py
```diff
@@ -6610,7 +6612,10 @@ def test_basic_plain(self):
     @skipIf(TYPING_3_15_0, "repr changed in 3.15")
     def test_repr(self):
         Ts = TypeVarTuple('Ts')
-        self.assertEqual(repr(Unpack[Ts]), f'{Unpack.__module__}.Unpack[Ts]')
+        if not hasattr(typing, 'TypeVarTuple') or sys.version_info >= (3, 15):
+            self.assertEqual(repr(Unpack[Ts]), f'{Unpack.__module__}.Unpack[~Ts]')
+        else:
+            self.assertEqual(repr(Unpack[Ts]), f'{Unpack.__module__}.Unpack[Ts]')
```
AlexWaygood: this test only runs on Python <3.15 (it's decorated with `@skipIf(TYPING_3_15_0, "repr changed in 3.15")`, so I'm a bit confused about the `if not hasattr(typing, 'TypeVarTuple') or sys.version_info >= (3, 15)` guard here. Testing what happens when `sys.version_info >= (3, 15)` is meant to be the job of the `test_repr_py315` test immediately below this one

src/test_typing_extensions.py
```diff
@@ -6814,7 +6819,44 @@ def test_basic_plain(self):
...
+        else:
+            # On other versions we use typing.TypeVarTuple, but it is not aware of
+            # variance. Not worth creating our own version of TypeVarTuple
+            # for this.
+            self.assertEqual(repr(Ts), 'Ts')
+            self.assertEqual(repr(Ts_2), 'Ts_2')
+
+            self.assertEqual(repr(Ts_co), 'Ts_co')
+            self.assertEqual(repr(Ts_contra), 'Ts_contra')
+            self.assertEqual(repr(Ts_infer), 'Ts_infer')
```
AlexWaygood: similarly here, I'm confused by the `if` guard -- this whole test is skipped on Python 3.15+

AlexWaygood-openai (approved):

AlexWaygood (approved):

JelleZijlstra (approved):

PR #743, opened 2026-04-29
Add module argument to TypeAliasType

ericmarkmartin:
https://github.com/python/cpython/pull/149133


ericmarkmartin:
Closing in favor of #744

PR #755, opened 2026-05-18
link to failing run directly from the issue when the daily test fails

AlexWaygood:
We already do this for the third-party workflow but forgot to apply the same change to this workflow. X-ref https://github.com/python/typing_extensions/issues/751, where the bot unhelpfully just pasted a link to all runs of that workflow rather than specifically the run that caused the failure


AlexWaygood:
(The PyPy/3.11 failure is pre-existing on `main`...)

JelleZijlstra (approved):

PR #757, opened 2026-05-19
Update sentinel

### LINKED ISSUE #742: Adapt Sentinel implementation to accepted version of PEP 661

zedzhen:
https://github.com/python/typing_extensions/issues/742#issuecomment-4489794101
close #742

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1075,12 +1075,12 @@ Capsule objects
 Sentinel objects
 ~~~~~~~~~~~~~~~~
 
-.. class:: sentinel(name, /)
+.. class:: sentinel(name, /, *, repr=None)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
 
-   Assigning attributes to a sentinel is deprecated.
+   Assigning attributes to a sentinel is deprecated (except for __module__).
 
    Example::
 
@@ -1106,7 +1106,7 @@ Sentinel objects
       Now supports pickle and will be reduced as a singleton.
       Renamed from `Sentinel` to `sentinel`, `Sentinel` is deprecated.
       Automatic `repr` string no longer has angle brackets.
-      `repr` parameter was deprecated.
+      `repr` as a positional argument is deprecated.
       `name` as a keyword is deprecated.
       Subclassing and attribute assignment are deprecated.
 
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -9660,6 +9660,8 @@ class TestSentinels(BaseTestCase):
             my_sentinel = Sentinel(name="my_sentinel")
         with self.assertWarnsRegex(DeprecationWarning, r"Setting attribute 'foo' on sentinel objects is deprecated"):
             my_sentinel.foo = "bar"
+        with self.assertWarnsRegex(DeprecationWarning, r"Setting attribute '__name__' on sentinel objects is deprecated"):
+            my_sentinel.__name__ = "bar"
 
 
 def load_tests(loader, tests, pattern):
--- a/src/test_typing_extensions.py
+++ b/src/test_typing_extensions.py
@@ -9600,7 +9600,7 @@ class TestSentinels(BaseTestCase):
         self.assertEqual(repr(sentinel_no_repr), 'sentinel_no_repr')
 
     def test_sentinel_deprecated_argument_repr(self):
-        with self.assertWarnsRegex(DeprecationWarning, r"Passing 'repr' as a position argument is deprecated; pass it by keyword instead."):
+        with self.assertWarnsRegex(DeprecationWarning, r"Passing 'repr' as a positional argument is deprecated; pass it by keyword instead."):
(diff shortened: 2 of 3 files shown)


HexDecimal:
I assume the `repr` keyword should be added to the sentinel docs at `doc/index.rst`, and the `versionchanged` directive should be updated to clarify that only positional `repr` is deprecated.

HexDecimal:
The `typing.Optional[str]` parameter not acknowledging `None` is invalid and also breaks backwards compatibility.

src/typing_extensions.py
```diff
@@ -204,13 +205,16 @@ def __init__(
...
-                    "The 'repr' parameter is deprecated "
-                    "and will be removed in Python 3.15.",
+                    "Passing 'repr' as a position argument is deprecated; "
+                    "pass it by keyword instead.",
                     DeprecationWarning,
                     stacklevel=2,
                 )
+                repr = __repr
+            if repr is _sentinel_placeholder:
+                repr = None
```
HexDecimal: If defaults were `repr=None` and `__repr=_sentinel_placeholder` then this check wouldn't be necessary.

zedzhen: Thank you.

src/typing_extensions.py
```diff
@@ -204,13 +205,14 @@ def __init__(
                 __name = name
             if __name is _sentinel_placeholder:
                 raise TypeError("First parameter 'name' is required")
-            if repr is not None:
+            if __repr is not _sentinel_placeholder:
                 warnings.warn(
-                    "The 'repr' parameter is deprecated "
-                    "and will be removed in Python 3.15.",
+                    "Passing 'repr' as a position argument is deprecated; "
```
JelleZijlstra:
```suggestion
                    "Passing 'repr' as a positional argument is deprecated; "
```

doc/index.rst
```diff
@@ -1075,12 +1075,12 @@ Capsule objects
...
 ~~~~~~~~~~~~~~~~
 
-.. class:: sentinel(name, /)
+.. class:: sentinel(name, /, *, repr=None)
 
    A type used to define sentinel values. The *name* argument should be the
    name of the variable to which the return value shall be assigned.
 
-   Assigning attributes to a sentinel is deprecated.
+   Assigning attributes to a sentinel is deprecated (except for __module__).
```
JelleZijlstra, line 1083: Looks like we actually allow assigning to `__name__`, which CPython doesn't.

Maybe we should store it in a private attribute (`_name`), and make `__name__` a property where the setter raises a DeprecationWarning.

src/test_typing_extensions.py
```diff
@@ -9599,11 +9599,16 @@ def test_sentinel_no_repr(self):
...
-        with self.assertWarnsRegex(DeprecationWarning, r"'repr' parameter is deprecated and will be removed"):
-            sentinel_explicit_repr = sentinel('sentinel_explicit_repr', repr='explicit_repr')
+    def test_sentinel_deprecated_argument_repr(self):
+        with self.assertWarnsRegex(DeprecationWarning, r"Passing 'repr' as a position argument is deprecated; pass it by keyword instead."):
+            sentinel_argument_repr = sentinel('sentinel_argument_repr', 'argument_repr')
 
-        self.assertEqual(repr(sentinel_explicit_repr), 'explicit_repr')
+        self.assertEqual(repr(sentinel_argument_repr), 'argument_repr')
+
+    def test_sentinel_deprecated_keyword_repr(self):
```
HexDecimal:
```suggestion
    def test_sentinel_keyword_repr(self):
```

zedzhen:
@JelleZijlstra, please review this PR.

JelleZijlstra (approved):

PR #758, opened 2026-05-26
Update pre-commit hooks; pin GitHub actions to exact refs

AlexWaygood:
Closes https://github.com/python/typing_extensions/pull/739.

This updates our pre-commit hooks to the latest version, and fixes new zizmor complaints about the security of our GitHub actions by pinning our actions to exact SHAs. This matches GitHub's [recommendation](https://docs.github.com/en/actions/reference/security/secure-use) for secure workflows, and shouldn't result in us failing to get security updates for any GitHub actions: we have dependabot configured for this repo, and it will immediately file an out-of-schedule PR updating a GitHub action if the latest update is marked as fixing a security issue.

Zizmor also complained about our codecov upload step due to https://docs.zizmor.sh/audits/#secrets-outside-env. This was fixed by switching to `use-oidc`, which is a documented alternative: https://github.com/codecov/codecov-action#using-oidc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,13 +55,17 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: read
+      id-token: write  # for codecov upload
+
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
         with:
           persist-credentials: false
 
       - name: Set up Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6
         with:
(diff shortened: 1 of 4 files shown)


JelleZijlstra (approved):

.github/workflows/ci.yml
```diff
@@ -55,13 +55,17 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: read
+      id-token: write  # for codecov upload
+
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
```
AlexWaygood, line 63: codex believes (from looking at dependabot's implementation) that dependabot will update these comments as well when it updates our actions pins. It can't find docs anywhere to confirm that, though it's also what renovate does, so it's what I'd expect

PR #762, opened 2026-06-11
Fail sphinx build on warnings

srittau:
Fix intersphinx links


srittau:
codecov fails due to a missing GPG signature.

AlexWaygood (approved):

PR #776, opened 2026-07-02
Bump version to 4.16.1.dev0

AlexWaygood:


srittau (commented):
Shouldn't the minor version be bumped to ensure correct ordering?

AlexWaygood:
Hmm, not according to https://github.com/python/typing_extensions/blob/main/CONTRIBUTING.md#workflow-for-pypi-releases and https://github.com/python/typing_extensions/blob/main/CONTRIBUTING.md#development-version

AlexWaygood:
Oh no you're right, it says "_next_ planned release", doesn't it

AlexWaygood:
Thanks!

pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "flit_core.buildapi"
 # Project metadata
 [project]
 name = "typing_extensions"
-version = "4.16.0"
+version = "4.16.0.dev"
```
AlexWaygood:
```suggestion
version = "4.17.0.dev"
```

srittau:
Sorry to be a nuisance, but wouldn't it be safer to use `4.16.1.dev`, since we don't know whether we might need a bug-fix release? We could then bump the minor version (and I know I said "minor" version in my last message, I should have said "patch") when we introduce a new feature.

For reference, the [Version specifiers reference](https://packaging.python.org/en/latest/specifications/version-specifiers/#summary-of-permitted-suffixes-and-relative-ordering) specifies the ordering: `1.0.0 < 1.0.1.dev < 1.0.1.a1 < 1.0.1 < 1.1.0.dev < 1.1.0` etc.

AlexWaygood:
Not a nuisance at all! I'm not an expert on packaging shenanigans, and it seems important to get this right...

pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "flit_core.buildapi"
 # Project metadata
 [project]
 name = "typing_extensions"
-version = "4.16.0"
+version = "4.17.0.dev"
```
AlexWaygood:
```suggestion
version = "4.17.1.dev0"
```

AlexWaygood:
I changed it to `4.17.1.dev0`

pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "flit_core.buildapi"
 # Project metadata
 [project]
 name = "typing_extensions"
-version = "4.16.0"
+version = "4.17.1.dev0"
```
srittau:
```suggestion
version = "4.16.1.dev0"
```

AlexWaygood: ugh. thank you.

srittau (approved):

PR #777, opened 2026-07-02
Clarify version policy

### LINKED ISSUE #528: Discussion: Upgrade repo version after a release to `dev` to not match release version

srittau:
We will increase the patch version after a release. The minor version is increased when features are added.

Also, use `dev0` instead of `dev`, since the former is the normalized version of the latter.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,9 +25,11 @@ Starting with version 4.0.0, `typing_extensions` uses
 for more detail.
 
 ## Development version
-After a release the version is increased once in [pyproject.toml](/pyproject.toml) and
-appended with a `.dev` suffix, e.g. `4.0.1.dev`.
-Further subsequent updates are not planned between releases.
+
+After a release the patch version is increased in [pyproject.toml](/pyproject.toml) and
+appended with a `.dev0` suffix, e.g. `4.0.1.dev0`.
+
+When a new feature is added we will increase the minor version once between releases. For example, from 4.0.1.dev0 to 4.1.0.dev0.
 
 # Type stubs
 
@@ -111,4 +113,4 @@ pipx run pre-commit run -a
 - Release automation will finish the release. You'll have to manually
   approve the last step before upload.
 
-- After the release has been published on PyPI upgrade the version in number in [pyproject.toml](/pyproject.toml) to a `dev` version of the next planned release. For example, change 4.1.1 to 4.X.X.dev, see also [Development versions](#development-version). # TODO decide on major vs. minor increase.
+- After the release has been published on PyPI upgrade the version in number in [pyproject.toml](/pyproject.toml) to a `dev` version of the next planned release. For example, change 4.1.1 to 4.1.2.dev0, see also [Development versions](#development-version).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,15 +88,18 @@ pipx run pre-commit run -a
 
 # Workflow for PyPI releases
 
-- Make sure you follow the versioning policy in the documentation
-  (e.g., release candidates before any feature release, do not release development versions)
+- Make sure to release at least one release candidate (starting with `rc1`) before making a full release. E.g. release `4.2.0rc1` before releasing `4.2.0`.
+  When an urgent hotfix is necessary (and you bump only the patch version), you can skip the release candidates.
 
 - Ensure that GitHub Actions reports no errors.
 
 - Check that `CHANGELOG.md` accurately reflects all important changes committed to `main` since the previous release.
 
 - Update the version number in `typing_extensions/pyproject.toml` and in
-  `typing_extensions/CHANGELOG.md`.
+  `typing_extensions/CHANGELOG.md` by removing `.dev0` and potentially adding the `rcX` suffix:
+  - `4.1.0.dev0` → `4.1.0rc1`
+  - `4.1.0rc2.dev0` → `4.1.0rc2` or `4.1.0` when making a full release
+  - `4.1.1.dev0` → `4.1.1rc1` or `4.1.1` when making an urgent hotfix
 
 - Create a new GitHub release at https://github.com/python/typing_extensions/releases/new.
   Details:
@@ -106,4 +109,6 @@ pipx run pre-commit run -a
 - Release automation will finish the release. You'll have to manually
   approve the last step before upload.
 
-- After the release has been published on PyPI upgrade the version in number in [pyproject.toml](/pyproject.toml) to a `dev` version of the next planned release. For example, change 4.1.1 to 4.1.2.dev0.
+- After the release has been published on PyPI upgrade the patch or rc version number in [pyproject.toml](/pyproject.toml) and append `.dev0`. For example:
+    - 4.1.1 → 4.1.2.dev0
+    - 4.1.1rc1 → 4.1.1rc2.dev0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,7 +98,7 @@ pipx run pre-commit run -a
 - Update the version number in `typing_extensions/pyproject.toml` and in
   `typing_extensions/CHANGELOG.md` by removing `.dev0` and potentially adding the `rcX` suffix:
   - `4.1.0.dev0` → `4.1.0rc1` for the initial release candidate
-  - `4.1.0rc2.dev0` → `4.1.0rc2` or `4.1.0` when making a full release
+  - `4.1.0rc2.dev0` → `4.1.0rc2` or `4.1.0`, depending on whether the release is a second release candidate or a full release
   - `4.1.1.dev0` → `4.1.1` when making a hotfix release
 
 - Create a new GitHub release at https://github.com/python/typing_extensions/releases/new.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,13 +24,6 @@ Starting with version 4.0.0, `typing_extensions` uses
 [Semantic Versioning](https://semver.org/). See the documentation
 for more detail.
 
-## Development version
-
-After a release the patch version is increased in [pyproject.toml](/pyproject.toml) and
-appended with a `.dev0` suffix, e.g. `4.0.1.dev0`.
-
-When a new feature is added we will increase the minor version once between releases. For example, from 4.0.1.dev0 to 4.1.0.dev0.
-
 # Type stubs
 
 A stub file for `typing_extensions` is maintained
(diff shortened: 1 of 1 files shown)


srittau:
@AlexWaygood Could you have a look whether this makes sense?

CONTRIBUTING.md
```diff
@@ -25,9 +25,11 @@ Starting with version 4.0.0, `typing_extensions` uses
...
 
 ## Development version
-After a release the version is increased once in [pyproject.toml](/pyproject.toml) and
-appended with a `.dev` suffix, e.g. `4.0.1.dev`.
-Further subsequent updates are not planned between releases.
+
+After a release the patch version is increased in [pyproject.toml](/pyproject.toml) and
+appended with a `.dev0` suffix, e.g. `4.0.1.dev0`.
+
+When a new feature is added we will increase the minor version once between releases. For example, from 4.0.1.dev0 to 4.1.0.dev0.
```
AlexWaygood: To be honest, I'm not sure we need this separate section? When putting out the releases over the last few weeks, I found it confusing that the Release instructions section at the bottom linked to this section, when this section didn't actually give any additional useful information compared to the release instructions.

Does anybody except the release manager really care what our development versioning policy is?

CONTRIBUTING.md
```diff
@@ -111,4 +113,4 @@ pipx run pre-commit run -a
 - Release automation will finish the release. You'll have to manually
   approve the last step before upload.
 
-- After the release has been published on PyPI upgrade the version in number in [pyproject.toml](/pyproject.toml) to a `dev` version of the next planned release. For example, change 4.1.1 to 4.X.X.dev, see also [Development versions](#development-version). # TODO decide on major vs. minor increase.
+- After the release has been published on PyPI upgrade the version in number in [pyproject.toml](/pyproject.toml) to a `dev` version of the next planned release. For example, change 4.1.1 to 4.1.2.dev0, see also [Development versions](#development-version).
```
AlexWaygood: the "next planned release" language here is a bit confusing, no? We never _plan_ to do a patch release; we only make a patch release if there was a bug in the latest minor release. But as your helpful example later on makes clear, we should really bump the version from `4.1.0 -> 4.1.1.dev0` rather than `4.1.0 -> 4.2.0.dev0`.

Also, what do we do for `rc` releases? After `4.17.0rc1` is released, does the version bump to `4.17.0.dev`? What if we then make an `rc2` release the next day? That might lead us to go from `4.17.0rc1 -> 4.17.0.dev0 -> 4.17.0rc2 -> 4.17.0.dev0 -> 4.17.0`... so I guess after `rc1` is released, we should actually bump the version to `4.17.0rc2.dev0`, even if we don't plan for an `rc2` release? Then the sequence would go `4.17.0rc1 -> 4.17.0rc2.dev0 -> 4.17.0rc2 -> 4.17.0rc3.dev0 -> 4.17.0 -> 4.17.0.dev0`.

It would be really helpful to have explicit examples for what to do after an `rc` release, and what to do between `rc` releases, in my opinion. We always do `rc` releases these days, so it's pretty relevant to the release process.

AlexWaygood (commented):
Thank you! I still find some of this a bit confusing though (sorry, I'm a bit of a packaging noob).

In the original PR that added this policy, there was [some talk](https://github.com/python/typing_extensions/pull/601#discussion_r2100852694) of not having to manually do these bumps at all, and instead have this kind of thing be done automatically. I guess nothing came of that in the end?

CONTRIBUTING.md
srittau: I agree, I also found it confusing. I'll revert this change (mostly).

srittau: Actually, I'll just remove it.

CONTRIBUTING.md
srittau: I've added some more examples and guidance. This seems fairly complex, but it really isn't. Before a release: Make sure the version number makes sense. After a release: Bump patch and add `.dev0`.

JelleZijlstra: So if we just released 4.1.1, when would we go to 4.2dev0 instead of 4.1.2dev0?

srittau: I think we could (should?) bump the version to `4.2.0dev0` when adding the first new feature that would require a minor version bump. (That's how I do it for my projects.)

srittau: Alternatively we could always bump the minor version after a release and use a separate branch from the last release when we need to do a patch release. (In my opinion the best solution for semver projects with frequent changes, but probably a bit too complicated for typing-extensions).

CONTRIBUTING.md
```diff
@@ -93,15 +88,18 @@ pipx run pre-commit run -a
...
+  When an urgent hotfix is necessary (and you bump only the patch version), you can skip the release candidates.
 
 - Ensure that GitHub Actions reports no errors.
 
 - Check that `CHANGELOG.md` accurately reflects all important changes committed to `main` since the previous release.
 
 - Update the version number in `typing_extensions/pyproject.toml` and in
-  `typing_extensions/CHANGELOG.md`.
+  `typing_extensions/CHANGELOG.md` by removing `.dev0` and potentially adding the `rcX` suffix:
+  - `4.1.0.dev0` → `4.1.0rc1`
```
AlexWaygood:
```suggestion
  - `4.1.0.dev0` → `4.1.0rc1` for the initial release candidate
```

CONTRIBUTING.md
```diff
@@ -93,15 +88,18 @@ pipx run pre-commit run -a
 
 # Workflow for PyPI releases
 
-- Make sure you follow the versioning policy in the documentation
-  (e.g., release candidates before any feature release, do not release development versions)
+- Make sure to release at least one release candidate (starting with `rc1`) before making a full release. E.g. release `4.2.0rc1` before releasing `4.2.0`.
+  When an urgent hotfix is necessary (and you bump only the patch version), you can skip the release candidates.
```
AlexWaygood: Maybe we should just link to https://typing-extensions.readthedocs.io/en/latest/#versioning-and-backwards-compatibility rather than duplicating that documentation in part here

CONTRIBUTING.md
```diff
@@ -93,15 +88,18 @@ pipx run pre-commit run -a
...
 - Ensure that GitHub Actions reports no errors.
 
 - Check that `CHANGELOG.md` accurately reflects all important changes committed to `main` since the previous release.
 
 - Update the version number in `typing_extensions/pyproject.toml` and in
-  `typing_extensions/CHANGELOG.md`.
+  `typing_extensions/CHANGELOG.md` by removing `.dev0` and potentially adding the `rcX` suffix:
+  - `4.1.0.dev0` → `4.1.0rc1`
+  - `4.1.0rc2.dev0` → `4.1.0rc2` or `4.1.0` when making a full release
+  - `4.1.1.dev0` → `4.1.1rc1` or `4.1.1` when making an urgent hotfix
```
AlexWaygood: I'm not sure an rc for a patch release really makes sense? In general we don't do patch releases unless it's to fix a regression in the latest minor release, and if there's a regression then I think it always makes sense to get the fix out as quickly as possible, rather than doing an rc for the patch and then waiting a week to release it "properly". So here I'd just say

```suggestion
  - `4.1.1.dev0` → `4.1.1` when making an urgent hotfix
```

CONTRIBUTING.md
```diff
@@ -111,4 +109,6 @@ pipx run pre-commit run -a
 - Release automation will finish the release. You'll have to manually
   approve the last step before upload.
 
-- After the release has been published on PyPI upgrade the version in number in [pyproject.toml](/pyproject.toml) to a `dev` version of the next planned release. For example, change 4.1.1 to 4.X.X.dev, see also [Development versions](#development-version). # TODO decide on major vs. minor increase.
+- After the release has been published on PyPI upgrade the patch or rc version number in [pyproject.toml](/pyproject.toml) and append `.dev0`. For example:
+    - `4.1.1` → `4.1.2.dev0`
+    - `4.1.1rc1` → `4.1.1rc2.dev0`
```
AlexWaygood: here as well I think I'd avoid an example that uses an rc for a patch release, because that just seems very unlikely to happen in practice

```suggestion
    - `4.1.0rc1` → `4.1.0rc2.dev0`
```

CONTRIBUTING.md
AlexWaygood: > I think we could (should?) bump the version to `4.2.0dev0` when adding the first new feature that would require a minor version bump. (That's how I do it for my projects.)

I guess that makes sense, but I don't feel excited about having to think every time we land a PR about whether the change is significant enough to warrant bumping the version number

CONTRIBUTING.md
srittau: Good point. I'll reword this with a link to the policy.

CONTRIBUTING.md
srittau: In practice, you will recheck that before making the next release anyway, so that's not super-critical.

CONTRIBUTING.md
```diff
@@ -93,15 +88,18 @@ pipx run pre-commit run -a
...
 
 - Ensure that GitHub Actions reports no errors.
 
 - Check that `CHANGELOG.md` accurately reflects all important changes committed to `main` since the previous release.
 
 - Update the version number in `typing_extensions/pyproject.toml` and in
-  `typing_extensions/CHANGELOG.md`.
+  `typing_extensions/CHANGELOG.md` by removing `.dev0` and potentially adding the `rcX` suffix:
+  - `4.1.0.dev0` → `4.1.0rc1` for the initial release candidate
+  - `4.1.0rc2.dev0` → `4.1.0rc2` or `4.1.0` when making a full release
```
AlexWaygood:
```suggestion
  - `4.1.0rc2.dev0` → `4.1.0rc2` or `4.1.0`, depending on whether the release is a second release candidate or a full release
```

AlexWaygood (approved):
Thanks!

AlexWaygood (approved):

AlexWaygood:
> In the original PR that added this policy, there was [some talk](https://github.com/python/typing_extensions/pull/601#discussion_r2100852694) of not having to manually do these bumps at all, and instead have this kind of thing be done automatically. I guess nothing came of that in the end?

I talked this idea through with Codex, and it seems like it would only be possible if we had setuptools-scm as a build dependency or maintained our own in-tree build backend. setuptools-scm depends on typing_extensions, which rules that out as an option, I think (the redistributors would kill us). And manual bumps seem like less complexity overall than having to maintain a custom build backend just for typing_extensions. So sticking with manual version bumps after releases does seem best to me.

PR #783, opened 2026-07-27
use `LiteralString` for `Sentinel` name

DetachHead:
see https://github.com/python/typeshed/pull/16094

### DIFF
--- a/src/typing_extensions.py
+++ b/src/typing_extensions.py
@@ -188,7 +188,7 @@ else:
 
         def __init__(
             self,
-            __name: str = _sentinel_placeholder,
+            __name: LiteralString = _sentinel_placeholder,
             __repr: typing.Optional[str] = _sentinel_placeholder,
             /,
             *,


AlexWaygood:
You have a NameError because you're trying to use `LiteralString` before it's defined in this file. But honestly I'd just remove the annotation here. We don't run a type checker in CI (this library defines the runtime shims for defining the type system, it would be very hard), and type checkers and typeshed just ignore any type hints that do appear here. Having anything except the most basic, uncontroversial type hints here is confusing because it implies that they're either checked in CI or useful to end users

DetachHead:
oops, i couldn't find `typing_extensions.pyi` in the typeshed repo so i assumed the types were defined here, but i guess that was the fault of github's abysmal search. i'll just close this then

AlexWaygood:
It's (somewhat confusingly) in the stdlib directory in typeshed, because if we pretend it's part of the stdlib type checkers let us use symbols defined there in `builtins.pyi` (otherwise we'd have to wait until they're defined in the `typing` module on all supported Python versions)

PR #792, opened 2026-08-29
Pin CI tools and manage dependency updates with Renovate

AlexWaygood:
> [!NOTE]
> We would need to enable the Renovate GitHub App for `python/typing_extensions` before landing this PR. Adding the configuration does not enable the app, and this PR removes the existing Dependabot update configuration. See [Renovate's installation instructions](https://docs.renovatebot.com/getting-started/installing-onboarding/#repository-installation).

## Pin more CI dependencies while keeping their updates automated.

Pinning GitHub Actions to commits pins the installer code, but actions can still download unpinned tools on each run: `setup-uv` and Codecov currently do so. This PR pins those tool versions and switches from Dependabot to Renovate so that these additional pins can receive automated update PRs instead of requiring manual maintenance.

The PR pins uv to `0.12.7` in all ten `setup-uv` steps, pins the Codecov CLI to `v11.3.1`, and pins the Python pre-commit hook dependencies `json5` and `validate-pyproject-schema-store[all]`. It also replaces the Dependabot schema hook with the Renovate schema hook and declares hook languages explicitly so Renovate can discover their additional dependencies.

The monthly schedule and seven-day release cooldown are retained, with updates grouped under "CI dependencies" and a Dependency Dashboard enabled. Initial digest pinning is exempt from the cooldown because it pins a version already in use. The configuration does not enable automerging.

Renovate will update:

| Dependency source | Updates covered |
| --- | --- |
| GitHub Actions | Action versions and commit hashes, retaining full commit pins and full SemVer comments. The PyPI publishing action gains a version comment so Renovate can identify its existing pin. |
| uv | The annotated `version` and SHA-256 `checksum` inputs together in `ci.yml` and `third_party.yml`. |
| Codecov CLI | The annotated `version` input in `ci.yml`. Codecov's existing download verification remains enabled. |
| pre-commit | Hook repository revisions, including their frozen hashes/version comments, and the versioned additional dependencies: JSON5, the pyproject schema store, and ShellCheck. |
| Python build requirements | Requirements under `[build-system].requires` in `pyproject.toml`, currently the `flit_core` version range. This PR does not change that range. |

Renovate will not update Python versions or support metadata, test matrices, GitHub runner images, containers/services, or the Read the Docs environment.

## uv checksums

The uv checksum is pinned in addition to the version because a version selects a release, while a checksum fixes the expected archive contents. The pinned `setup-uv` action uses an explicit checksum or its built-in checksum table for the default download source; if neither supplies a checksum, it skips verification.

Renovate's native `setup-uv` support updates the version without the checksum. The custom manager therefore uses Renovate's `github-release-attachments` datasource to update both fields together.


JelleZijlstra (approved):

AlexWaygood:
@python/organization-owners -- I _believe_ I requested access for the Renovate GitHub app to this repo, but I think I need an organisation admin to approve it. Would one of you be able to do so?

